PR #9, opened 2020-10-15
Added Triton server, client interfaces and support for perf_analyzer 

xprotobeast2:


Tabrizian (changes requested):

model_analyzer/triton/client.py
```diff
@@ -0,0 +1,166 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
whoisj, line 1: The model analyzer is provided under the Apache 2 license, are these headers compatible? Have you cleared this with NVIDIA legal?

xprotobeast2: @dzier mentioned that we needed to switch to BSD3 for this in [another PR](https://github.com/triton-inference-server/model_analyzer/pull/7#discussion_r505006282).

model_analyzer/analyzer/perf_analyzer.py
```diff
@@ -0,0 +1,192 @@
...
+            '-v' : None,
+            '-v -v' : None
+        }
+
+    def to_cli_string(self):
+        """Utility function to convert a config into a
+        string of arguments to the perf_analyzer with CLI.
+        """
+        args = ['{} {}'.format(k,v) for k,v in self._options.items() if v]
+
```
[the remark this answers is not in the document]
xprotobeast2: I'd left them for comments I forgot to add. Wil fix this.

model_analyzer/triton/client.py
```diff
@@ -0,0 +1,166 @@
...
+
+class TritonGRPCClient(TritonClient):
+    """
+    Concrete implementation of TritonClient
+    for GRPC
+    """
+    def __init__(self, config):
+        super().__init__(config=config)
+        self._client = grpcclient.InferenceServerClient(
+            url=self._client_config['url'])
```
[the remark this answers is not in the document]
xprotobeast2: Yeah this might be better. Perhaps I should make separate files for the configs too.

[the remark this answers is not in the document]
dzier: It may not need to be a different folder, but it definitely needs to be separate .py files.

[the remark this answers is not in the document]
dzier: Actually, yes, it should be `.../triton/client/...` for all client related code and `.../triton/server/...` for all server related code.

model_analyzer/triton/server.py
```diff
@@ -0,0 +1,309 @@
...
+    
+class TritonServerFactory:
+    """
+    A factory for creating TritonServer instances
+    """
+
+    def __init__(self):
+        pass
+
+    def create_server(self, run_type, model_path, version, config):
```
dzier, line 136: This is a slight misuse of the Factory pattern here. It forces you to have the same signature for the different object. The correct way is to have separate create_x_server methods.  Here is a good example.

https://refactoring.guru/design-patterns/factory-method/python/example

The point of the factory method to separate the dependency from user of the interface and the classes that derive the interface.

model_analyzer/triton/server.py
```diff
@@ -0,0 +1,309 @@
...
+        # FIXME this should not really be a server function
+        if self._server_config['allow-http'] is not False:
+            http_port = self._server_config['http-port'] or 8000
+            url = "http://localhost:{}/v2/health/ready".format(http_port)
+        else:
+            # FIXME to use GRPC to check for ready also
+            raise Exception('allow-http must be True in order to use wait_for_server_ready')
+        
+        retries = num_retries
+        # poll ready endpoint for number of retries
```
dzier, line 182: add empty line before comments, unless comment is the first thing after a structural line (def, if, else, etc.)

model_analyzer/triton/server.py
```diff
@@ -0,0 +1,309 @@
...
+        self._tritonserver_image = TRITONSERVER_IMAGE + version + '-py3'
+        self._tritonserver_container = None
+        
+
+    def start(self):
+        """
+        Starts the tritonserver docker container
+        """
+        # get devices using CUDA_VISIBLE_DEVICES
+        CUDA_VISIBLE_DEVICES = os.environ.get('CUDA_VISIBLE_DEVICES', '') # default to cpu
```
dzier, line 214: refrain from trailing comments.

tests/test_perf_analyzer.py
```diff
@@ -0,0 +1,117 @@
...
+        
+        # Create a config
+        self.config = PerfAnalyzerConfig()
+        self.config['model-name'] = TEST_MODEL_NAME
+
+        # Triton Server
+        self.server = None
+
+    def test_perf_analyzer_config(self):
+
```
dzier, line 54: Do not have empty lines after a `def`

tests/test_perf_analyzer.py
```diff
@@ -0,0 +1,117 @@
...
+        self.server.stop()
+
+    def tearDown(self):
+
+        # In case test raises exception
+        if self.server is not None:
+            self.server.stop()
+
+if __name__ == '__main__':
+    unittest.main()
```
dzier, line 117: Add newline

tests/test_triton_client.py
```diff
@@ -0,0 +1,224 @@
...
+        self.server_config = TritonServerConfig()
+        self.server_config['model-repository'] = MODEL_REPOSITORY_PATH
+        self.server_config['model-control-mode'] = 'explicit'
+
+        # Set CUDA_VISIBLE_DEVICES
+        os.environ['CUDA_VISIBLE_DEVICES'] = '0'
+
+        # Create and start the server
+        self.server_factory = TritonServerFactory()
+        try:
```
dzier, line 56: Is there a way to wrap this call in a `self.assert_xyz(...)` test?

model_analyzer/triton/client.py
```diff
@@ -0,0 +1,166 @@
...
+            communicate with the server
+        
+        config: TritonClientCnfig
+        """
+        if protocol is 'grpc':
+            return TritonGRPCClient(config=config)
+        elif protocol is 'http':
+            return TritonHTTPClient(config=config)
+        else:
+            raise Exception("The protocol '{}' for Triton Inference Server"
```
dzier, line 76: Use fstring

tests/test_tritonserver.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def tearDown(self):
+
+        # In case test raises exception
+        if self.server is not None:
+            self.server.stop()
+
+if __name__ == '__main__':
+    unittest.main()
+
```
dzier, line 168: remove any white space from this line

tests/test_tritonclient.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def tearDown(self):
+
+        # In case test raises exception 
+        self.server.stop()
+        pass
+
+if __name__ == '__main__':
+    unittest.main()
+
```
dzier, line 213: remove white space from this line

dzier (changes requested):

model_analyzer/analyzer/perf_analyzer.py
```diff
@@ -0,0 +1,192 @@
...
+        
+        Returns
+        -------
+        List of tuples of (params, output) where params are the 
+        set of parameters provided for that run and output is the 
+        stdout from the perf_analyzer.
+        """
+
+        # Create a config for each combination of parameters
+        combos = list(product(*tuple(sweep_params.values())))
```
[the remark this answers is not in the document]
xprotobeast2: This function takes a dict of lists like `{ 'batch-size' : [1,2], 'concurrency-range' : [2,4] }`. This line creates a list of tuples like [(1,2), (1,4), (2,2), (2,4)]. This can then be zipped with the keys from dict to make a config that specifies the parameters for a single run of perf_analyzer.

Maybe the variable should be named differently, or some more commentary is needed?

model_analyzer/triton/client.py
```diff
@@ -0,0 +1,166 @@
...
+                            " is not supported by the model analyzer.".format(protocol))
+
+class TritonClient:
```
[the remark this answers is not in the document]
xprotobeast2: This is the interface defined for a Triton Client in the original ModelAnalyzer. (load, unload). I also added a wait_for_serve_ready, because I was unsure whether it's best implemented in the server interface or the client interface. I think thats the only thing this file has in common with triton/server.py

model_analyzer/triton/server.py
xprotobeast2: So the correct way according to that example method would be to have a `TritonHTTPClientFactory` and a `TritonGRPCClientFactory` derived from `TritonClientFactory`? Would it be enough to just have 2 static create methods in the `TritonClientFactory` class : `create_http_client` and `create_grpc_client`?

tests/test_triton_client.py
xprotobeast2: I don't understand. Am I trying to ensure that self.server_factory is defined? Maybe I could try `self.assertIsInstance(self.server_factory, TritonServerFactory)`

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        cmd = ['perf_analyzer']
+        cmd += self._config.to_cli_string().replace('=', ' ').split()
+
+        # Synchronously start and finish run
+
+        try:    
+            out = check_output(cmd, stderr=STDOUT, encoding='utf-8')
+        except CalledProcessError as e:
+            raise Exception(f"perf analyzer returned with exit"
```
Tabrizian, line 105: What about creating `TritonModelAnalyzerException` class so that this can be reused across the application?

model_analyzer/triton/client/client_factory.py
```diff
@@ -0,0 +1,45 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+class TritonClientFactory:
+    """
+    Base client creator class that declares
+    a factory method
+    """
+
+    def create_client(self, config):
```
Tabrizian, line 34: We can probably use ABC and marking this as an abstract class instead of raising `NotImplementedError`

model_analyzer/triton/client/http_client_factory.py
```diff
@@ -0,0 +1,49 @@
...
+from .http_client import TritonHTTPClient
+
+
+class TritonHTTPClientFactory(TritonClientFactory):
+    """
+    Concrete creator class that creates 
+    TritonHTTPClients
+    """
+
+    def create_client(self, config) -> TritonClient:
```
Tabrizian, line 38: Do we want to type hint the return values? If we want to do this we should be consistent across all the methods.

model_analyzer/triton/server/server.py
```diff
@@ -0,0 +1,90 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import requests
+import time
+
+WAIT_FOR_READY_NUM_RETRIES = 100    
+SERVER_HTTP_PORT = 8000
+
+class TritonServer:
```
Tabrizian, line 33: I think using `ABC` and abstract methods would be better.

tests/test_perf_analyzer.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.server.stop()
+
+    def tearDown(self):
+        # In case test raises exception
+        if self.server is not None:
+            self.server.stop()
+
+if __name__ == '__main__':
+    unittest.main()
```
Tabrizian, line 102: Newline

Tabrizian (changes requested):

tests/test_triton_client.py
dzier: So in C++ Boost:test, we have a test for something like `test_no_throw(obj = do_something())`. Basically, it will fail if any exception is thrown. This way, you don't have to wrap the call in a `try...catch` statement.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -0,0 +1,108 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from itertools import product
+from subprocess import check_output, CalledProcessError, STDOUT
+
+
+class PerfAnalyzer:
+    """A wrapper class for the perf_analyzer. This
```
dzier, line 32: Let's be consistent with the definition strings. All definitions for either a class or method should follow:
```
"""
The definition starts on it's own newline. There should
be no newline after the definition text.
"""
```

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -0,0 +1,108 @@
...
+                self._config[key] = val
+            
+            # Run perf_analyzer
+            run_output = self._run_perf_analyzer()
+            run_outputs.append((run_config, run_output))
+        
+        return run_outputs
+
+    def _run_perf_analyzer(self):
+        """Runs the perf_analyzer tool locally
```
dzier, line 87: fix definition string

dzier (changes requested):

model_analyzer/triton/client/http_client_factory.py
dzier: Hints are needed, however, so is `@abstractmethod` which is missing here.

model_analyzer/triton/server/server.py
dzier: Agree

model_analyzer/triton/server.py
dzier: It C++, that is the way. However, I think python does things a bit differently. For now, create the two separate methods. Note you will need to do the same thing in your other Factory class.  I will look more in to the Factory pattern in python to see if further refactoring is needed.

dzier (changes requested):

model_analyzer/triton/server.py
xprotobeast2: In my latest commit I created separate classes, the advantage being that it is extensible without needing to change the base interface (as opposed to adding static create_x methods.

tests/test_triton_client.py
xprotobeast2: I think `unittest` has an `assertRaises()` but nothing for the opposite. In my latest commit I've just removed the try catch, as the test will fail if an exception is thrown anyway.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
xprotobeast2: Ah yes. Good idea thank you.

Tabrizian (approved):

model_analyzer/triton/client.py
whoisj: This project, in particular, was cleared to use Apache 2.0. Switching the license means involving NVIDIA legal, if for no other reason that they need to know things changed.

model_analyzer/triton/client/client.py
```diff
@@ -0,0 +1,144 @@
...
+                    time.sleep(0.1)
+                    num_retries -= 1
+            except Exception as e:
+                time.sleep(0.1)
+                num_retries -= 1
+                pass
+        raise TritonModelAnalyzerException("Could not determine server readiness. "
+                                           "Number of retries exceeded.")
+    
+    def load_model(self, model_name):
```
Tabrizian, line 70: Ashwin, I was looking at the PR again. I was wondering what do you think about adding a `TritonModel` or `Model` class that encapsulates all the information related to a Triton Model and using that instead of only relying on the `model_name` string.

xprotobeast2: I think you're right. For future phases, I'm sure we'll have to create either a `TritonModel` or a `TritonModelConfig` class that encapsulates the information. Do you think the load model, and unload model interface should belong to `TritonModel`?

Tabrizian: Sorry for the delayed response. I think it would be great to have `load` and `unload` as a part of the model class. But I guess the problem will be we have to pass a client to every TritonModel. I guess we can probably create a TritonClient in every model or make the TritonClient singleton (if it doesn't need to be instantiated) so that in the model constructor we used the same instance all the time.

xprotobeast2: Hmm I suppose what we could have is a singleton class for each concrete implementation of the TritonClient? So there is one GRPC client and one HTTP client for each? Now I'm wondering whether we even need a standalone TritonClient interface, or whether we'll only ever need to interact via a `TritonModel` interface.

Tabrizian: In the C# version of the Model Analyzer, they only used TritonClient to `load` models into the Triton. Maybe in the future (esp. with the introduction of rate limiter) we may need TritonClient for setting up the general way that Triton works independently from the models. So I think it would be better to keep it as a separate entity. The other thing is that we will have a single type of TritonClient (either gRPC or HTTP) throughout the application execution. So if it is instantiated once (using gRPC or HTTP), the same instance can be used throughout the application lifetime.

dzier: You don't want a singleton TritonClient. That would imply that both HTTPClient and GRPClient exist at the same time. A singleton object is the same a global object. Why would everything need to know about both HTTPClient and GRPClient? 

In this case, you want a top level Model Interface that defines the load/unload and other common methods. Then a derived class for HTTP Client and GRPC Client. You can then use a Constructor Pattern to create them. However, unless there is a use case, both don't need to exist at the same time. You will use one or the other.

dzier: FYI, We will definitely need both a Model and ModelConfig object. I don't think you need to specify Triton exactly unless we are analyzing models for a non Triton platform. The Model object should hold all the meta data about the model, while the ModelConfig holds the configuration options for a Model under test. You may have an array of ModelConfigs, one for each test. Then when we create the actual Triton ModelConfig, the writer would take both as input and write out the correct config file.

dzier: But also don't plan for what ifs right now. Instead, create the minimal amount of code needed to pass the tests and use cases for this phase. Keep the code agile and clean. Then we can refactor as needed when more use cases are added.

dzier (approved):
Make sure that you are running the unit tests and they pass before merging.

Tabrizian:
Also, make sure to run the formatter before merge.

dzier (approved):

PR #10, opened 2020-10-22
Add model-analyzer CLI

Tabrizian:
This is how the `model-analyzer` CLI look like right now:
```
model-analyzer                               
usage: model_analyzer [-h] -m MODEL_REPOSITORY [--triton-version TRITON_VERSION] --model-names MODEL_NAMES [-b B] [--export EXPORT] [--concurrency-range CONCURRENCY_RANGE]
model_analyzer: error: the following arguments are required: -m/--model-repository, --model-names
```

CLI Help
```
model-analyzer --help
usage: model_analyzer [-h] -m MODEL_REPOSITORY [--triton-version TRITON_VERSION] --model-names MODEL_NAMES [-b B] [--export EXPORT] [--concurrency-range CONCURRENCY_RANGE]

optional arguments:
  -h, --help            show this help message and exit
  -m MODEL_REPOSITORY, --model-repository MODEL_REPOSITORY
                        Model repository location
  --triton-version TRITON_VERSION
                        Triton Server version
  --model-names MODEL_NAMES
                        Comma delimited list of the model names to be profiled
  -b B                  Batch size to use for the profiling
  --export EXPORT, --export-path EXPORT
                        Export path to store the results
  --concurrency-range CONCURRENCY_RANGE, -c CONCURRENCY_RANGE
                        Concurrency range values <start:end:step>
```


model_analyzer/cli/model.py
```diff
@@ -0,0 +1,72 @@
...
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+from argparse import ArgumentParser
```
dzier, line 26: We should have an empty line after the copyright statement and the start of the code.

model_analyzer/cli/model.py
```diff
@@ -0,0 +1,72 @@
...
+            '--model-repository',
+            type=str,
+            required=True,
+            help='Model repository location')
+        self._parser.add_argument(
+            '--triton-version',
+            type=str,
+            help='Triton Server version')
+        self._parser.add_argument(
+            '--model-names',
```
dzier, line 50: Is this how the original did it? We should make this input additive as well.

model_analyzer/cli/model.py
```diff
@@ -0,0 +1,72 @@
...
+            type=str,
+            required=True,
+            help='Comma delimited list of the model names to be profiled')
+        self._parser.add_argument(
+            '-b',
```
dzier, line 55: Always provide a long name `--batch-size`. In the future, this will be a list as well.

model_analyzer/cli/model.py
```diff
@@ -0,0 +1,72 @@
...
+            default=1,
+            help='Batch size to use for the profiling')
+        self._parser.add_argument(
+            '--export',
+            '--export-path',
+            type=str,
+            help='Export path to store the results')
+        self._parser.add_argument(
+            '--concurrency-range',
+            '-c',
```
dzier, line 66: Short name first, then long name

model_analyzer/cli/model.py
```diff
@@ -0,0 +1,72 @@
...
+            required=True,
+            help='Comma delimited list of the model names to be profiled')
+        self._parser.add_argument(
+            '-b',
+            type=int,
+            default=1,
+            help='Batch size to use for the profiling')
+        self._parser.add_argument(
+            '--export',
+            '--export-path',
```
dzier, line 61: Is there a reason for both long names?

model_analyzer/entrypoints/main.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
dzier, line 1: This should just be part of cli/ or at the top model_analyzer/ folder. What is the reason for a separate entrypoints folder?

dzier (changes requested):

model_analyzer/cli/model.py
```diff
@@ -0,0 +1,72 @@
...
+    CLI class to parse the commandline arguments
+    """
+
+    def __init__(self):
+        self._parser = ArgumentParser(prog='model_analyzer')
+        self._add_arguments()
+
+    def _add_arguments(self):
+        self._parser.add_argument(
+            '-m',
```
xprotobeast2, line 40: `-m` might get confusing because its what perf_analyzer uses for `model-name`. I think there is a decent amount of separation in the code between the CLI and the perf_analyzer, but might be worth noting.

model_analyzer/cli/model.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
xprotobeast2, line 1: This file is named `model.py`. We have a `model.py` in `model_analyzer/monitor/` as well as `model_analyzer/triton/model/`. Is there a reason for this convention? I think its a little confusing. This file can be named `cli.py` and the other one `monitor.py`

xprotobeast2 (changes requested):

model_analyzer/cli/model.py
Tabrizian: I was just trying to follow the C# arguments. I will remove `export-path`.

model_analyzer/cli/model.py
Tabrizian: This is for model-repository and was used in the C# configurations. Do we want to break this?

model_analyzer/cli/model.py
Tabrizian: I was trying to follow David Zier's comments for the base classes. We can use `_base.py` if this is less confusing. Or I can use the same name for as the package folder.

model_analyzer/cli/model.py
Tabrizian: They had a comma-delimited list of models. Like we have here. We can make it input additive too.

model_analyzer/cli/model.py
xprotobeast2: Ah okay i think we can keep it for phase 1

model_analyzer/cli/model.py
Tabrizian: By input additive do you mean having multiple --model-name instead of this list?

model_analyzer/cli/model.py
xprotobeast2: I assumed that was a particular example, rather than a general rule to call the base class `model`. I also believe it may have been a typo (I could be wrong about that). I think since we have the one class per file rule, we should name the file corresponding to the class inside.

Tabrizian: Got it, changed the file name to `cli.py`.

requirements.txt
```diff
@@ -0,0 +1,28 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+docker>=4.3.1
+pynvml>=8.0.4
```
xprotobeast2, line 28: Should `numba` be here now? I remember you said you were going to use `pycuda` instead. Is that why it isn't here?

xprotobeast2 (approved):

requirements.txt
Tabrizian: I was planning to install it using conda in Dockerfile since I thought there will be issues when installing it using pip. Fixed it.

Tabrizian: PyCUDA doesn't seem to be supported by NVIDIA so I think it would be better to stick with the numba for now.

model_analyzer/cli/model.py
dzier: yes, but let's hold off on that if they already expect a CSV list

dzier (approved):

PR #11, opened 2020-10-23
Record Aggregator to collect and aggregate records

xprotobeast2:


model_analyzer/record/record_aggregator.py
```diff
@@ -0,0 +1,159 @@
...
+        the given list of criteria.
+
+        Parameters
+        ----------
+
+        headers : list of str
+            the names of the records we are
+            imposing the filter criteria on.
+
+        filters : list of callables
```
Tabrizian, line 65: Nice! I think it would great if we mention in this docstring that the filters should have the same size as the headers or should not be provided.

tests/test_record_aggregator.py
```diff
@@ -0,0 +1,139 @@
...
+        # Test for malformed inputs
+        with self.assertRaises(Exception):
+            record_aggregator.get_records(filters=[(lambda x: False)])
+        with self.assertRaises(Exception):
+            record_aggregator.get_records(headers=["header1", "header2"],
+                                          filters=[(lambda x: False)])
+
+        # Insert 2 more throughputs
+        record_aggregator.insert_record(PerfThroughput(1,
+                                                       PerfAnalyzerConfig()))
```
Tabrizian, line 83: I think it is a little bit strange that we have to include PerfAnalyzerConfig with every PerfRecord.

tests/test_record_aggregator.py
```diff
@@ -0,0 +1,139 @@
...
+        self.assertEqual(min_vals[throughput_record.header()],
+                         0,
+                         msg="Aggregation failed with min")
+        self.assertEqual(average_vals[throughput_record.header()],
+                         4.6,
+                         msg="Aggregation failed with average")
+
+
+if __name__ == "__main__":
+    unittest.main()
```
Tabrizian, line 139: New line

model_analyzer/record/record_aggregator.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class RecordAggregator:
+    """
+    Stores a collection of Record objects.
+    """
+    def __init__(self):
+        self._records = defaultdict(list)
+
+    def insert_record(self, record):
```
dzier, line 37: '_record' is redundant and not needed. Just use 'insert'

model_analyzer/record/record_aggregator.py
```diff
@@ -0,0 +1,159 @@
...
+            A record to be inserted
+        """
+        if isinstance(record, Record):
+            header = record.header()
+            self._records[header].append(record)
+        else:
+            raise TritonModelAnalyzerException(
+                "Can only add objects of type 'Record' to RecordAggregator")
+
+    def get_records(self, headers=None, filters=None):
```
dzier, line 53: This method is doing more than getting the records. Consider renaming to a more appropriate descriptor, such as `aggregate_records`,

model_analyzer/record/record_aggregator.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Returns
+        -------
+        dict
+            keys are requested headers
+            and values are the aggregated values
+        """
+
+        return {
+            k: reduce_func([record.value() for record in self._records[k]])
```
dzier, line 157: I am not a big fan of this style of programming as it obfuscates what is actually going on. Pull it out of the return and create an actual object the is returned. The compiler will optimize out any unnecessary variables.

model_analyzer/record/record_aggregator.py
dzier: or `format_records`

tests/test_record_aggregator.py
```diff
@@ -0,0 +1,139 @@
...
+from model_analyzer.analyzer.perf_analyzer.perf_config import PerfAnalyzerConfig
+
+
+class TestRecordAggregatorMethods(unittest.TestCase):
+    def test_insert_record(self):
+        record_aggregator = RecordAggregator()
+
+        self.assertEqual(record_aggregator.total(), 0)
+
+        throughput_record = PerfThroughput(5, PerfAnalyzerConfig())
```
dzier, line 40: Is PerfThroughput a simple data structure, or does it actually talk to the PerfAnalyzer?

tests/test_record_aggregator.py
```diff
@@ -0,0 +1,139 @@
...
+        record_aggregator.insert_record(throughput_record)
+
+        # Assert record is added
+        self.assertEqual(record_aggregator.total(), 1)
```
dzier, line 44: Test the value of the record it returns. I have no idea that the aggregator didn't change the value. Lock in the behavior that the record does not change.

tests/test_record_aggregator.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def test_headers(self):
+        record_aggregator = RecordAggregator()
+
+        throughput_record = PerfThroughput(5, PerfAnalyzerConfig())
+        record_aggregator.insert_record(throughput_record)
+
+        self.assertEqual(record_aggregator.headers()[0],
```
dzier, line 52: What are the headers? Are there actual strings that are compared? What are those strings?

tests/test_record_aggregator.py
```diff
@@ -0,0 +1,139 @@
...
+                                          filters=[(lambda x: False)])
+
+        # Insert 2 more throughputs
+        record_aggregator.insert_record(PerfThroughput(1,
+                                                       PerfAnalyzerConfig()))
+        record_aggregator.insert_record(
+            PerfThroughput(10, PerfAnalyzerConfig()))
+
+        # Test get with filters
+        retrieved_records = record_aggregator.get_records(
```
dzier, line 88: Test cases should be specific. You should have a separate one for `get_records_default` and `get_records_filtered`.

dzier (changes requested):

tests/test_record_aggregator.py
xprotobeast2: Yes, this has been changed in a different PR.

model_analyzer/record/record_aggregator.py
xprotobeast2: `aggregate_records` seems to have a different meaning to it, since we have a function that is called `aggregate`. I'll use `filter_records` for now.

model_analyzer/record/record_aggregator.py
xprotobeast2: Will change it for readability, but I think list/dictionary comprehensions are the most pythonic way to write this. They read like mathematical objects, and are something like 50% faster than the corresponding for loop semantics. Most python libraries are full of them.

tests/test_record_aggregator.py
xprotobeast2: Its just a simple data structure.

tests/test_record_aggregator.py
xprotobeast2: This is tested in `test_filter_records`

tests/test_record_aggregator.py
xprotobeast2: The headers are the human readable names of the metrics. They are what appear as the column name in the final output table.

dzier (approved):

PR #12, opened 2020-10-23
Define output interface, concrete file writer, and output table interface

xprotobeast2:


model_analyzer/output/file_writer.py
```diff
@@ -0,0 +1,78 @@
...
+            the output.
+        """
+        allowed_write_modes = ['a', 'w', 'a+', 'w+', 'rw', 'ra']
+        if self._filename:
+            if write_mode in allowed_write_modes:
+                try:
+                    with open(self._filename, write_mode) as f:
+                        f.write(table)
+                except IOError as e:
+                    f.close()
```
Tabrizian, line 72: I think `f.close()` is not required when you are using contexts but I'm not sure.

tests/test_file_writer.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Remove test files
+        if os.path.exists("test_file"):
+            os.remove("test_file")
+        if os.path.exists("test_file_stdout"):
+            os.remove("test_file_stdout")
+
+
+if __name__ == '__main__':
+    unittest.main()
```
Tabrizian, line 77: New line

tests/test_output_table.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def test_to_formatted_string(self):
+        table = OutputTable(headers=TEST_HEADERS)
+        for row in TEST_ROWS:
+            table.add_row(row)
+        self.assertEqual(table.to_formatted_string(), TEST_TABLE_STR)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
Tabrizian, line 125: New line

tests/test_file_writer.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # open and read file
+        with open(filename, 'r') as f:
+            content = f.read()
+
+        self.assertEqual(content, 'test')
+
+        # redirect stdout and create writer with no filename
+        old_stdout = sys.stdout
+        sys.stdout = open('test_file_stdout', 'w')
```
dzier, line 49: You do not want to actually write out to STD_OUT in a unit test. The unit test could not fail due to reasons beyond the code (out of memory, no space on disk, etc).

In cases like this, you want to use a string buffer in place of stdout. Then have a file pointer to the buffer instead of stdout and set the writer to use that. You may need to create a `TestFileWriter` that overrides the methods that actually write to a file or have it point internally to a buffer.

Then you also want to test the output. Write now, your test only shows that something was written, not what. 

Finally assertion tests should be in separate test cases.

dzier (changes requested):

tests/test_file_writer.py
xprotobeast2: So what I'm doing here is writing to a file on disk called `test_file_stdout`. `sys.stdout` is a file pointer to `STDOUT `maintained by the `sys `module. I don't think I'm actually writing to `STDOUT`. 

Also the test currently writes the string `"test"` to these files, and then the asserts check that the string `"test"` is present

Should it be a single assert per test case? I think that will be too many test cases.

dzier: You should have not have any file IO in a unit test. If possible, write to a string. This string should be compared to a known string constant. Preferably defined in the same test.

tests/test_output_table.py
dzier: still need a new line

dzier (changes requested):

tests/test_output_table.py
dzier: Again, still need a new line here.

dzier (changes requested):

dzier (approved):

PR #13, opened 2020-10-23
Refactor PerfRecord to allow aggregation

xprotobeast2:


model_analyzer/analyzer/perf_analyzer/perf_throughput.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Move this to `model_analyzer/record`.

model_analyzer/analyzer/perf_analyzer/perf_throughput.py
```diff
@@ -0,0 +1,74 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from model_analyzer.record.record import Record
+from model_analyzer.record.perf_record import PerfRecord
+from model_analyzer.model_analyzer_exceptions import TritonModelAnalyzerException
+
+
+class PerfThroughput(Record, PerfRecord):
```
Tabrizian, line 32: Why do we need to inherit from multiple classes? Can we inherit from `PerfRecord` such that `PerfRecord` is also an abstract class that inherits from `Record`?

model_analyzer/analyzer/perf_analyzer/perf_throughput.py
```diff
@@ -0,0 +1,74 @@
...
+        metric
+
+        Returns
+        -------
+        (any)
+            value of the metric
+        """
+        return self._throughput
+
+    def config(self):
```
Tabrizian, line 64: Do we want config to be a part of the record?

Tabrizian (changes requested):

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -27,16 +27,16 @@
...
+from .perf_throughput import PerfThroughput
 
 
 class PerfAnalyzer:
     """
     This class provides an interface for running workloads
     with perf_analyzer.
     """
-
     def __init__(self, config):
```
dzier, line 40: Leave the empty line. Vertical spacing is important to readability and clean code.

model_analyzer/analyzer/perf_analyzer/perf_latency.py
```diff
@@ -0,0 +1,75 @@
...
+        metric
+
+        Returns
+        -------
+        (any)
+            value of the metric
+        """
+        return self._latency
+
+    def config(self):
```
dzier, line 65: Why are we adding config to a record? What is the use case? I image that there will be a higher level object, something like `RunMetrics` that would hold the single instance of the config for that run, as well as arrays for all the metrics.

model_analyzer/analyzer/perf_analyzer/perf_throughput.py
dzier: Unless one is a pure interface, we should not inherit from multiple classes. If PerfRecord inherits Record, then this class just needs PerfRecord.

tests/test_perf_analyzer.py
```diff
@@ -105,26 +120,16 @@ def test_run(self):
             model_path=MODEL_LOCAL_PATH,
             version=TRITON_VERSION,
             config=server_config)
-        client = PerfAnalyzer(config=self.config)
+        perf_client = PerfAnalyzer(config=self.config)
```
dzier, line 123: There is no test here. A test requires some assert condition. Look for a way to make sure an assert is not thrown.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -48,75 +48,74 @@ def __init__(self, config):
...
+        perf_out : str
+            The raw output from the subprocess
+            call to perf_analyzer
+        """
+        data = {}
+        # Split output into lines and ignore last 3 lines
+        perf_out_lines = perf_out.split('\n')
+        for line in perf_out_lines[:-3]:
+            # Last word in line with Batch and concurrency
+            if 'Batch' in line:
```
dzier, line 97: There needs to be a unit test to test every one of this if/else conditions. I didn't see this.

dzier (changes requested):

model_analyzer/analyzer/perf_analyzer/perf_throughput.py
xprotobeast2: Both are pure interfaces in this case so I'm trying iman's suggestion.

model_analyzer/analyzer/perf_analyzer/perf_throughput.py
xprotobeast2: I think its useful to have the config for all the perf metrics, that way we can get the batch and concurrency that they were run with whenever needed.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
xprotobeast2: I'm using yapf python fomatting (It's what is used in triton as well). It deletes lines like this.

model_analyzer/analyzer/perf_analyzer/perf_latency.py
xprotobeast2: Thought it might be useful for the metric to store the batch concurrency etc that it was generated with, but i'll remove this.

tests/test_perf_analyzer.py
xprotobeast2: I'll add some more unit tests for cases where exceptions are thrown, but if an exception is thrown, the test will fail. The `unittest` module doesn't seem to have an `assertNotRaises`. This is implicit.  I could use a `try...except` block and force fail the test if there is an exception but I don't see the need.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
dzier: that's inconsistent. The empty line exists in other files.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -48,75 +48,64 @@ def __init__(self, config):
...
-            the perf_analyzer
+        List of Records for metrics
+        that are generated by the perf analyzer
 
         Raises
         ------
         TritonModelAnalyzerException
             If subprocess throws CalledProcessError
         """
+
```
dzier, line 66: Your formatter is not consistent. Empty line after comment here, but not in others.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
xprotobeast2: We should change everything to match this format. VSCode supports the yapf autoformatting as well. I will do this to all the files in my PRs. @Tabrizian should do the same.

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
xprotobeast2: Yeah it leaves out some things, like empty lines between code blocks (which I assume is what it thought this was).

model_analyzer/analyzer/perf_analyzer/perf_analyzer.py
Tabrizian: Sure, I will do this too.

dzier: Can you change the YAPF formatting to add the line. Without it, it looks like the descriptor will go to __init__, not the class. Unless a comment is tied to a specific line of code, there should always be an empty line after a comment.

dzier (changes requested):
Everything looks good except the empty line after comment. Try to configure YAPF to add the line.

dzier (approved):

Tabrizian (approved):

PR #14, opened 2020-10-23
Refactor GPURecord and GPUMemoryRecord to allow aggregation

xprotobeast2:


model_analyzer/monitor/nvml.py
```diff
@@ -66,23 +66,26 @@ def _monitoring_loop(self, tags):
...
                         gpu_device = self._gpus[i]
-                        memory_record = GPUMemoryRecord(
-                            gpu_device, memory.used, memory.total, memory.free)
-                        record_collector.insert(memory_record)
+                        records.append(
+                            GPUUsedMemory(device=gpu_device,
+                                          used_mem=memory.used))
+                        records.append(
+                            GPUFreeMemory(device=gpu_device,
+                                          free_mem=memory.free))
```
Tabrizian, line 82: `GPUTotalMemory` seems to be missing here.

model_analyzer/record/gpu_record.py
```diff
@@ -8,7 +8,7 @@
 #  * Redistributions in binary form must reproduce the above copyright
 #    notice, this list of conditions and the following disclaimer in the
 #    documentation and/or other materials provided with the distribution.
-#  * Neither the name of NVIDIA CORPORATION nor the names of its
```
Tabrizian, line 11: Did you want to change the license?

model_analyzer/record/record.py
```diff
@@ -24,14 +24,34 @@
...
 
-class Record:
+class Record(ABC):
     """
     This class is used for representing monitoring
     records
     """
 
-    def __init__(self, name):
-        self.time = time.time()
```
Tabrizian, line 36: What about the record times?

Tabrizian (changes requested):

model_analyzer/monitor/gpu_free_memory.py
```diff
@@ -23,17 +23,43 @@
 # OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+from model_analyzer.record.record import Record
```
dzier, line 26: Add empty lines after copyright

model_analyzer/monitor/gpu_used_memory.py
```diff
@@ -0,0 +1,66 @@
...
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+from model_analyzer.record.record import Record
+from model_analyzer.record.gpu_record import GPURecord
+
+
+class GPUUsedMemory(Record, GPURecord):
```
dzier, line 30: Remove Record

model_analyzer/monitor/gpu_free_memory.py
```diff
@@ -23,17 +23,43 @@
...
 from model_analyzer.record.gpu_record import GPURecord
 
 
-class GPUMemoryRecord(GPURecord):
+class GPUFreeMemory(Record):
```
dzier, line 30: Shouldn't this inherit GPURecord?

model_analyzer/monitor/gpu_free_memory.py
dzier: Do this for all the files

model_analyzer/monitor/gpu_free_memory.py
```diff
@@ -23,17 +23,43 @@
 # OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
```
dzier, line 23: These are records, why did you move them from `records/`

model_analyzer/record/record.py
```diff
@@ -8,7 +8,7 @@
 #  * Redistributions in binary form must reproduce the above copyright
 #    notice, this list of conditions and the following disclaimer in the
 #    documentation and/or other materials provided with the distribution.
-#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#  * Neither the record_type of NVIDIA CORPORATION nor the record_types of its
```
dzier, line 11: revert these changes

model_analyzer/record/gpu_record.py
```diff
@@ -8,7 +8,7 @@
...
+#  * Neither the record_type of NVIDIA CORPORATION nor the record_types of its
```
dzier, line 11: revert these changes

model_analyzer/record/record.py
```diff
@@ -24,14 +24,34 @@
...
+        """
+        Returns
+        -------
+        str
+            The full name of the
+            metric.
+        """
+
+    @abstractmethod
+    def value(self):
```
dzier, line 47: You should set this to return a float.

dzier (changes requested):

model_analyzer/monitor/nvml.py
xprotobeast2: I was wondering whether we need that or whether we could just compute it and pass it if we needed it. Is it not just GPUFreeMemory + GPUUsedMemory? Seemed like a redundant record.

model_analyzer/monitor/gpu_free_memory.py
xprotobeast2: They are records, but they are associated with the monitor. I'll move them back I think that makes sense.

model_analyzer/record/record.py
xprotobeast2: Should we restrict the value of a record to always be a float? We may need to have them be strings or custom types. 
I suppose I could change it because all we have now are floats.

model_analyzer/record/record.py
xprotobeast2: Added a timestamp abstract method that must be implemented.

model_analyzer/record/gpu_free_memory.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class GPUFreeMemory(GPURecord):
+    """
+    The free memory in the GPU.
+    """
+    def __init__(self, device, free_mem):
+        self._free_mem = free_mem
+        self._device = device
+        self._timestamp = time.time()
```
Tabrizian, line 38: Can you please also add the ability to override this value? For example, having a time parameter that allows you to override the default value which is the current time.

Tabrizian (dismissed):
Apart from that comment the rest LGTM.

model_analyzer/monitor/nvml.py
Tabrizian: Right, it may be redundant.

model_analyzer/record/gpu_free_memory.py
xprotobeast2: Yes. I think I can do that. Should that be for all records in general? Or only GPURecords?

Tabrizian: I think it should be for all the records. The monitoring library may provide the time for when the record was measured. I think DCGM provided that and that is the primary reason I am asking for this. Thanks.

xprotobeast2: Okay, I will add a method to the super class `update_time` that just sets it to the current time. Or do you want this to act like a sequence number? Maybe we can fix this in a separate PR closer to the use case?

dzier: This should be a relative time to the start of the test, not a date time stamp. This is how profilers like nsight systems work.

model_analyzer/record/gpu_used_memory.py
```diff
@@ -23,17 +23,58 @@
...
-        super().__init__('memory', device)
-        self.total = total
-        self.used = used
-        self.free = free
+        Returns
+        -------
+        float
+            time in seconds since epoch
+        """
+        return self._timestamp
```
dzier, line 80: new line

model_analyzer/record/record.py
```diff
@@ -23,15 +23,48 @@
...
+        metric
+
+        Returns
+        -------
+        float
+            value of the metric
+        """
+
+    @abstractmethod
+    def timestamp(self):
```
dzier, line 59: This functionality is common to all inherited classes, and you are duplicating the work in all derived classes. Instead, move the actual time variable into this class and then set that time in an _init method for this parent class. I assume inheritence works the same in Python, so the init of a parent class is called. The only thing that should go in child classes are code that is unique to that class. In this case, that is just the header. All other code can be in the parent class

dzier (changes requested):

dzier (commented):
Please rerun YAPF after the empty line fix.

model_analyzer/monitor/nvml.py
```diff
@@ -95,22 +99,22 @@ def start_recording_metrics(self, tags):
             *memory* and *compute*.
         """
```
dzier, line 100: empty line

model_analyzer/monitor/nvml.py
```diff
@@ -66,23 +67,26 @@ def _monitoring_loop(self, tags):
...
                 if tag == 'memory':
                     for i, handle in enumerate(self._nvml_handles):
                         memory = nvmlDeviceGetMemoryInfo(handle)
                         gpu_device = self._gpus[i]
-                        memory_record = GPUMemoryRecord(
-                            gpu_device, memory.used, memory.total, memory.free)
-                        record_collector.insert(memory_record)
+                        records.append(
+                            GPUUsedMemory(device=gpu_device,
+                                          used_mem=memory.used))
```
dzier, line 80: How does this work? The __init__ method for GPUUsedMemory and GPUFreeMemory requires a timestamp.

dzier (changes requested):

model_analyzer/monitor/nvml.py
xprotobeast2: I thought I set a default value at the very least. Not sure how this happened. Must have passed the test pre-merge conflict resolution.

dzier (approved):
I will approve for now, but it seems that this can be further refactored. There will be a lot of code duplication between the NVMLMonitor and the DCGMMonitor. For example, the monitor loop is common to both, so is the time sampling. This means that it should pulled to a common base class and shared.  The only thing that is unique is the call to `nvmlDeviceGetMemoryInfo`  In this case, I would wrap that in a virtual method in a parent class. Something like `@abstract get_device_memory() -> memory`. Then just have the child classes implement that one method. The main while loop would call the virtual method in the base class.

PR #21, opened 2020-11-02
Add end-to-end support

xprotobeast2:


model_analyzer/entrypoint.py
```diff
@@ -0,0 +1,210 @@
...
+
+def main():
+    """
+    Main entrypoint of model_analyzer
+    """
+
+    # Get args
+    cli = CLI()
+    args = cli.parse(sys.argv)
+    print(args)
```
dzier, line 84: This should be gated by a verbosity flag. We should also use logging, and not just prints

dzier (changes requested):

model_analyzer/entrypoint.py
```diff
@@ -0,0 +1,210 @@
...
+    param_combinations = list(product(*tuple(sweep_params.values())))
+    run_params = [
+        dict(zip(sweep_params.keys(), vals)) for vals in param_combinations
+    ]
+
+    # Create a record aggregator to collect records, and output table
+    record_aggregator = RecordAggregator()
+    table_headers = [
+        "Model", "Batch", "Concurrency",
+        PerfThroughput.header(), "Max " + GPUFreeMemory.header(),
```
Tabrizian, line 121: Can we refactor this to make it more generalizable? For example, the whole metrics that are being collected depends on a list that we can simply add/remove to it to change the metrics being monitored.

```
monitoring_metrics = [GPUFreeMemory, GPUUsedMemory, GPUUtilization]

# The rest of the code will depend on this array...
```

model_analyzer/device/gpu_device_factory.py
```diff
@@ -89,8 +89,7 @@ def create_device_by_cuda_index(index):
         cuda_device = devices[index]
         device_identity = cuda_device.get_device_identity()
         device_bus_id = '{:02x}:{:02x}:{:02x}'.format(
```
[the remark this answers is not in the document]
xprotobeast2: This isn't code I wrote, so maybe I can change this in a separate cleanup PR?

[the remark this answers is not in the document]
dzier: You have touched the file, it is OK to clean up the code a little bit here and there. It is called the "Boy Scout" rule. Leave every piece a code a little bit cleaner than you found it. This way we tackle technical debt a little bit at a time. A small change like this is OK. If you want, make sure it stays as a separate commit.  Before merging, do an interactive rebase on your changes to create the 1-2 commits. Both can be in the same PR.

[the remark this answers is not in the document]
Tabrizian: I fixed this in #23.

model_analyzer/entrypoint.py
```diff
@@ -0,0 +1,340 @@
...
+    return aggregated_records
+
+
+def main():
+    """
+    Main entrypoint of model_analyzer
+    """
+
+    # Get args and set monitoring metrics
+    args = CLI().parse(sys.argv)
```
dzier, line 269: You are still mixing levels of abstraction and main is still too long. Try to further extract methods so that main is less than 15 lines of code.  Off the top of my head, I can see something like....
```
def main():
   ... = initialize_model_analyzer(...)
   ... = start_triton(...)
   ... = profile_models(...)
   ... = stop_triton(...)
   ... = write_reports(...)
```

Note how in five lines of code, I clearly see the top level business logic of what ModelAnalyzer is doing. Everything is at the same high level of abstraction. And you don't need "section comments".  

Note that if you find yourself needing  a comment to explain what the next block of code does, then you should  extract a method.

dzier: Also note that you create several methods above. Considering grouping methods together that share the same purpose and creating a class to contain them.  This simple refactoring is know as Extract Class (https://refactoring.guru/extract-class).

dzier (changes requested):

model_analyzer/analyzer.py
```diff
@@ -0,0 +1,227 @@
...
+            perf_analyzer to the output stream/file
+        """
+
+        dcgm_monitor = DCGMMonitor(self.monitoring_interval, self.dcgm_tags)
+        perf_analyzer = PerfAnalyzer(
+            path=self.perf_analyzer_path,
+            config=self._create_perf_config(run_config))
+
+        # Get metrics for model inference and write perf_output
+
```
Tabrizian, line 132: Remove space

model_analyzer/analyzer.py
```diff
@@ -0,0 +1,227 @@
...
+        ----------
+        run_config : dict
+            The keys are arguments to perf_analyzer
+            The values are their values
+        write_perf_output : boolean
+            Whether or not to write the output from
+            perf_analyzer to the output stream/file
+        """
+
+        dcgm_monitor = DCGMMonitor(self.monitoring_interval, self.dcgm_tags)
```
Tabrizian, line 126: You may want to put a try-catch here or add a `Raises` to the docstring.

model_analyzer/analyzer.py
```diff
@@ -0,0 +1,227 @@
...
+        dcgm_monitor.start_recording_metrics()
+        if perf_analyzer:
+            try:
+                perf_records = perf_analyzer.run(self.perf_tags)
+            except FileNotFoundError as e:
+                raise TritonModelAnalyzerException(
+                    f"perf_analyzer binary not found : {e}")
+        else:
+            perf_records = []
+            time.sleep(0.1)
```
Tabrizian, line 193: Why should we sleep here?

model_analyzer/analyzer.py
```diff
@@ -0,0 +1,227 @@
...
+        self.monitoring_interval = args.monitoring_interval
+        self.monitoring_metrics = monitoring_metrics
+        self.writer = writer
+        self.param_headers = ['Model', 'Batch', 'Concurrency']
+
+        # Separates metric tags into perf_analyzer related and DCGM related tags
+        self.dcgm_tags = []
+        self.perf_tags = []
+        for metric in self.monitoring_metrics:
+            if metric in [GPUFreeMemory, GPUUsedMemory]:
```
Tabrizian, line 72: I think it would be great if we use `list(DCGMMonitor.model_analyzer_to_dcgm_field)` so that we wouldn't have to update this list every time we add a new metric.

model_analyzer/analyzer.py
xprotobeast2: I ran into an issue where if I try to run 
``` 
dcgm_monitor.start_recording_metrics()
dcgm_monitor.stop_recording_metrics()
```
No metrics are collected because the monitoring thread is inactive when `stop_recording_metrics` is called. This triggers a TritonModelAnalyzerException [here](https://github.com/triton-inference-server/model_analyzer/blob/6b4f68469e198ce4959da9433c82929ca6903092/model_analyzer/monitor/monitor.py#L108)

Tabrizian: I'll look into it. It looks like a concurrency issue.

xprotobeast2: Sounds good.

model_analyzer/analyzer.py
```diff
@@ -0,0 +1,260 @@
...
+        self.base_duration = args.base_duration
+        self.monitoring_interval = args.monitoring_interval
+        self.monitoring_metrics = monitoring_metrics
+        self.param_headers = ['Model', 'Batch', 'Concurrency']
+
+        # Separates metric tags into perf_analyzer related and DCGM related tags
+        self.dcgm_tags = []
+        self.perf_tags = []
+        for metric in self.monitoring_metrics:
+            if metric in list(DCGMMonitor.model_analyzer_to_dcgm_field.keys()):
```
Tabrizian, line 70: In Python 3 `.keys()` is not required and `list(dictionary)` only gives you the keys. I think this is fine too.

Tabrizian (dismissed):

model_analyzer/analyzer.py
xprotobeast2: I didn't know this. Will fix it.

Tabrizian (dismissed):

dzier (approved):

PR #23, opened 2020-11-05
Fix minor bug in DCGM monitor

Tabrizian:


model_analyzer/monitor/dcgm/dcgm_monitor.py
```diff
@@ -104,16 +104,18 @@ def _collect_records(self):
...
             metrics = self.group_watcher.values[device_id]
-
             # Find the first key in the metrics dictionary to find the
             # dictionary length
             first_key = list(metrics)[0]
             num_metrics = len(metrics[first_key].values)
             for i in range(num_metrics):
                 for tag in self._tags:
                     dcgm_field = self.MODEL_ANALYZER_TO_DCGM_FIELD[tag]
+                    # DCGM timestamps are in nano seconds, Record's timestamps
```
dzier, line 114: If a comment come after a line of code, there should be an empty line before it. The exception is if there it is the first line in a code block (i.e. the start of a for loop)

model_analyzer/monitor/dcgm/dcgm_monitor.py
```diff
@@ -104,16 +104,18 @@ def _collect_records(self):
...
+                    # are of type second.
                     records.append(
-                        tag(gpu, metrics[dcgm_field].values[i], time.time()))
+                        tag(gpu, float(metrics[dcgm_field].values[i].value),
+                            metrics[dcgm_field].values[i].ts / 1e6))
```
dzier, line 118: This should not be converted here. Keep data as raw as possible until otherwise needed. For example, if nanoseconds is not understandable when printing to screen, then do the conversion to seconds just before printing, not as you are recording the raw data.

dzier (changes requested):

model_analyzer/record/record.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, value, timestamp):
         timestamp : float
             Elapsed time from start of program
         """
+        assert type(value) is float
```
dzier, line 45: Why is this a float? If the value is nanoseconds, it should not be a float.

model_analyzer/record/record.py
```diff
@@ -42,6 +42,8 @@ def __init__(self, value, timestamp):
...
+        assert type(timestamp) is float
```
dzier, line 46: Timestamp should also not be a float.

dzier (changes requested):

model_analyzer/record/record.py
Tabrizian: Previously it was in seconds and we used precision to represent everything after the decimal point to represent ms. I will fix this.

xprotobeast2 (dismissed):

model_analyzer/record/record.py
```diff
@@ -40,8 +40,10 @@ def __init__(self, value, timestamp):
         value : float
             The value of the GPU metrtic
         timestamp : float
-            Elapsed time from start of program
+            Elapsed time from start of program in nanoseconds
         """
+        assert type(value) is float
```
xprotobeast2, line 45: We can allow `int`s here also as discussed.

xprotobeast2 (changes requested):

model_analyzer/record/record.py
```diff
@@ -37,11 +37,13 @@ def __init__(self, value, timestamp):
         """
         Parameters
         ----------
-        value : float
+        value : float or int
             The value of the GPU metrtic
         timestamp : float
-            Elapsed time from start of program
+            Elapsed time from start of program in nanoseconds
```
xprotobeast2, line 43: The doc strings for all derived classes may need to be updated as well.

tests/test_dcgm_monitor.py
```diff
@@ -49,11 +50,14 @@ def test_record_memory(self):
         # Assert instance types
         for record in records:
             self.assertIsInstance(record.device(), GPUDevice)
+            self.assertIsInstance(record.value(), float)
+            self.assertIsInstance(record.timestamp(), int)
```
xprotobeast2, line 54: Wondering if these tests should be done here or in a different test that just tests the constructors of all Records. Looks good for now.

xprotobeast2 (changes requested):

xprotobeast2 (approved):

dzier (approved):

PR #24, opened 2020-11-06
Mock perf_client and triton server for model analyzer unit tests.

xprotobeast2:


tests/mock_client.py
```diff
@@ -0,0 +1,103 @@
...
+
+class MockInferenceServerException(Exception):
+    """
+    A mock object that can be used in
+    place of an InferenceServerException
+    """
+    pass
+
+
+class MockTritonClientMethods:
```
dzier, line 40: Mock objects usually inherit from a shared base class or interface to the actual object. That way the test and end use will not know the difference.

dzier (changes requested):

tests/mock_client.py
xprotobeast2: But these are unit tests that test the objects themselves. The things I am mocking are external APIs or modules. I think I've removed most of the dependency of the test with the mock object.

dzier (approved):

PR #31, opened 2020-11-18
Make model analyzer deployable with kubernetes

xprotobeast2:
* Update helm charts
* Some test fixes
* Client bug fixes

### DIFF
--- a/helm-chart/Chart.yaml
+++ b/helm-chart/Chart.yaml
@@ -1,19 +1,31 @@
-# Copyright 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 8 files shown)


helm-chart/Chart.yaml
```diff
@@ -1,19 +1,31 @@
-# Copyright 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
dzier, line 1: Leave the old license. We need to convert everything back to Apache 2.

helm-chart/values.yaml
```diff
@@ -1,18 +1,30 @@
-# Copyright 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
dzier, line 1: Leave the original license

dzier (changes requested):

helm-chart/values.yaml
xprotobeast2: Hmm Then should we do PR that only changes the licenses back to the apache?

dzier (approved):

PR #35, opened 2020-11-19
Add default value for http and grpc endpoints

Tabrizian:


model_analyzer/cli/cli.py
```diff
@@ -116,16 +116,19 @@ def _add_arguments(self):
         )
         self._parser.add_argument(
             '--triton-version',
+            default='20.11-py3',
```
dzier, line 119: This is fine for now, but I would prefer not to have to maintain this for every release.

dzier (approved):

PR #47, opened 2020-12-02
Add automatic port allocation

Tabrizian:


qa/common/util.sh
```diff
@@ -210,3 +210,31 @@ function check_exported_metrics() {
...
+    export ports=()
+    for i in `seq 1 $num_port`; do
+        current_port=$((10000 + $RANDOM % 10000))
+        while [ `netstat -ano tcp |& grep :$current_port > /dev/null 2>&1` ]; do
+            current_port=$(echo "$port + 1" | bc)
+        done
+        ports=("${ports[@]}" "$current_port")
+    done
+    echo ${ports[@]}
+}
```
xprotobeast2, line 240: newline

model_analyzer/entrypoint.py
```diff
@@ -93,12 +97,21 @@ def get_server_handle(args):
     elif args.triton_launch_mode == 'docker':
         triton_config = TritonServerConfig()
         triton_config['model-repository'] = args.model_repository
+        http_port = args.triton_http_endpoint.split(':')[-1]
```
xprotobeast2, line 100: You can just load them into the server config right? Do you need the dict here?

model_analyzer/triton/server/server_docker.py
```diff
@@ -82,10 +85,15 @@ def start(self):
...
+        # Docker ports
+        docker_http_port = self._docker_ports['http_port'] or 8000
+        docker_grpc_port = self._docker_ports['grpc_port'] or 8001
+        docker_metrics_port = self._docker_ports['metrics_port'] or 8002
+
         ports = {
-            server_http_port: LOCAL_HTTP_PORT,
-            server_grpc_port: LOCAL_GRPC_PORT,
-            server_metrics_port: LOCAL_METRICS_PORT
+            server_http_port: docker_http_port,
```
xprotobeast2, line 94: Because we set both to the ports coming in from args.xxx_endpoint, we might as well just tell triton to use the free ports.

xprotobeast2 (changes requested):

model_analyzer/entrypoint.py
Tabrizian: Yes, loading it there will lead to changing the `tritonserver` port inside the container which is not what we want.

xprotobeast2: Can we not change the `tritonserver` port to match free ports?

model_analyzer/triton/server/server_docker.py
Tabrizian: Because I didn't change the `TritonConfig` in the caller function `server_grpc_port` will be equal to 8000 always. I think that is Ok because we have the full control of the applications inside the container. I can set the ports inside the container to `8000,8001,8002` always so that the code is more readable.

model_analyzer/entrypoint.py
Tabrizian: You're right I can just change the server config and use the same values for both inside and outside the container.

xprotobeast2 (approved):

PR #50, opened 2020-12-03
Build and install pip wheel during docker build

xprotobeast2:


Tabrizian (approved):
I guess the same issue with the version exists in the Triton client build.

Dockerfile
```diff
@@ -15,7 +15,7 @@
 ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:20.11-py3
 ARG TRITONSDK_BASE_IMAGE=nvcr.io/nvidia/tritonserver:20.11-py3-clientsdk
 
-ARG MODEL_ANALYZER_VERSION=1.0.0dev
+ARG MODEL_ANALYZER_VERSION=1.0.0.dev0
```
dzier, line 18: Just `dev`, do not add the 0. The build_wheel.sh will add that automatically.

VERSION
```diff
@@ -1 +1 @@
-1.0.0dev
+1.0.0.dev0
```
dzier, line 1: No, just `dev`

dzier: revert this change

Dockerfile
dzier: revert this change.

dzier (changes requested):

Tabrizian (dismissed):

Dockerfile
```diff
@@ -16,12 +16,14 @@ ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:20.11-py3
 ARG TRITONSDK_BASE_IMAGE=nvcr.io/nvidia/tritonserver:20.11-py3-clientsdk
 
 ARG MODEL_ANALYZER_VERSION=1.0.0dev
+ARG MODEL_ANALYZER_WHEEL_VERSION=1.0.0.dev0
```
dzier, line 19: No, we cannot do this. Why does the old one above not work, when we can get it to work in all other cases?  I absolutely do not want to VERSION strings.

dzier (changes requested):

Dockerfile
xprotobeast2: pip takes `1.0.0dev` and turns it into `1.0.0.dev0`. I added this only so that we can use this arg in line 55.

dzier (approved):

PR #54, opened 2020-12-14
Add per GPU metrics in the output table

Tabrizian:
Before
```
Server Only:
Model                         Batch   Concurrency   Throughput(infer/sec)   Max GPU Utilization(%)   Max GPU Used Memory(MB)   Max GPU Free Memory(MB)  
triton-server                 0       0             0                       0.0                      228.0                     23991.0                  

Models:
Model                         Batch   Concurrency   Throughput(infer/sec)   Max GPU Utilization(%)   Max GPU Used Memory(MB)   Max GPU Free Memory(MB)  
classification_chestxray_v1   1       1             49.2                    37.0                     5186.0                    23359.0                  
```

After
```
Server Only:
Model           Batch   Concurrency   Max GPU Utilization(%)   Max GPU Used Memory(MB)   Max GPU Free Memory(MB)   GPU ID  
triton-server   0       0             0.0                      319.0                     15961.0                   0       
triton-server   0       0             0.0                      319.0                     15961.0                   1       
triton-server   0       0             0.0                      319.0                     15961.0                   2       
triton-server   0       0             0.0                      319.0                     15961.0                   3       

Models (GPU Metrics):
Model                         Batch   Concurrency   Max GPU Utilization(%)   Max GPU Used Memory(MB)   Max GPU Free Memory(MB)   GPU ID  
classification_chestxray_v1   1       1             66.0                     4765.0                    15439.0                   0       
classification_chestxray_v1   1       1             54.0                     4729.0                    15439.0                   1       
classification_chestxray_v1   1       1             65.0                     4729.0                    15439.0                   2       
classification_chestxray_v1   1       1             64.0                     4729.0                    15439.0                   3       

Models (Inference):
Model                         Batch   Concurrency   Throughput(infer/sec)  
classification_chestxray_v1   1       1             56.0 
```

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -14,8 +14,10 @@
 
 import time
 import logging
+from collections import defaultdict
 
-from model_analyzer.model_analyzer_exceptions import TritonModelAnalyzerException
+from model_analyzer.model_analyzer_exceptions \
+    import TritonModelAnalyzerException
 from .monitor.dcgm.dcgm_monitor import DCGMMonitor
 from .perf_analyzer.perf_analyzer import PerfAnalyzer
 from .perf_analyzer.perf_config import PerfAnalyzerConfig
@@ -37,7 +39,6 @@ class Analyzer:
     exposes profiling and result writing
     methods.
     """
-
     def __init__(self, args, monitoring_metrics):
         """
         Parameters
@@ -64,9 +65,10 @@ class Analyzer:
             elif metric in PerfAnalyzer.perf_metrics:
                 self.perf_tags.append(metric)
 
-        self.tables = {
-            "Server Only:": self._create_output_table(),
-            "Models:": self._create_output_table()
+        self._tables = {
+            'model_gpu_metrics': self._create_gpu_output_table('Models (GPU Metrics)'),
+            'server_gpu_metrics': self._create_gpu_output_table('Server Only'),
+            'model_inference_metrics': self._create_inference_output_table('Models (Inference)')
         }
 
     def profile_server_only(self, default_value='0'):
@@ -87,19 +89,21 @@ class Analyzer:
         logging.info('Profiling server only metrics...')
         dcgm_monitor = DCGMMonitor(self._gpus, self._monitoring_interval,
                                    self.dcgm_tags)
-        server_only_metrics = self._profile(perf_analyzer=None,
-                                            dcgm_monitor=dcgm_monitor)
+        server_only_gpu_metrics, _ = self._profile(perf_analyzer=None,
+                                                   dcgm_monitor=dcgm_monitor)
 
-        # Model name here is triton-server, batch and concurrency are defaults
-        output_row = ['triton-server', default_value, default_value]
+        gpu_metrics = defaultdict(list)
+        for _, metric in server_only_gpu_metrics.items():
+            for gpu_id, metric_value in metric.items():
+                gpu_metrics[gpu_id].append(metric_value)
 
-        # add the obtained metrics
-        for metric in self._monitoring_metrics:
-            if metric in server_only_metrics:
-                output_row.append(server_only_metrics[metric])
-            else:
-                output_row.append(default_value)
-        self.tables["Server Only:"].add_row(output_row)
+        for gpu_id, metric in gpu_metrics.items():
+            # Model name here is triton-server, batch and concurrency
+            # are defaults
+            output_row = ['triton-server', default_value, default_value]
+            output_row += metric
+            output_row.append(gpu_id)
+            self._tables['server_gpu_metrics'].add_row(output_row)
         dcgm_monitor.destroy()
 
     def profile_model(self, run_config, perf_output_writer=None):
@@ -129,20 +133,38 @@ class Analyzer:
             config=self._create_perf_config(run_config))
 
         # Get metrics for model inference and write perf_output
-        model_metrics = self._profile(perf_analyzer=perf_analyzer,
-                                      dcgm_monitor=dcgm_monitor)
+        model_gpu_metrics, inference_metrics = self._profile(
+            perf_analyzer=perf_analyzer, dcgm_monitor=dcgm_monitor)
         if perf_output_writer:
             perf_output_writer.write(perf_analyzer.output() + '\n')
 
+        # Process GPU Metrics
+        gpu_metrics = defaultdict(list)
+        for _, metric in model_gpu_metrics.items():
+            for gpu_id, metric_value in metric.items():
+                gpu_metrics[gpu_id].append(metric_value)
+
+        for gpu_id, metrics in gpu_metrics.items():
+            # Model name here is triton-server, batch and concurrency
+            # are defaults
+            output_row = [
+                run_config['model-name'], run_config['batch-size'],
+                run_config['concurrency-range']
+            ]
+            output_row += metrics
+            output_row.append(gpu_id)
+            self._tables['model_gpu_metrics'].add_row(output_row)
+
+        # Process Inference Metrics
         output_row = [
             run_config['model-name'], run_config['batch-size'],
             run_config['concurrency-range']
         ]
-        output_row += [
-            model_metrics[metric] for metric in self._monitoring_metrics
-        ]
 
-        self.tables["Models:"].add_row(output_row)
+        output_row += [
+            inference_metrics[metric] for metric in inference_metrics
+        ]
+        self._tables['model_inference_metrics'].add_row(output_row)
 
         dcgm_monitor.destroy()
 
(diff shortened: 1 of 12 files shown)


model_analyzer/analyzer.py
```diff
@@ -129,20 +133,38 @@ def profile_model(self, run_config, perf_output_writer=None):
...
+
+        # Process Inference Metrics
         output_row = [
             run_config['model-name'], run_config['batch-size'],
             run_config['concurrency-range']
         ]
+
         output_row += [
-            model_metrics[metric] for metric in self._monitoring_metrics
+            inference_metrics[metric] for metric in inference_metrics
```
xprotobeast2, line 165: `output_row += inference_metrics.values()`  will make it clear that you're getting the values from a dict

model_analyzer/analyzer.py
```diff
@@ -266,25 +300,45 @@ def _profile(self, perf_analyzer, dcgm_monitor):
...
-        return record_aggregator.aggregate()
 
-    def _create_output_table(self, aggregation_tag='Max'):
+        def gpu_device_id(record):
+            return record.device().device_id()
+
+        records_groupby_gpu = {}
+        for tag in self.dcgm_tags:
+            records_groupby_gpu[tag] = record_aggregator.groupby(
+                tag, gpu_device_id)
```
xprotobeast2, line 312: For single use, maybe a lambda would be preferable to nested def. 
```
for tag in self.dcgm_tags:
    records_groupby_gpu[tag] = record_aggregator.groupby(
         tag, (lambda record : record.device().device_id()))
```
I suppose this is stylistic. I think nested defs are recommended when there's repeated use but in this case it is use and throw right?

model_analyzer/record/record_aggregator.py
```diff
@@ -63,16 +78,23 @@ def filter_records(self, record_types=None, filters=None):
 
         Returns
         -------
-        Dict of List of Records
-            keys are record_types, values
-            are all records of that type
+        RecordAggregator
```
xprotobeast2, line 81: If this function needs to have this type of functionality, we should think about making it static.

model_analyzer/record/record_aggregator.py
```diff
@@ -87,14 +109,52 @@ def filter_records(self, record_types=None, filters=None):
...
-            filtered_records[h] = [
-                record for record in self._records[h] if f(record.value())
-            ]
+            for record in self._records[h]:
+                if f(record):
+                    filtered_records.insert(record)
 
         return filtered_records
 
+    def groupby(self, record_type, groupby_criteria, reduce_func=max):
```
xprotobeast2, line 119: perhaps this one can be static too?

xprotobeast2 (changes requested):

model_analyzer/record/record_aggregator.py
Tabrizian: The goal with this was to make the return types such that the return values can be chained together.

model_analyzer/analyzer.py
Tabrizian: Right, good suggestion!

model_analyzer/record/record_aggregator.py
```diff
@@ -40,10 +41,24 @@ def insert(self, record):
...
+
+        Parameters
+        ----------
+        record_type : Record
+            record_type to add to the records.
+        records : list
+            List of new records to be added.
+        """
+
+        self._records[record_type] = records
```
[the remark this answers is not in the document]
Tabrizian: The goal here was to completely replace a key with a new key. It was not for appending to it.

model_analyzer/analyzer.py
```diff
@@ -266,25 +301,42 @@ def _profile(self, perf_analyzer, dcgm_monitor):
...
         # Insert all records into aggregator and get aggregated DCGM records
         record_aggregator = RecordAggregator()
-        for record in perf_records + dcgm_records:
+        for record in dcgm_records:
             record_aggregator.insert(record)
-        return record_aggregator.aggregate()
 
-    def _create_output_table(self, aggregation_tag='Max'):
+        records_groupby_gpu = {}
+        for tag in self.dcgm_tags:
```
xprotobeast2, line 308: If we move this loop into `RecordAggregator.groupby` we can turn this into
```
records_groupby_gpu = RecordAggregator.groupby(
                             record_aggregator, 
                             self.dcgm_tags, 
                             (lambda record: record.device().device_id()))
```

model_analyzer/record/record_aggregator.py
```diff
@@ -160,3 +220,23 @@ def aggregate(self, record_types=None, reduce_func=max):
                 values.append(record.value())
             aggregated_records[record_type] = reduce_func(values)
         return aggregated_records
+
+    def get_records(self):
```
xprotobeast2, line 224: can we add a parameter/option to get all records of a specific type/ list of types?
```
def get_records(self, record_types):
     # Construct a subset of self._records
```

I'm not sure if this is a workable idea as the calling code needs to be aware of the types.

model_analyzer/record/record_aggregator.py
xprotobeast2: Is it okay that a member function returns a whole new aggregator? I agree that we do need filter_records to return a `RecordAggregator` however,  could this be achieved just by passing in a RecordAggregator?
```
"""
@staticmethod
def filter_records(aggregator, record_types=None, filters=None):
        """
        Get records that satisfy the given list of criteria.
        Parameters
        ----------
        aggregator : The RecordAggregator
            containing records to be filtered
        record_types : list of types of Records
            the types of the records we are
            imposing the filter criteria on.
        filters : list of callables
            conditions that determine whether
            a given record should be returned.
            If no filters specified, all records
            of types specified by record_types will be
            returned.
            Note : This must be of the same length
                   as the list of record_types, or omitted.
        Returns
        -------
        RecordAggregator
            Returns a new RecordAggregator containing the filtered
            records
        """
        
        input_records = aggregator.get_records()
        filtered_records = RecordAggregator()
        if not record_types and not filters:
            for record_type, records in input_records.items():
                filtered_records.add_key(record_type, records)
            return filtered_records

        if record_types and not filters:
            try:
                for record_type in record_types:
                    filtered_records.add_key(record_type,
                                             input_records[record_type])
                return filtered_records
            except KeyError as k:
                    f"Record type '{k.header()}' not found in this RecordAggregator"
                )
        if filters and not record_types:
                "Must specify the record types corresponding to each filter criterion."
            )
        if len(record_types) != len(filters):
                "Must specify the same number of record types as filter criteria."
            )

        # Remove records that do not satisfy criteria
        for h, f in zip(record_types, filters):
            for record in input_records[h]:
                if f(record):
                    filtered_records.insert(record)

        return filtered_records
```

xprotobeast2 (changes requested):

model_analyzer/record/record_aggregator.py
Tabrizian: I agree that returning a whole new aggregator is not a good solution. I can add variants of these member functions that perform the filtering in place but I think making them static and passing the object as an argument may break the encapsulations that we have.

model_analyzer/record/record_aggregator.py
Tabrizian: Yeah, I guess we can add this later if we found a place where this was required.

model_analyzer/analyzer.py
Tabrizian: I fixed this in the new commit.

model_analyzer/record/record_aggregator.py
```diff
@@ -87,14 +109,52 @@ def filter_records(self, record_types=None, filters=None):
...
+        for field_value in field_values:
+            aggregated_result = self.filter_records(
+                record_types=[record_type],
+                filters=[
+                    lambda record: groupby_criteria(record) == field_value
+                ]).aggregate(record_types=[record_type],
+                             reduce_func=reduce_func)
+            groupby_result[field_value] = aggregated_result[record_type]
+
+        return groupby_result
```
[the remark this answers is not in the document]
Tabrizian: Fixed in the new commit.

xprotobeast2 (dismissed):

dzier:
I think the GPU ID should be the second column instead of the last, since it is really part of the run identifier and not a resulting metric.

xprotobeast2 (approved):

Tabrizian:
> I think the GPU ID should be the second column instead of the last, since it is really part of the run identifier and not a resulting metric.
```
Models (GPU Metrics):
Model             GPU ID   Batch   Concurrency   Max GPU Utilization(%)   Max GPU Used Memory(MB)   Max GPU Free Memory(MB)  
vgg_16_graphdef   0        2       1             93.0                     7628.0                    23383.0                  

Server Only:
Model           GPU ID   Batch   Concurrency   Max GPU Utilization(%)   Max GPU Used Memory(MB)   Max GPU Free Memory(MB)  
triton-server   0        0       0             0.0                      228.0                     23991.0                  

Models (Inference):
Model             Batch   Concurrency   Throughput(infer/sec)  
vgg_16_graphdef   2       1             55.2 

```

@dzier

PR #57, opened 2021-01-05
Added CPU memory usage to model-analyzer metrics

xprotobeast2:
```
Server Only:

Models (GPU Metrics):
classification_chestxray_v1   0        1       1             5215.0                     23332.0                        41.0                    
classification_chestxray_v1   0        2       1             5215.0                     19004.0                        40.0                    
classification_breast_v1      0        1       1             5215.0                     19004.0                        13.0                    
classification_breast_v1      0        2       1             5215.0                     19004.0                        14.0                    

Models (Inference):
classification_chestxray_v1   1       1             47.2                    20240.0               5467.0              28483.0                
classification_chestxray_v1   2       1             76.8                    23908.0               5864.0              26644.0                
classification_breast_v1      1       1             361.0                   2380.0                5838.0              26422.0                
classification_breast_v1      2       1             702.0                   2482.0                5724.0              26537.0          

```


Tabrizian (changes requested):

model_analyzer/triton/server/server_docker.py
```diff
@@ -141,3 +141,15 @@ def logs(self):
         """
 
         return b''.join(list(self._tritonserver_log_gen)).decode("utf-8")
+
+    def cpu_stats(self):
+        """
+        Returns the CPU memory usage and CPU available memory in MB
+        """
+
+        stats = self._tritonserver_container.stats(stream=False)
```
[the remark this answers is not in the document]
xprotobeast2: Getting almost identical values. Weirdly they are less than when tritonserver runs outside the container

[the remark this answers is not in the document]
Tabrizian: That is really interesting.

model_analyzer/analyzer.py
```diff
@@ -299,20 +320,26 @@ def _profile(self, perf_analyzer, dcgm_monitor):
...
             time.sleep(self._duration_seconds)
         dcgm_records = dcgm_monitor.stop_recording_metrics()
+        cpu_records = cpu_monitor.stop_recording_metrics()
 
         # Insert all records into aggregator and get aggregated DCGM records
         record_aggregator = RecordAggregator()
         for record in dcgm_records:
             record_aggregator.insert(record)
 
+        cpu_record_aggregator = RecordAggregator()
```
dzier, line 330: Why can't we use the same Record Aggregator here?

model_analyzer/entrypoint.py
```diff
@@ -33,6 +33,8 @@
 from .record.gpu_free_memory import GPUFreeMemory
 from .record.gpu_used_memory import GPUUsedMemory
 from .record.gpu_utilization import GPUUtilization
+from .record.cpu_used_memory import CPUUsedMemory
```
dzier, line 36: All of these imports are creating huge dependency chains. I shouldn't need to edit this top level file if I create a new record type. All record types should use the same interface and this top level file should reference only the single interface. Not the individual instances of the records should be abstracted away.

dzier (changes requested):

model_analyzer/entrypoint.py
xprotobeast2: I think this will likely move into the input parsing side of the code, because we want the user to control what metrics they measure. However, I don't see how we can implement this without some dispatching code where all of these imports are in the same place.

model_analyzer/analyzer.py
xprotobeast2: Wanted to make explicit the where the cpu records are coming from in `Analyzer.profile_model` but I agree it can be combined with `perf_record_aggregator`. I will fix that.

model_analyzer/entrypoint.py
Tabrizian: We can probably import all the records. For example, we can use `from .record import *` so that we do not need to update this list every time. But I guess still the dependencies will exist.

xprotobeast2: I like the idea, though I think doing `import *` is recommended against in general right?

xprotobeast2: Additionally, this moves the problem to the line 
```
monitoring_metrics = [
        PerfThroughput, GPUUtilization, GPUUsedMemory, GPUFreeMemory,
        CPUUsedMemory, CPUFreeMemory
    ]
```

Tabrizian: I see...

model_analyzer/analyzer.py
dzier: how/where is cpu_record_aggregator consumed/used?

model_analyzer/entrypoint.py
dzier: It is not the responsibility of the top level application to know what the monitoring metrics are. What you needs is a Object Constructor design pattern. A class that constructs each type and breaks the dependency on the individual type.

model_analyzer/analyzer.py
xprotobeast2: It was previously aggregated and returned from the function.

model_analyzer/entrypoint.py
xprotobeast2: A single class per type or a factory?

model_analyzer/record/metrics_mapper.py
```diff
@@ -0,0 +1,57 @@
...
+from .perf_latency import PerfLatency
+
+
+class MetricsMapper:
+    """
+    Class maps metrics to record types
+    """
+
+    tag_to_metric = {
+        "Throughput": PerfThroughput,
```
Tabrizian, line 30: I think there is no need to make the key human readable. Can we make all the keys lower case and use underscore instead of spaces?

model_analyzer/record/metrics_mapper.py
```diff
@@ -0,0 +1,57 @@
...
+        ----------
+        tags : list of str
+            Human readable names for the 
+            metrics to monitor. They correspond
+            to actual record types.
+        """
+
+        self._tags = tags
+
+    def get_monitoring_metrics(self):
```
Tabrizian, line 51: Also, I think it would be great to make this class a semi-factor class that doesn't need to be instantiated. How about making this method static and getting an argument for the list of metrics?

model_analyzer/record/metrics_mapper.py
```diff
@@ -0,0 +1,57 @@
...
+from .perf_throughput import PerfThroughput
+from .perf_latency import PerfLatency
+
+
+class MetricsMapper:
+    """
+    Class maps metrics to record types
+    """
+
+    tag_to_metric = {
```
Tabrizian, line 29: We can remove the need for `tag_to_metric` array using a code like below:

```py
import abc
import importlib


class BaseProvider(abc.ABCMeta):
    providers = {}

    def __new__(cls, name, bases, namespace):
        instance = abc.ABCMeta.__new__(cls, name, bases, namespace)

        if isinstance(instance.name, str):
            cls.providers[instance.name] = instance

        return instance

    @classmethod
    def get(cls, name):
        if name not in cls.providers:
            try:
                importlib.import_module('model_analyzer.record.%s' % name)
            except ImportError as e:
                print(e)
        return cls.providers[name]


class Provider(metaclass=BaseProvider):
    def __init__(self):
        pass

    @property
    @abc.abstractmethod
    def name(self):
        raise NotImplemented()
```

This assumes that we have a name field for each of the records.

model_analyzer/monitor/cpu_monitor.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+
+        return self._cpu_memory_records
+
+    def destroy(self):
+        """
+        Cleanup threadpool resources
+        """
+
+        self._thread_pool.terminate()
```
Tabrizian, line 73: Also, I think it would be great if we can move these lines to the parent class.

Tabrizian (changes requested):

model_analyzer/record/metrics_mapper.py
xprotobeast2: The idea, was that eventually you'd be able to select metrics from the config file in which case it should be human readable no?

model_analyzer/record/metrics_mapper.py
xprotobeast2: The name field being equal to the filename right? This seems to require that the class is hardcoded with the name of the file/module.

model_analyzer/record/metrics_mapper.py
xprotobeast2: Agreed.

model_analyzer/record/metrics_mapper.py
xprotobeast2: I'm not sure how we would modify the metrics we measure using this method either?

model_analyzer/record/metrics_mapper.py
Tabrizian: Yes, I think we can do that even with the small cases. But it is not that important.

model_analyzer/record/metrics_mapper.py
Tabrizian: Yes exactly, we just need to add a name to the field to the subclasses that is equal to the file name. I can do this too if it looks good to you.

xprotobeast2: This means that the config options for the metrics will have to be translated into the filenames as well, or we would have to use the filenames themselves to configure which metrics are being measured.

Tabrizian: Yes, we would have to use the filenames. I think our filenames look good but it can be subjective 😆

xprotobeast2: Done

Tabrizian: Looks great!

Tabrizian (approved):

model_analyzer/analyzer.py
```diff
@@ -324,16 +340,18 @@ def _create_inference_output_table(self, title, aggregation_tag='Max'):
         # Create headers
         table_headers = self._param_inference_headers[:]
         for metric in self._monitoring_metrics:
-            if metric not in self.dcgm_tags:
+            if metric in self.perf_metrics:
```
dzier, line 343: Let's not do this. This is a double search here. You are looping through the list of metrics and then looping through a second list for a potential O(n^2) implementation. This is bad algorithmically. 

Ideally, every metric should know how to print out the headers, you do not need to distinguish anything. For example, if``aggregation_tag is needed by cpu_metrics, then it should added in the `header()` method for cpu_metrics derivation. 

Try to re-write to avoid the if-else all together.

model_analyzer/analyzer.py
```diff
@@ -324,16 +340,18 @@ def _create_inference_output_table(self, title, aggregation_tag='Max'):
...
         return OutputTable(headers=table_headers, title=title)
 
     def _create_gpu_output_table(self, title, aggregation_tag='Max'):
 
         # Create headers
         table_headers = self._param_gpu_headers[:]
         for metric in self._monitoring_metrics:
-            if metric in self.dcgm_tags:
+            if metric in self.dcgm_metrics:
                 table_headers.append(aggregation_tag + ' ' + metric.header())
```
dzier, line 355: The dcgm metric should know the aggregation tag if it needs to print it.

dzier (changes requested):

model_analyzer/analyzer.py
```diff
@@ -56,21 +55,26 @@ def __init__(self, args, monitoring_metrics):
         self._param_inference_headers = ['Model', 'Batch', 'Concurrency']
         self._param_gpu_headers = ['Model', 'GPU ID', 'Batch', 'Concurrency']
         self._gpus = args.gpus
+        self._server = server
+
+        # Separates metric list into perf_analyzer related and DCGM related lists
+        self.dcgm_metrics = []
```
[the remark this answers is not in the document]
xprotobeast2: I'm refactoring this class as part of the Comparator/RunConfig generator story. I will create a `MetricsMananger` class in that story.

model_analyzer/analyzer.py
xprotobeast2: This was done to preserve the order of the metrics in the original monitoring_metrics list, as well as to leave out the dcgm metrics from this table. What I can do is change it into 1 if statement just checking that its not a DCGM metric. Then I can convert the metric lists into sets, giving us O(1) lookup.

dzier: O(1) is better, but the point is to avoid RTTI (runtime type identification) all together. Anytime you see code that adds special behavior based on a "type" you are committing a code smell. Instead, you should derive child classes for each type that have the same method, so the caller doesn't need to know. There are several Design Patterns that can better address the situation above that are cleaner.

https://refactoring.guru/replace-type-code-with-subclasses

model_analyzer/monitor/dcgm/dcgm_monitor.py
```diff
@@ -71,14 +71,15 @@ def __init__(self, gpus, frequency, tags, dcgmPath=None):
 
         frequency = int(self._frequency * 1000)
         fields = []
-        for tag in tags:
-            if tag in self.model_analyzer_to_dcgm_field:
-                dcgm_field = self.model_analyzer_to_dcgm_field[tag]
+        for metric in metrics:
+            if metric in self.model_analyzer_to_dcgm_field:
+                dcgm_field = self.model_analyzer_to_dcgm_field[metric]
```
dzier, line 76: What is going on here? This has a code smell to it. It is almost like this should be a visitor, or something similar.

dzier (changes requested):

model_analyzer/analyzer.py
xprotobeast2: I think I can move the creation of the metrics tables into the MetricsManager. It will look more like a visitor pattern then.

model_analyzer/monitor/dcgm/dcgm_monitor.py
xprotobeast2: I think its checking to make sure that the metric is registered with the DCGM monitor. I will remove the if, and change it to a try/except

dzier (approved):

PR #58, opened 2021-01-12
Add Model Analyzer config

Tabrizian:
Added Model Analyzer Config.  Currently, supports all the flags supported by the current version of Model Analyzer. The priority is with the CLI and then with the Config file as mentioned in the engineering doc.

An example configuration looks like below:

```yaml
model_repository: /data/inferenceserver/20.11dev/tf_model_store/
export: true
triton_launch_mode: docker
model_names: vgg_16_graphdef
concurrency: 1
triton_http_endpoint: localhost:8030
triton_grpc_endpoint: localhost:8040
triton_metrics_url: http://localhost:8090/metrics
batch_sizes: 2
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,12 @@
 <!--
-Copyright 2020, NVIDIA CORPORATION. All rights reserved.
+Copyright (c) 2020-2021, NVIDIA CORPORATION. All rights reserved.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/model_analyzer/cli/cli.py
+++ b/model_analyzer/cli/cli.py
@@ -12,12 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import sys
 import logging
+import argparse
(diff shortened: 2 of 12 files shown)


xprotobeast2 (approved):

xprotobeast2 (commented):
Actually, one change I think would be nice is to call the Config class something specific to the whole model analyzer. Perhaps `AnalyzerConfig` or `ModelAnalyzerConfig` instead of just Config.

xprotobeast2 (approved):

README.md
```diff
@@ -1,9 +1,12 @@
 <!--
-Copyright 2020, NVIDIA CORPORATION. All rights reserved.
+Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
dzier, line 2: 2020-2021

model_analyzer/cli/cli.py
```diff
@@ -12,241 +12,64 @@
...
+        accordingly.
         """
 
         # Remove the first argument which is the program name
         args = self._parser.parse_args(sys.argv[1:])
-        self._preprocess_and_verify_arguments(args)
-        self._setup_logger(args)
-
-        return args
+        self._config.set_config_values(args)
```
dzier, line 75: How does this work. If I pass a config file AND command line arguments, do the CLI args override the settings in the config file? What if elements are missing from the config file, is there a default value?

Tabrizian: @dzier Yes, exactly. The CLI args have a higher priority, then config values, and then the default value. There is a test for this too. 
https://github.com/triton-inference-server/model_analyzer/pull/58/files#diff-82fdafeb870e92fb6c61f571ad81add4eb2173aa56f521175cab4139f3015baeR320-R330

dzier (approved):

PR #67, opened 2021-01-21
Update launch_modes.md

Tabrizian:


xprotobeast2 (approved):
Ah good catch, its unfortunate this paragraph even went into master in the first place. It was never correct.

dzier (approved):

PR #68, opened 2021-01-22
Add Model Analyzer configuration doc

Tabrizian:


docs/config.md
```diff
@@ -0,0 +1,159 @@
...
+* `<list>`: a list of values.
+* `<range>`: An object containing `start` and `end` key with an optional `step` value. The types that support `<range>` can be described by a list or using the example structure below:
+```yaml
+batch_sizes:
+    start: 2
+    end: 6
+    step: 2
+```
+
+This YAML represents the array `[2, 4]`. Note that `end` is not included.
```
dzier, line 41: Somehow, this doesn't make sense to me. If I have a start of 2, end of 6, and step of 2, I would expect this to run [2,4,6]. Why is `end` not included?

dzier: Also, have you considered specifying the range like... `[start:step:stop]`. For example a value of `[1:1:6]` would create an array of `[1,2,3,4,5,6]`, or `[3:5:15]` will generate `[3,8,13]`

docs/config.md
```diff
@@ -0,0 +1,159 @@
...
+    start: 4
+    end: 9
+concurrency:
+    - 2
+    - 4
+    - 8
+```
+
+If you save this file to the `config.yml`, you can start
+Model Analyzer using the `-f` flag.
```
dzier, line 154: Is there a long name for the CLI? May want to specify that as well.

dzier (changes requested):

docs/config.md
Tabrizian: Yes, we have a long name for the CLI too. It is `--config-file`. I will include this here too.

docs/config.md
Tabrizian: I will fix the config parsing to include the last value too. Regarding the alternative format, it would make it a bit difficult because of the way we parse configs (we parse configs according to the different types that they have) but I can rename  `end` to `stop` so that it is more intuitive.

dzier: Just a thing to think about in the future.

dzier (approved):

PR #76, opened 2021-01-28
Added Single Model constrained analysis without instance count/dynamic batching support

xprotobeast2:
Running the following config
```
model_repository: /qa_model_repository/
export: true
triton_launch_mode: docker
model_names: classification_chestxray_v1,classification_malaria_v1 
concurrency: 2,4
perf_measurement_window: 5000
triton_http_endpoint: localhost:8000
triton_grpc_endpoint: localhost:8001
triton_metrics_url: http://localhost:8002/metrics
batch_sizes: 16,32
no_perf_output: true
objectives:
    - perf_throughput
    - gpu_used_memory
constraints:
  perf_latency:
    max: 100000
```

Results in
```
Server Only:

Models (GPU Metrics):
classification_chestxray_v1   0        16      4             9695.0                     14524.0                        90.0                    
classification_chestxray_v1   0        32      2             9695.0                     14524.0                        93.0                    
classification_chestxray_v1   0        16      2             9695.0                     23332.0                        82.0                    
classification_malaria_v1     0        32      4             9695.0                     14524.0                        75.0                    
classification_malaria_v1     0        32      2             9695.0                     14524.0                        67.0                    
classification_malaria_v1     0        16      4             9695.0                     14524.0                        44.0                    
classification_malaria_v1     0        16      2             9695.0                     14524.0                        43.0                    

Models (Inference):
classification_chestxray_v1   16      4             556.8                   97359.0               5818.0              26544.0                
classification_chestxray_v1   32      2             550.4                   89657.0               5924.0              26377.0                
classification_chestxray_v1   16      2             499.2                   48139.0               5760.0              28308.0                
classification_malaria_v1     32      4             4742.4                  25130.0               6027.0              26317.0                
classification_malaria_v1     32      2             4281.6                  13095.0               6027.0              26330.0                
classification_malaria_v1     16      4             2822.4                  21213.0               6026.0              26325.0                
classification_malaria_v1     16      2             2499.2                  11539.0               6026.0              26334.0                

classification_chestxray_v1   0        32      4             9695.0                     14524.0                        92.0                    

classification_chestxray_v1   32      4             588.8                   188923.0              6003.0              26336.0                

```


Tabrizian (changes requested):
I reviewed one half of the PR. I will review the rest tomorrow.

model_analyzer/result/result_comparator.py
```diff
@@ -0,0 +1,161 @@
...
+                if value_diff > threshold:
+                    return 1
+                elif value_diff < -threshold:
+                    return -1
+            else:
+                raise TritonModelAnalyzerException(
+                    f"Category unknown for objective : '{priority}'")
+        return 0
+
+    def _average_measurements(self, result):
```
Tabrizian, line 128: Can we make this more general? For example, we can add a new argument `aggregator`, and rename the function to `_aggregate_measurements`. This new argument is a `callable` that will be called on the list of Records that we have here.

model_analyzer/result/result_manager.py
```diff
@@ -0,0 +1,274 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: 2021

model_analyzer/result/run_result.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: 2021

model_analyzer/triton/model/model_config.py
```diff
@@ -0,0 +1,144 @@
...
+            raise TritonModelAnalyzerException(
+                f"Model's name not found in ModelConfig at {self._model_path}")
+        return self._config['name']
+
+    def platform(self):
+        """
+        Returns
+        -------
+        str
+            The platform name of this model
```
Tabrizian, line 65: What is platform? Is it cpu/gpu?

model_analyzer/triton/model/model_config.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def dynamic_batching(self):
+        """
+        Returns
+        -------
+        dict
+            The dynamic batching parameters for this model
+        """
+
+        if 'dynamic_batching' not in self._config:
```
Tabrizian, line 110: What if a field is optional? It looks like all the fields have to be specified in the current setting.

model_analyzer/triton/model/model_config.py
```diff
@@ -0,0 +1,144 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: 2021

model_analyzer/triton/model/model_config_utils.py
```diff
@@ -0,0 +1,134 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: 2021

model_analyzer/triton/model/model_config_utils.py
```diff
@@ -0,0 +1,134 @@
...
+            config_str = parse_dict(config_str[dict_delimiter + 1:],
+                                    next_dict).lstrip()
+            output[next_item[:dict_delimiter].rstrip()] = next_dict
+        else:
+            break
+
+    return config_str
+
+
+def parse_list(config_str, output):
```
Tabrizian, line 91: Let's remove the functions related to parsing and use the code snippet below:
```
from google.protobuf import text_format as pbtf
from tritonclient.grpc.model_config_pb2 import ModelConfig


with open('../triton/python_backend/examples/add_sub/config.pbtxt') as f:
    x = f.read()
a = pbtf.Parse(x, ModelConfig())
```

model_analyzer/triton/model/model_config_utils.py
```diff
@@ -0,0 +1,134 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+
+def parse_model_config(model_path):
```
Tabrizian, line 18: What about moving this function to the ModelConfig and renaming it to a static factory method named `create_from_file`?

Tabrizian (changes requested):

model_analyzer/config/full_run_config_generator.py
```diff
@@ -0,0 +1,162 @@
...
+        list of dicts
+            keys are parameters to perf_analyzer
+            values are individual combinations of argument values
+        """
+
+        all_model_configs = self._generate_all_model_configs()
+        all_perf_configs = self._generate_all_perf_configs()
+
+        num_models = len(self._analyzer_config.model_names)
+        mc_list_interval = len(all_model_configs) // num_models
```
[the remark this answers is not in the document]
xprotobeast2: This is no longer in the code, i've simplified it to be per model. It looks better now.

model_analyzer/config/run_config_generator.py
```diff
@@ -0,0 +1,48 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+
+
+class RunConfigGenerator(ABC):
```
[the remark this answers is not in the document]
xprotobeast2: Yes I think currently thats correct, but when we eventually implement AutoTestConfigGenerator, we may want to define the interface that a RunConfigGenerator will need.

model_analyzer/record/metrics_manager.py
```diff
@@ -0,0 +1,252 @@
...
+        Splits up monitoring metrics into various 
+        categories, defined in ___init___
+        """
+
+        for metric in metrics:
+            if metric in DCGMMonitor.model_analyzer_to_dcgm_field:
+                self._dcgm_metrics.append(metric)
+            elif metric in PerfAnalyzer.perf_metrics:
+                self._perf_metrics.append(metric)
+                self._non_dcgm_metrics.append(metric)
```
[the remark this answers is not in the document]
xprotobeast2: This is mainly done to preserve the order specified in `metric_tags`  passed in.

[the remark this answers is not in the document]
xprotobeast2: But it may not be necessary anymore.

model_analyzer/record/metrics_manager.py
```diff
@@ -0,0 +1,252 @@
...
+
+        self._result_manager.add_model_data(
+            measurements=model_non_gpu_metric_values, has_gpu_ids=False)
+
+    def _start_monitors(self):
+        """
+        Start any metrics monitors
+        """
+
+        self._dcgm_monitor = DCGMMonitor(self._gpus, self._monitoring_interval,
```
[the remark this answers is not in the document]
xprotobeast2: But then we would have to move the call to `_destroy_monitors` out of the class no?

model_analyzer/config/run_config_generator.py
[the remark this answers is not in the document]
xprotobeast2: Inheriting from list didn't work, so instead I removed the  `__next__` implementation and returned `iter(self._run_configs)` in `__iter__`

model_analyzer/triton/model/model_config.py
xprotobeast2: I believe this is the backend name. (TF, ONNX, libtorch) etc

model_analyzer/triton/model/model_config_utils.py
xprotobeast2: I've been looking for this. Is this used anywhere in Triton?

Tabrizian: It is not. I was able to load pbtxt using this code.

model_analyzer/record/metrics_manager.py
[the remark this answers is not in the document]
Tabrizian: Right, let's ignore my comment.

model_analyzer/triton/model/model_config.py
Tabrizian: I see.

model_analyzer/result/result_comparator.py
```diff
@@ -0,0 +1,202 @@
...
+
+        gpu_rows1 = []
+        gpu_rows2 = []
+
+        for _, gpu_row in gpu_data1.items():
+            gpu_rows1.append(gpu_row)
+        for _, gpu_row in gpu_data2.items():
+            gpu_rows2.append(gpu_row)
+
+        avg_gpu_data1 = average_list(gpu_rows1)
```
Tabrizian, line 142: If I understood correctly, the reason that we need to separate gpu_data and non_gpu data is that gpu_data is a list of measurements while the perf_analyzer result is a single number. I think it would be great if we can aggregate all the GPU (or group of metrics) into a single number for each metric being monitored. So that in the measurement class, we have a list of metrics and a single number representing the result for that specific metric. Then, we can treat all the metrics uniformly without having to separate them.

model_analyzer/result/result_comparator.py
```diff
@@ -0,0 +1,202 @@
...
+                elif value_diff < -threshold:
+                    return -1
+            elif priority in self._gpu_type_to_idx:
+                metric_idx = self._gpu_type_to_idx[priority]
+                threshold = self._comparison_threshold_factor * avg_gpu_data1[
+                    metric_idx].value()
+                value_diff = (avg_gpu_data1[metric_idx] -
+                              avg_gpu_data2[metric_idx]).value()
+
+                if value_diff > threshold:
```
Tabrizian, line 168: What about comparing the absolute value of the difference and threshold instead?

qa/L0_config/config_generator.py
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Have you rebased? It shows these files as if they are part of the current PR.

model_analyzer/triton/model/model_config.py
```diff
@@ -0,0 +1,166 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: This file can be removed.

qa/L0_config/test.sh
```diff
@@ -0,0 +1,126 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Same

Tabrizian (changes requested):

model_analyzer/config/full_run_config_generator.py
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        model_configs = []
+        for params in self._generate_parameter_combinations(
+                model_config_params):
+            model_path = os.path.join(self._analyzer_config.model_repository,
+                                      params['model_name'])
+            model_config = ModelConfig.create_from_file(model_path=model_path)
+
+            # TODO edit model_config parameters
```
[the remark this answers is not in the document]
xprotobeast2: Hmm I can get rid of it, but I put it here as a reminder that there is some work to be done in this section when we're enabling instance counts/dynamic batching support. Not sure if thats still true.

model_analyzer/config/full_run_config_generator.py
```diff
@@ -0,0 +1,134 @@
...
+        for model_config in all_model_configs:
+            self._run_configs.append(RunConfig(model_config, all_perf_configs))
+
+    def _generate_all_model_configs(self):
+        """
+        Generates all the possible ModelConfig objects
+        """
+
+        # TODO replace [None] with list from the config
+        model_config_params = {
```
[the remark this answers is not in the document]
xprotobeast2: Only in line 83. I'll get rid of it as part of enabling dynamic batching/instance count sweeping.

model_analyzer/record/record.py
```diff
@@ -136,3 +137,42 @@ def timestamp(self):
...
+        Defines left multiplication for records
+        with floats or ints
+
+        Returns
+        -------
+        Record
+        """
+
+        if isinstance(other, int) or isinstance(other, float):
+            self._value = self._value * other
```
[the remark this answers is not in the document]
xprotobeast2: Yes, I did this because I thought it would be hard to extract the underlying type, and justified it by saying that only temporaries (records that are not actual data) are being multiplied or divided anyway.

model_analyzer/record/record.py
```diff
@@ -136,3 +137,42 @@ def timestamp(self):
...
+            return self
```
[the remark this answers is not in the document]
xprotobeast2: Agreed

model_analyzer/result/constraint_manager.py
```diff
@@ -0,0 +1,75 @@
...
+        ------
+        True if measurement passes constraints
+        False otherwise
+        """
+
+        gpu_data = measurement.gpu_data()
+        non_gpu_data = measurement.non_gpu_data()
+
+        for metrics in gpu_data.values():
+            if not self._check_constraints(metrics=metrics):
```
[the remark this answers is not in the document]
xprotobeast2: This should be not too bad if we introduce a tolerance parameter to `_check_constraints` and perhaps pull the outermost for loop out of the function, and call it with/without tolerance based on the type of the metric. But for passing/failing constraints, I don't see a reason why we would want to have a tolerance given the constraints can be adjusted if needed. For example if I want all the results that have over 90% utilization, do I really want results that have 85%? If I am okay with that why not just set my lower limit to 85%?

model_analyzer/result/result_comparator.py
xprotobeast2: wouldn't that remove the information about which is greater? `abs(value_diff) < threshold` would allow me to `continue` but within `abs(value_diff) > threshold` I would need to add further checks of the sort I've used here.

model_analyzer/triton/model/model_config.py
xprotobeast2: Will do. But I think #78 will need to be merged first right?

qa/L0_config/config_generator.py
xprotobeast2: yes I rebased it this is quite strange

model_analyzer/result/result_comparator.py
xprotobeast2: You're saying we don't need the per gpu data? I explicitly average over GPU ID in the result comparator, but do you mean we should do that earlier/elsewhere in the process?

qa/L0_config/test.sh
xprotobeast2: Fixed

model_analyzer/config/full_run_config_generator.py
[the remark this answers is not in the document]
Tabrizian: Sure, let's keep it.

model_analyzer/result/constraint_manager.py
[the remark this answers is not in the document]
Tabrizian: I see what you are saying. It is not just tolerance. I meant some metrics are very volatile like GPU utilization. For example, the user may want to host multiple models on the same GPU, and if 95% of the time the GPU utilization is below 50% that is ok. If a couple of measurements represent a different value, that will not be a deal-breaker for the user. We can fix this later too, if it looks like a reasonable adjustment to you.

model_analyzer/result/result_comparator.py
Tabrizian: Right.

model_analyzer/triton/model/model_config.py
Tabrizian: Right, sorry I thought it is not used anywhere.

model_analyzer/result/result_comparator.py
Tabrizian: I see. I thought the reason that this is split out is because of the GPU metrics being a list of measurements per GPU not because we may have multiple GPUs.

Tabrizian (dismissed):

Tabrizian (approved):

PR #77, opened 2021-02-04
Add validation to Model Analyzer config

Tabrizian:
Previously the following problematic configs were accepted in the Model Analyzer Config:

```yaml
batch_sizes: []
```

```yaml
concurrency:
  start: 1
  stop: 10
  undefined_key: 12
```

Now an error stating that this is an invalid config is shown.

```
ERROR:root:Model Analyzer encountered an error: Can't set value { [] } for field batch_sizes.
```

I've also unified the output of model names. Model names can be represented in various different formats and parsing different variations could be difficult. Now, model_names will always be a list of `ConfigModel`.

### DIFF
--- a/model_analyzer/config/config.py
+++ b/model_analyzer/config/config.py
@@ -107,23 +107,20 @@ class AnalyzerConfig:
                         required=True),
                     ConfigListString(),
                 ],
-                description=
-                'Comma-delimited list of the model names to be profiled'))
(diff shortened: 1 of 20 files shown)


xprotobeast2 (approved):

xprotobeast2 (commented):
Looks great! output_mapper seems like a nice solution to the problem.

dzier (approved):

PR #78, opened 2021-02-09
Add support for parsing the config values from the Triton protobuf

Tabrizian:
~~This needs to be merged after #77 and #69.~~

### DIFF
--- a/model_analyzer/config/config_value.py
+++ b/model_analyzer/config/config_value.py
@@ -17,6 +17,8 @@ from abc import abstractmethod
 
 from model_analyzer.constants import \
     MODEL_ANALYZER_SUCCESS, MODEL_ANALYZER_FAILURE
+from model_analyzer.model_analyzer_exceptions \
+    import TritonModelAnalyzerException
 
 
 class ConfigValue(abc.ABC):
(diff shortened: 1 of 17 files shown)


Tabrizian:
You can click on the individual commits to see the related changes.

model_analyzer/config/config_value.py
```diff
@@ -85,23 +83,33 @@ def value(self):
...
+            for key, value_ in return_result.items():
                 if hasattr(value_, 'value'):
-                    return_result[key] = value_.value()
+                    final_return_result[key] = value_.value()
                 else:
-                    return_result[key] = value_.value()
-        elif type(self._value) is list:
-            return_result = []
-            for item in self._value:
+                    final_return_result[key] = value_.value()
```
xprotobeast2, line 95: Should this be 

```
final_return_result[key] = value_
```

Seems like both branches are the same

model_analyzer/triton/model/model_config.py
```diff
@@ -0,0 +1,143 @@
...
+            )
+
+        if os.path.isfile(model_path):
+            raise TritonModelAnalyzerException(
+                'Model output path must be a directory.')
+
+        with open(os.path.join(model_path, "config.pbtxt"), 'r+') as f:
+            config_str = f.read()
+
+        print(config_str)
```
xprotobeast2, line 63: Do we need this `print`

xprotobeast2 (changes requested):

model_analyzer/config/config_value.py
Tabrizian: Good catch! The other branch should never happen. I replaced it with an exception.

model_analyzer/triton/model/model_config.py
Tabrizian: Right, I removed it.

xprotobeast2 (approved):

PR #79, opened 2021-02-12
Improve config error messages

Tabrizian:
Some example error messages:

 ```
 Value "{'a': 'b'}" cannot be set for field "my_field". This field allows multiple types of values. You only need to fix one of the errors below:
        * If a dictionary is used for field "my_field", it should only contain "start" and "stop" key with an optional "step" key. Currently, contains ['a'].
        * Value "{'a': 'b'}" for field "my_field" should be a primitive type.
 ```    
 
 ```
 The value for field "my_field" should be a list and the length must be larger than zero.
 ```
 
 
 ```
 Can't set the value for field "model_names". The error below occured when parsing the config:
Value "{'vgg_19_graphdef': {'model_config_parameters': {'instance_group': [{'kind': ['KIND_CPU', 'KIND_GPU', 'KIND_DPU'], 'count': 1}]}}}" cannot be set for field "model_names". This field allows multiple types of values. You only need to fix one of the errors below:
        * Value "[{'kind': ['KIND_CPU', 'KIND_GPU', 'KIND_DPU'], 'count': 1}]" cannot be set for field "model_names.vgg_19_graphdef.model_config_parameters.instance_group". This field allows multiple types of values. You only need to fix one of the errors below:
                * Value "['KIND_CPU', 'KIND_GPU', 'KIND_DPU']" cannot be set for field "model_names.vgg_19_graphdef.model_config_parameters.instance_group.kind". This field allows multiple types of values. You only need to fix one of the errors below:
                        * Value "['KIND_CPU', 'KIND_GPU', 'KIND_DPU']" for field "model_names.vgg_19_graphdef.model_config_parameters.instance_group.kind" is not acceptable. Value should be one of the following values: "['KIND_AUTO', 'KIND_GPU', 'KIND_CPU', 'KIND_MODEL']".
                        * Value "KIND_DPU" for field "model_names.vgg_19_graphdef.model_config_parameters.instance_group.kind" is not acceptable. Value should be one of the following values: "['KIND_AUTO', 'KIND_GPU', 'KIND_CPU', 'KIND_MODEL']".
                 
                * Value for field "model_names.vgg_19_graphdef.model_config_parameters.instance_group" must be a list, value is "{'kind': ['KIND_CPU', 'KIND_GPU', 'KIND_DPU'], 'count': 1}".
         
        * Value for field "model_names" must be a list, value is "{'vgg_19_graphdef': {'model_config_parameters': {'instance_group': [{'kind': ['KIND_CPU', 'KIND_GPU', 'KIND_DPU'], 'count': 1}]}}}".
        * The value for field "model_names" should not be a dictionary, current value is "{'vgg_19_graphdef': {'model_config_parameters': {'instance_group': [{'kind': ['KIND_CPU', 'KIND_GPU', 'KIND_DPU'], 'count': 1}]}}}".
 ```
I have included the field name in every error message too so that they can find it easier.

### DIFF
--- a/model_analyzer/config/config_union.py
+++ b/model_analyzer/config/config_union.py
@@ -12,23 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from model_analyzer.constants import \
-    MODEL_ANALYZER_FAILURE, MODEL_ANALYZER_SUCCESS
+from .config_status import ConfigStatus
 from .config_value import ConfigValue
+from model_analyzer.constants import \
+    CONFIG_PARSER_FAILURE, CONFIG_PARSER_SUCCESS
 
 
 class ConfigUnion(ConfigValue):
     """
     ConfigUnion allows the value to be any of multiple ConfigValue types.
     """
-
     def __init__(self,
                  types,
                  preprocess=None,
                  required=False,
                  validator=None,
-                 output_mapper=None
-                 ):
+                 output_mapper=None,
+                 name=None):
         """
         Create a new ConfigUnion.
 
@@ -44,11 +44,13 @@ class ConfigUnion(ConfigValue):
             A validator for the final value of the field.
         output_mapper: callable or None
             This callable unifies the output value of this field.
+        name : str
+            Fully qualified name for this field.
         """
 
         self._types = types
         self._used_type_index = 0
-        super().__init__(preprocess, required, validator, output_mapper)
+        super().__init__(preprocess, required, validator, output_mapper, name)
 
     def set_value(self, value):
         """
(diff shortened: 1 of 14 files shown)


model_analyzer/config/config_union.py
```diff
@@ -58,13 +60,44 @@ def set_value(self, value):
...
+            config_status = type_.set_value(value)
+            config_statuses.append(config_status)
+            if config_status.status() == CONFIG_PARSER_SUCCESS:
                 self._used_type_index = i
                 return super().set_value(type_)
         else:
-            return MODEL_ANALYZER_FAILURE
+            message = f'Value "{value}" cannot be set for field "{self.name()}".'
+            message += ' This field allows multiple types of values.'
+            message += ' You only need to fix one of the errors below:\n'
```
xprotobeast2, line 73: Here you can also do
```
message = (
    f'Value "{value}" cannot be set for field "{self.name()}".'
    ' This field allows multiple types of values.'
    ' You only need to fix one of the errors below:\n'
)
```

I'm not 100% sure but I think this will be less copies. If the above doesn't work maybe the single quotes and double quotes need to be exchanged.

xprotobeast2 (approved):

xprotobeast2 (approved):

PR #81, opened 2021-02-15
Add support for parameter sweeping

Tabrizian:
Below shows an example output table:

```
Models (GPU Metrics):
Model                      GPU ID   Batch   Concurrency   Model Config Path             Max GPU Memory Usage(MB)   Max GPU Memory Available(MB)   Max GPU Utilization(%)  
classification_breast_v1   0        1       1             classification_breast_v1_i0   1916.7951198990324         22302.201514514094             8.315944467816575       
classification_breast_v1   0        1       1             classification_breast_v1_i1   1920.0479596129576         22298.95204038704              7.98359276398822        
classification_breast_v1   0        1       1             classification_breast_v1_i3   1922.0159932659933         22296.98400673401              8.22053872053872        
classification_breast_v1   0        1       1             classification_breast_v1_i2   1918.9962073324905         22300.00379266751              7.739570164348925       

Models (Inference):
Model                      Batch   Concurrency   Model Config Path             Throughput(infer/sec)   Average Latency(us)   Max RAM Usage(MB)   Max RAM Available(MB)  
classification_breast_v1   1       1             classification_breast_v1_i0   348.0                   2495.0                2768.0              25928.0                
classification_breast_v1   1       1             classification_breast_v1_i1   322.4                   2727.0                8.0                 25824.0                
classification_breast_v1   1       1             classification_breast_v1_i3   320.2                   2745.0                2820.0              25840.0                
classification_breast_v1   1       1             classification_breast_v1_i2   320.6                   2737.0                2827.0              25841.0                

Model   GPU ID   Batch   Concurrency   Model Config Path   Max GPU Memory Usage(MB)   Max GPU Memory Available(MB)   Max GPU Utilization(%)  

Model   Batch   Concurrency   Model Config Path   Throughput(infer/sec)   Average Latency(us)   Max RAM Usage(MB)   Max RAM Available(MB)  
```

There is a new directory created that contains all the generated/used model configurations. "Model Config Path" indicates the location of the output model config that was used for generating a given run.

You can refer to the `test_run_config_generator` to see some example sweeps.


model_analyzer/result/result_manager.py
```diff
@@ -227,12 +234,15 @@ def _compile_measurement(self, measurement, gpu_table_key,
         """
 
         perf_config = measurement.perf_config()
-        model_name = perf_config['model-name']
+        model_name = perf_config.get_config_model().model_name()
```
aramesh7, line 237: Instead of having the perf_config store a reference to the config model, we can get the temporary name from the RunConfig stored in the RunResult in `compile_results` [here](https://github.com/triton-inference-server/model_analyzer/blob/2c393d75f14b1ea24932b3928a8aa0ef66de438a/model_analyzer/result/result_manager.py#L205) by calling `next_best_result.get_run_config().model_name()`. We can add the `model_name` parameter to both `_compile_measurements` and `_compile_measurement`.

model_analyzer/perf_analyzer/perf_config.py
```diff
@@ -103,6 +107,17 @@ def to_cli_string(self):
 
         return ' '.join(args)
 
+    def get_config_model(self):
```
aramesh7, line 110: This function hopefully will not be necessary

aramesh7 (changes requested):

model_analyzer/analyzer.py
```diff
@@ -65,45 +66,64 @@ def run(self):
...
+                    config_model=model)
+            run_config_generator = RunConfigGenerator(model,
+                                                 analyzer_config=self._config)
+            for run_config in run_config_generator.get_run_configs():
+                model_config = run_config.model_config()
+                original_model_name = run_config.model_name()
+                model_name = model_config.get_field('name')
+
+                # Create the directory for the new model
+                os.mkdir(f'{output_model_repo_path}/{model_name}')
```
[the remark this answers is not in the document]
Tabrizian: It is a good idea but I guess it is better to keep it the way it currently is for two reasons:

1. If one of the configs that the user is trying to sweep through has an issue, it can be quickly debugged by `tritonserver --model-repository <output model repository>` and find the models that cannot be loaded.
2. If we plan to integrate the multi-model analysis part, having all the models in a single folder helps quick loading of multiple models to study the co-located performance.

model_analyzer/entrypoint.py
```diff
@@ -76,9 +77,16 @@ def get_server_handle(config):
         logger.info('Using remote Triton Server...')
         server = TritonServerFactory.create_server_local(path=None,
                                                          config=triton_config)
+        logger.info(
+            'GPU memory metrics reported in the remote mode are not'
```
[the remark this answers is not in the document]
Tabrizian: `start`/`stop` does not lead to any action in the remote mode. I think the values should be the same as before.

model_analyzer/entrypoint.py
```diff
@@ -272,6 +276,15 @@ def main():
         f'Triton Model Analyzer started: config={config.get_all_config()}')
 
     server = None
+    # Generate the output model repository path folder.
+    output_model_repo_path = config.output_model_repository_path
+    try:
+        os.mkdir(output_model_repo_path)
+    except OSError:
+        raise TritonModelAnalyzerException(
```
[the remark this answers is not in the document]
Tabrizian: Add a new flag to allow overriding of the output dir content.

model_analyzer/result/result_manager.py
Tabrizian: Thanks for the detailed comment. I fixed this.

model_analyzer/perf_analyzer/perf_config.py
Tabrizian: Removed.

model_analyzer/result/result_manager.py
```diff
@@ -227,14 +229,13 @@ def _compile_measurements(self, measurements):
                     inference_table_key=self.model_inference_table_failing_key)
```
aramesh7, line 229: You will need to add the model_name parameter [here](https://github.com/triton-inference-server/model_analyzer/blob/08deaa42f2067af9303fa99de6aa7f1f0fe236a4/model_analyzer/result/result_manager.py#L226) as well. None of the tests caught it unfortunately. The result, constraint, and metrics managers are difficult to unit test.

aramesh7 (changes requested):

model_analyzer/result/result_manager.py
Tabrizian: I see. Good catch. I guess we should add some tests to `L0_config` that some experiments do not pass the constraints so that the other branch of the code is tested as well.

aramesh7: I will add it in L0_constraints

aramesh7 (approved):

dzier (approved):

PR #82, opened 2021-02-19
Added weighted comparison to ResultComparator

aramesh7:
We now pass a dictionary of objectives and corresponding weights in the Model Analyzer Config
```
model_repository: /qa_model_repository/
export: true
triton_launch_mode: docker
concurrency: 2,4
perf_measurement_window: 5000
triton_http_endpoint: localhost:8000
triton_grpc_endpoint: localhost:8001
triton_metrics_url: http://localhost:8002/metrics
batch_sizes: 16,32
no_perf_output: true
model_names: 
  classification_chestxray_v1:
    parameters:
      batch_sizes: 2,4
  classification_malaria_v1:
    parameters:
      concurrency: 4,6
objectives:
  perf_throughput: 10
  perf_latency: 5 
constraints:
  perf_latency:
    max: 100000
```

Observe the order of the results (top to bottom is best to worst, and observe that relative weighting is working). For example, between the top two results for classification_chestxray_v1, the throughput is higher for the second row but the latency is much worse, so it gets shifted down. 

```
Server Only:

Models (GPU Metrics):
classification_chestxray_v1   0        4       2             5765.0                     18454.0                        63.0                    
classification_chestxray_v1   0        4       4             5765.0                     18454.0                        61.0                    
classification_chestxray_v1   0        2       2             929.0                      23290.0                        63.0                    
classification_chestxray_v1   0        2       4             5765.0                     18454.0                        45.0                    
classification_malaria_v1     0        32      4             5765.0                     18454.0                        68.0                    
classification_malaria_v1     0        32      6             5765.0                     18454.0                        67.0                    
classification_malaria_v1     0        16      4             5765.0                     18454.0                        40.0                    
classification_malaria_v1     0        16      6             5765.0                     18454.0                        41.0                    

Models (Inference):
classification_chestxray_v1   4       2             226.4                   31160.0               5613.0              25659.0                
classification_chestxray_v1   4       4             250.4                   59517.0               5634.0              25621.0                
classification_chestxray_v1   2       2             129.6                   29705.0               5569.0              26946.0                
classification_chestxray_v1   2       4             134.4                   57911.0               5590.0              25670.0                
classification_malaria_v1     32      4             4883.2                  24262.0               5641.0              25581.0                
classification_malaria_v1     32      6             4729.6                  38376.0               5647.0              25505.0                
classification_malaria_v1     16      4             2582.4                  23613.0               5637.0              25622.0                
classification_malaria_v1     16      6             2505.6                  37104.0               5637.0              25621.0                

Model   GPU ID   Batch   Concurrency   Max GPU Memory Usage(MB)   Max GPU Memory Available(MB)   Max GPU Utilization(%)  

Model   Batch   Concurrency   Throughput(infer/sec)   Average Latency(us)   Max RAM Usage(MB)   Max RAM Available(MB)
```


model_analyzer/config/config.py
```diff
@@ -208,8 +210,8 @@ def _fill_config(self):
         self._add_config(
             ConfigField(
                 'objectives',
-                field_type=ConfigListString(),
-                default_value=['perf_throughput'],
+                field_type=objectives_scheme,
+                default_value={'perf_throughput': 10},
```
dzier, line 214: There should be some default value for all of the metrics we use in the comparison. I am assuming that this is just for an initial starting point. You will be adding the other metrics later.

model_analyzer/config/config_object.py
```diff
@@ -84,29 +85,44 @@ def set_value(self, value):
...
-                new_value[key] = new_item
-
-                # Set the value of the new field
-                config_status = new_item.set_value(value_)
-
-                # If it was not able to set the value, for this
-                # field, we fail.
-                if config_status.status() == CONFIG_PARSER_FAILURE:
-                    return config_status
+            if '**' in schema:
```
dzier, line 88: What are you actually doing here? I think you need a higher level comment. The description does not seem to match needing to "deep copy schemas"

model_analyzer/result/result_comparator.py
```diff
@@ -63,17 +56,16 @@ def __init__(self,
             for i in range(len(gpu_metric_types))
         }
 
-        self._metric_priorities = metric_priorities
-        self._comparison_threshold_factor = (comparison_threshold_percent *
-                                             1.0) / 100
-
-        # TODO implement a weighted comparison
-        self._weights = weights
+        # Normalize metric objectives
```
dzier, line 59: I would use `_metric_weights` as that is a more accurate description and purpose. `objectives` do not make sense here.

model_analyzer/result/result_comparator.py
```diff
@@ -146,33 +138,30 @@ def compare_measurements(self, measurement1, measurement2):
...
-
-                if value_diff > threshold:
-                    return 1
-                elif value_diff < -threshold:
-                    return -1
-            elif priority in self._gpu_type_to_idx:
-                metric_idx = self._gpu_type_to_idx[priority]
-                threshold = self._comparison_threshold_factor * avg_gpu_data1[
-                    metric_idx].value()
+                value_sum = (non_gpu_data1[metric_idx] +
```
dzier, line 147: You don't add the two values for the denominator. The equation to compare run A to run B will be...

```
Score = sum^0_n{w_i * (B_i - A_i) / (A_i)}
```
Where `w_i` i s the weight of the metric, `B_i` is the metric value for run B, and `A_i` is the is the metric value for run A.

The equation `(B_i - A_i) / (A_i)` is commonly used it research to determine the gain one test over another. For example if A_i = 100, and B_i = 150, then the gain is `(150-100)/100 = .5`. So in this example, B is 50% better than A. It works the other way too. A_i = 100, and B_i = 50, then the gain is `(50-100)/100 = -.5`, so B is 50% worse than A.

dzier: Do keep in mind, that some metrics are opposite. For example, GPU memory usage, lower is better. In this case the equation becomes.
```
Score = sum^0_n{w_i * (-1)*(B_i - A_i) / (A_i)}
```

model_analyzer/result/result_comparator.py
```diff
@@ -146,33 +138,30 @@ def compare_measurements(self, measurement1, measurement2):
...
-                    return 1
-                elif value_diff < -threshold:
-                    return -1
+                value_sum = (avg_gpu_data1[metric_idx] +
+                             avg_gpu_data2[metric_idx]).value()
             else:
                 raise TritonModelAnalyzerException(
-                    f"Category unknown for objective : '{priority}'")
+                    f"Category unknown for objective : '{objective}'")
+            score_diff += (weight * (value_diff / value_sum))
```
dzier, line 158: This is correct, except we do not sum the value (denominator).

tests/test_config.py
```diff
@@ -183,13 +183,13 @@ def test_range_and_list_values(self):
                             'batch_sizes': [1],
                             'concurrency': [1]
                         },
-                        objectives=['perf_throughput']),
+                        objectives={'perf_throughput': 10}),
```
dzier, line 186: What good does this do with a single metric? Are you just testing to see that the one with the highest through put is selected always?

dzier (changes requested):

tests/test_config.py
xprotobeast2: This particular unit test does not test anything to do with objectives, so it is just ensuring the default value is present.

model_analyzer/config/config.py
xprotobeast2: We can change the default right now. What this means is that by default, the model analyzer will just order the results by throughput, highest to lowest. What do you suggest regarding the default? I can imagine adding a weightage for latency, but its relative value will be arbitrary.

tests/test_config.py
xprotobeast2: There are unit tests test_result_comparator.py that test with multiple metrics.

model_analyzer/config/config.py
```diff
@@ -144,6 +144,8 @@ def _fill_config(self):
                         field_type=ConfigPrimitive(str, required=True),
                         description='Model repository location'))
 
+        objectives_scheme = ConfigObject(
+            schema={'**': ConfigPrimitive(type_=int)})
```
Tabrizian, line 148: Can you be more explicit about the list of parameters allowed here? I think we know all the keys that should be allowed under this key.

model_analyzer/config/config_object.py
Tabrizian: Why did you need to move `*` to the top of the for loop?

tests/test_result_comparator.py
```diff
@@ -0,0 +1,316 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: 2021

model_analyzer/config/config.py
Tabrizian: Also, I think we should be more flexible here and allow the user to both provide a list of objectives + a dictionary. If a list is provided the weights will be equal. You can unify the output results using the `output_mapper` and creating a new class that contains the unified output. This way we won't have to specify a default weight for the output too.

Tabrizian (changes requested):
I think it would be great if you can add unit tests for each underlying CPUAvailableRam, etc to make sure that the addition and multiplication work as expected.

model_analyzer/config/config.py
xprotobeast2: If we are explicit, we will have to update the list of keys when a new metric is added, removed, or existing metric name is changed. I will increase the flexibility using `output_mapper`, but do you think a separate class is necessary here?

xprotobeast2: Actually we can be explicit by getting all registered record types from record.py

Tabrizian: Right, sounds good!

Tabrizian (approved):
Looks good! It would be great if you can add one more test for the case when `key` is not in the list of allowed objectives too but it is only a suggestion.

Tabrizian (approved):

model_analyzer/config/config.py
dzier: Why are you ordering anything? Nothing needs to be ordered. The integer weight values are normalized and can be added in any order.  As to what the defaults should be, that is something we need to experiment with. Remember, the weights are there to determine emphasis priority. If we think that latency is more important that GPU memory usage, then it should have a larger value. If we think it is 2x more important, then it should have a value 2x as large.

dzier: Think of it this way. If a user does not specify custom weights, what values would MA use to determine the best run. Every metric that we use in the comparison heuristic will need a default weight value.

aramesh7: Let me clarify, the above default value means that by default, the model analyzer will only compare the throughput values for each run, and sort the runs using that. This is because not only does the model analyzer give us the best result as defined by this weighting dictionary, it will give us the results in descending order of 'goodness'.  This way we have the best, the second best etc. 

If we haven't specified the weights for a metric, it is assumed to be 0. (That metric is not considered for comparison). We can experiment with and determine what an effective default should be.

dzier: That might not be the best. The best throughput could be disastereous in every other metric.  But setting that aside, where are you setting the defaults to 0 for all other metrics?

aramesh7: [Here](https://github.com/triton-inference-server/model_analyzer/blob/fe4bbe28d502413fc393db812bf87ced65f1d557/model_analyzer/result/result_comparator.py#L142) in the result comparator, there's a loop over the objectives specified in the dictionary. If the tag isn't present in this dictionary, the value isn't considered for comparison. This is equivalent to specifying it but setting its weight to 0.

dzier (approved):

PR #84, opened 2021-03-02
Added support for generating plots

aramesh7:
Config like this 

```
plots:
  first_plot:
    title: Throughput vs. Latency
    x_axis: perf_latency
    y_axis: perf_throughput
  second_plot:
    title: GPU Memory vs. Latency
    x_axis: perf_latency
    y_axis: gpu_used_memory
top_n_configs: 3
top_n_measurements: 5
```

Generates plots like the following 

image

image


model_analyzer/analyzer.py
```diff
@@ -122,145 +126,28 @@ def run(self):
...
+            # Write plots to disk
+            self._plot_manager.complete_plots(model_name=model.model_name())
 
-        self._write_result(self._result_manager.get_server_table(),
-                           writer,
-                           column_separator,
-                           ignore_widths=True,
-                           include_title=False)
+            # Write results to tables
+            self._result_manager.compile_results()
```
Tabrizian, line 144: Can we rename `compile_results` to `write_results`?

model_analyzer/config/input_config/config.py
```diff
@@ -233,7 +245,38 @@ def objective_list_output_mapper(objectives):
...
+                default_value={
+                    'default_plot': {
+                        'title': 'Throughput vs. Latency',
+                        'x_axis': 'perf_latency',
+                        'y_axis': 'perf_throughput'
+                    }
+                },
+                description=
+                'Model analyzer uses the information in this section to construct plots of the results.'
+            ))
```
Tabrizian, line 265: Nice!

model_analyzer/config/run_config/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: What about moving files to `config/run` and `config/input`? instead of run_config and input_config.

model_analyzer/result/measurement.py
```diff
@@ -36,11 +40,32 @@ def __init__(self, gpu_data, non_gpu_data, perf_config, comparator):
             Handle for ResultComparator that knows how to order measurements
         """
 
+        # average values over all GPUs
         self._gpu_data = gpu_data
+        self._avg_gpu_data = average_list(list(self._gpu_data.values()))
```
Tabrizian, line 45: Can `average_list` be renamed to `aggregator` and add a new parameter to the constructor which by default is equal to `average_list`?

Tabrizian (commented):
Some minor comments, rest LGTM.

model_analyzer/analyzer.py
aramesh7: This function doesn't actual 'write results' in the way that `ResultManager.write_and_export_results` does. It just puts them into ResultTables, which was why i used the word compile. I didn't want to confuse it with writing output.

model_analyzer/result/measurement.py
aramesh7: yes, i'll make this a priority for any future aggregation.

dzier (commented):
It seems like this commit is doing many things. Remember, we should separate commits for different purposes when possible. There are four main purposes to commit a change

1. Add new behavior - This is generally brand new code and new tests, but only minimally changes existing code or tests
2. Fix incorrect behavior (i.e. bug fixing) - This involves correcting/adding new tests and changing existing code to make tests pass
3. Refactoring - This involves changing existing production code (new classes, general structure, etc.), but neither tests nor behavior changes
4. Performance Improvements - This is like refactoring, except no structures change. There may be algorithm changes, etc,

By keeping these to separate commits, we can easily see changes, and revert/cherry pick specific improvements. It also makes PRs easier. This PR seems like a combination of 1 and 3 above. It would be nice to see two commits in this PR, so I can review the types of changes explicitly.

dzier (approved):

Tabrizian (approved):

dzier (approved):

PR #88, opened 2021-03-06
Added support for result summary

aramesh7:


model_analyzer/reports/report_manager.py
```diff
@@ -0,0 +1,223 @@
...
+
+        for _, measurement in self._measurements[model_name]:
+            for gpu in cuda.gpus:
+                if gpu.id in measurement.gpus_used(
+                ) and gpu.id not in seen_gpus:
+                    seen_gpus.add(gpu.id)
+                    gpu_names.append((gpu.name).decode('ascii'))
+                    with gpu:
+                        mems = cuda.current_context().get_memory_info()
+                        max_memories.append(mems.total // (2**30))
```
Tabrizian, line 199: What does 2^30 represent? It looks like a very large number

model_analyzer/result/result_manager.py
```diff
@@ -46,7 +47,7 @@ class ResultManager:
     model_gpu_table_failing_key = 'model_gpu_metrics_failing'
     model_inference_table_failing_key = 'model_inference_metrics_failing'
 
-    def __init__(self, config):
+    def __init__(self, config, statistics):
```
Tabrizian, line 50: docstring for statistics is missing.

model_analyzer/result/run_result.py
```diff
@@ -57,33 +80,111 @@ def add_measurement(self, measurement):
...
+    def instance_group_string(self):
+        """
+        Returns
+        -------
+        str
+            representation of the instance group used 
+            to generate this result
+        """
+        instance_group_list = self.run_config().model_config().get_config(
+        )['instance_group']
```
Tabrizian, line 165: if the config does not have instance_group will this work?

model_analyzer/result/run_result.py
```diff
@@ -57,33 +80,111 @@ def add_measurement(self, measurement):
...
+        Returns
+        -------
+        str
+            representation of the dynamic batcher
+            configuration used to generate this result
         """
 
-        return heapq.nsmallest(n, self._measurements)
+        model_config = self.run_config().model_config()
+        if 'dynamic_batching' in model_config.get_config():
```
Tabrizian, line 182: If the config has dynamic batching but it does not have preferred batch size I think there will be issues.

Tabrizian (changes requested):

model_analyzer/reports/report_manager.py
aramesh7: I'll add a comment. Its to convert bytes to GB

model_analyzer/triton/model/model_config.py
```diff
@@ -203,3 +203,43 @@ def get_field(self, name):
...
+        model_config = self.run_config().model_config().get_config()
+        if 'instance_group' in model_config:
+            instance_group_list = ['instance_group']
+            group_str_list = [
+                f"{group['count']}-{group['kind'].split('_')[1]}"
+                for group in instance_group_list
+            ]
+            return ','.join(group_str_list)
+        else:
+            return '1-GPU'
```
Tabrizian, line 245: What if it is CPU model?

Tabrizian (changes requested):

Tabrizian (dismissed):

Tabrizian (dismissed):

model_analyzer/record/record.py
```diff
@@ -106,13 +107,22 @@ def __init__(self, value, timestamp):
         self._value = value
         self._timestamp = timestamp
 
+    @staticmethod
+    def aggregation_function():
```
Tabrizian, line 111: It would be great if this can be one of the constructor's arguments with the default that you have here.

model_analyzer/record/types/gpu_utilization.py
```diff
@@ -40,11 +40,23 @@ def __init__(self, value, device=None, timestamp=0):
         super().__init__(value, device, timestamp)
 
     @staticmethod
-    def header(aggregation_tag=None):
+    def aggregation_function():
```
Tabrizian, line 43: Same, and why is the function static?  Same about the header function.

model_analyzer/reports/report_manager.py
```diff
@@ -74,103 +71,120 @@ def export_summary(self,
...
         for model_config, measurement in self._measurements[model_name]:
             dynamic_batching_str = model_config.dynamic_batching_string()
             instance_group_str = model_config.instance_group_string()
+            if best:
+                summary_sentence = (
+                    f"In {num_measurements} measurements "
+                    f"{instance_group_str} model instances at concurrency "
+                    f"{measurement.perf_config()['concurrency-range']} "
+                    f"and max dynamic batch size of {dynamic_batching_str} "
+                    f"on platform {model_config.get_field('platform')} delivers "
```
Tabrizian, line 171: platform is going to be deprecated in Triton soon. I think only TF-TRT is using the platform field. Other backends are using the `backend` field. You can check here whether the `backend` field exist if it doesn't exist you can use `platform`.

model_analyzer/reports/report_manager.py
```diff
@@ -74,103 +71,120 @@ def export_summary(self,
...
+        best = True
+        summary_sentence = ""
         for model_config, measurement in self._measurements[model_name]:
             dynamic_batching_str = model_config.dynamic_batching_string()
             instance_group_str = model_config.instance_group_string()
+            if best:
+                summary_sentence = (
+                    f"In {num_measurements} measurements "
+                    f"{instance_group_str} model instances at concurrency "
+                    f"{measurement.perf_config()['concurrency-range']} "
```
Tabrizian, line 169: I think Joey didn't mention concurrency and probably we should not present concurrency here.

Tabrizian (changes requested):

model_analyzer/record/types/gpu_utilization.py
aramesh7: Because I think the header and aggregation function belong to the class rather than the individual instance of the metric, but perhaps the aggregation function would need to be changed within a single run of model_analyzer ? It will be complicated to passs in the aggregation function in the constructor because a lot of the time we are not explicit about which metric we are constructing.

model_analyzer/record/record.py
aramesh7: Addressed below

model_analyzer/reports/report_manager.py
aramesh7: > Can we summarize in one sentence: maximum throughout at 40ms is model concurrency of 2 with max dynamic batching of 32 from ( number of variations) explored in (number of minutes)  running in (framework name) on (gpu model) with (precision) ?

Joey did ask for model concurrency is that different from our request concurrency?

Tabrizian: I see. Sorry I missed that part. No that is the same.

model_analyzer/record/types/gpu_utilization.py
Tabrizian: Ok, I see. Probably because we don't change the aggregation function it should be fine.

Tabrizian (approved):

dzier (approved):

PR #93, opened 2021-03-10
Add docs for objectives + constraints + config search

Tabrizian:


docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+# Disables automatic config search
+[ run_config_search_disable: <bool> | default: false ]
 ```
 
-If you save configuration file in `config.yml`, you can provide
-the path to the Model Analyzer config using the 
-`-f` flag.
+## `<constraint>`
+A constraint, specifies the bounds that determine a successful run. There are
+three constraints allowed:
```
dzier, line 149: Why are we limiting to just three constraints? Shouldn't we be able to constrain any measurable output?

Tabrizian: @aramesh7 Can we support other constraints too? I can add them to the config.

docs/config.md
```diff
@@ -16,12 +16,13 @@ limitations under the License.
 
 # Model Analyzer Config
 
-In addition to the CLI flags, you can config the Model Analyzer
-using a [YAML](https://yaml.org/) file too. The scheme
-for this configuration file is described below. Brackets
-indicate that a parameter is optional. For non-list
-and non-object parameters the value is set to the
-specified default.
+In addition to the CLI flags, you can config the Model Analyzer using a
```
dzier, line 19: Try to be less superfluous. Just something simple is better. 

```
The Model Analyzer can be configured with a [YAML](https://yaml.org/) file. ...
```

docs/config.md
```diff
@@ -16,12 +16,13 @@ limitations under the License.
...
+[YAML](https://yaml.org/) file too. The scheme for this configuration file is
+described below. Brackets indicate that a parameter is optional. For non-list
+and non-object parameters the value is set to the specified default.
+
+All the flags supported in the CLI, are supported in the configuration file
```
dzier, line 24:
```
Every flag supported by the CLI is supported in the configuration file.
```

docs/config.md
dzier: Can we make this a table? Something like this?

| Option Name          | Units       |  Constraint  |   Description                                                      |
| :--------------------- |  :-------:  |  :----------:  |  :---------------------------------------------------------- |
| `perf_throughput`     | inf / sec  |      min       | Specify minimum desired throughput.                           |
| `perf_latency`            | ms          |       max     | Specify maximum tolerable latency or latency budget. |
| `gpu_used_memory` | MB         |       max     | Specify maximum GPU memory used by model.           |

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+The unit for this field is megabytes.
+```yaml
+gpu_used_memory:
+    max: 200
+```
+
+The config above will filter out the results that their `gpu_used_memory` is larger than
+200 MBs.
+
+You can also use these keys together:
```
dzier, line 180: Try to avoid using "you"

```
Keys can be combined for more complex constraints:
```

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+100 milliseconds.
+
+3. `gpu_used_memory`: Using this key you can specify the maximum GPU memory used by the model.
+The unit for this field is megabytes.
+```yaml
+gpu_used_memory:
+    max: 200
+```
+
+The config above will filter out the results that their `gpu_used_memory` is larger than
```
dzier, line 177: `` ... results when `gpu_use_memory` is larger ...``

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+        latency:
+            max: 50
+```
+
+## `<objective>`
+
+Objectives allow you to specify the sorting criteria for the final results.
+The fields below are supported under this object type:
+
+1. `perf_throughput`
```
dzier, line 226: Convert to table like above

dzier (changes requested):

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+The above config will multiply obtained throughput and latency for each
+measurement by 0.5 (i.e. 0.5 * latency + 0.5 * throughput) and use this as
+the "score" to sort the measurements according to. This mode can be useful
+if you have a more advanced criteria upon which you want to sort the results.
+
+An extension of the above `objectives` is explicitly specifying the weights.
+For example:
+```yaml
+objectives:
+    perf_latency: 10
```
dzier, line 267: Why 10 and 15, and not 2 and 3? Both achieve the same thing.

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+    perf_throughput: 15
+```
+
+The score for each measurement will be a weighted average using the weights specified here.
+
+Similar to `<constraint>`, `<objective>` can be specified both globally and on a
```
dzier, line 273: Get ride of everything before the ,

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+per model basis.
+
+## `<parameter>`
```
dzier, line 276: ``## Test configuration `<parameter>` `` perhaps

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+
+This type can contain `concurrency` or `batch_sizes` as the possible keys. You
```
dzier, line 278: "A user can specify a range of test configurations that Model Analyzer will profile over. The possible configuration parameters are `concurrency` and `batch_size`. One or more parameters are specified per model only. Parameters cannot be globally specified"

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+
+```yaml
+parameters:
+    concurrency:
+        start: 2
+        stop: 64
+        step: 8
+    batch_sizes: 1,2,3
+```
+
```
dzier, line 292: You should write a paragraph here explaining what the parameters above will do. For example,

"These parameters will result in testing the concurrency configurations of 2, 10, 18, 26, 34, 42, 50, 58, and 64, for each of different batch sizes of 1, 2 and 3. This will result in 27 individual test runs of the model."

docs/config.md
dzier: Could also start a table here. Tables are useful and make it easy to document new configurations

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+        preferred_batch_size: [[1], [2], [3]]
+        max_queue_delay_microseconds: [200, 300]
+    instance_group:
+    -
+        kind: KIND_GPU
+        count: [1, 2]
+```
+
+Note that for values that accept a list by default you need to specify one
+additional list if you want to sweep through it. Otherwise, it will be only
```
dzier, line 329: "Otherwise, it will only change ..."

docs/config.md
dzier: Are there default values here? What is chosen if the user specifies nothing?

docs/config.md
```diff
@@ -122,18 +124,304 @@ model_names: <comma-delimited-string|list>
...
+
+A complete YAML config looks like below:
+```yaml
+model_repository: /path/to/model/repository/
+
+run_config_search_disable: True
+model_names:
+  model_1:
+    model_config_parameters:
+        max_batch_size: 2
```
dzier, line 344: Can you create table for all possible model_config_parameters that can be swept through?  Also, do these have a multiplying affect on the number of tests with the `<parameters>` above? If so, you may want to mention this. Otherwise some people may question why Model Analyzer is running 1000+ configurations and taking 2 days to complete.

docs/config.md
aramesh7: We do support constraints on any metric we measure

dzier (changes requested):

docs/config.md
Tabrizian: If the user does not specify anything the auto config search mode will be enabled which is explained in `config_search.md`.

docs/config.md
Tabrizian: Yes, it has a multiplying effect.

Tabrizian: I mentioned that all the Triton config parameters can be changed in `<model-config-parameters>`. I'll add a list of common parameters there.

docs/config.md
```diff
@@ -220,16 +224,18 @@ model_names:
...
-1. `perf_throughput`
-2. `perf_latency`
-3. `gpu_used_memory`
-4. `gpu_free_memory`
-5. `gpu_utilization`
-6. `cpu_used_ram`
-7. `cpu_available_ram`
+| Option Name       |   Units   | Description                                            |
+| :---------------- | :-------: | :----------------------------------------------------- |
+| `perf_throughput` | inf / sec | Use throughput as the objective.                       |
```
dzier, line 232: Why are units here? The value they put here is not the same as weights below and can cause confusion. Use units only when the user needs to enter a value in those units.

docs/config.md
```diff
@@ -258,26 +264,34 @@ objectives:
...
 For example:
 ```yaml
 objectives:
-    perf_latency: 10
-    perf_throughput: 15
+    perf_latency: 2
+    perf_throughput: 3
 ```
 
 The score for each measurement will be a weighted average using the weights specified here.
```
dzier, line 277: You should mention that the options above will 40% (2/5) to perf_latency and 60% (3/5) to perf_throughput scores.

docs/config.md
```diff
@@ -258,26 +264,34 @@ objectives:
...
+A user can specify a range of test configurations that Model Analyzer will
+profile over. The possible configuration parameters are `concurrency` and
+`batch_sizes`. One or more parameters are specified per model only. Parameters
+cannot be specified globally.
+
+Options available under this parameter are described in table below:
+
+| Option Name   | Description                                                                                                          |
+| :------------ | :------------------------------------------------------------------------------------------------------------------- |
+| `concurrency` | Request concurrency used for generating the input load. Can be a `<range>`, `<comma-delimited-list>`, or a `<list>`. |
```
dzier, line 292: Create a third column `supporting types` to list the `` `<range>`, `<comma-delimited-list>`, or a `<list>` ``

docs/config.md
```diff
@@ -304,14 +318,26 @@ model_names:
...
 using Model Analyzer. All the values supported in the [Triton
 Config](https://github.com/triton-inference-server/server/blob/master/docs/model_configuration.md)
 can be specified or swept through here. `<model-config-parameters>`  should be
 specified on a per model basis and cannot be specified globally (like `<parameter>`).
 
+Table below presents the list of common parameters that can be used for manual sweeping:
+
+|       Option       |                                                 Description                                                  |
+| :----------------: | :----------------------------------------------------------------------------------------------------------: |
+| `dynamic_batching` |  https://github.com/triton-inference-server/server/blob/master/docs/model_configuration.md#dynamic-batcher   |
```
dzier, line 337: Can you add the link to the Option Name and put a simple top level description in the description?

dzier (changes requested):

dzier (approved):

PR #98, opened 2021-03-15
Fixed null value bug in instance group string and L0_config

aramesh7:


model_analyzer/triton/model/model_config.py
```diff
@@ -235,14 +234,25 @@ def instance_group_string(self):
         """
 
         model_config = self.get_config()
+
+        # TODO change when remote mode is fixed
```
Tabrizian, line 238: What do you mean by "change when remote mode is fixed"?

aramesh7: We can't check whether CUDA is available on the machine if Triton is running remotely. but currently we require Triton to be running on the same system even if on remote mode.

Tabrizian: We can check that using the metrics API. We already do it below:

https://github.com/triton-inference-server/model_analyzer/blob/8593a473bcc923f90a892cffe59fa9980b55c27f/model_analyzer/entrypoint.py#L156

I have a suggestion for this. Since It currently won't work when we have multiple instances and it looks like we are trying to guess what is the "correct" instances, what do you think about printing whatever config is available in the format below:

* 1/KIND_GPU
* 1/KIND_CPU

If there are multiple instances, a single bullet when there is a single instance, or nothing when there isn't any instance available.

Tabrizian: I guess we can leave that field empty if there is no config provided so that we don't have to guess what is the current config.

aramesh7: True but would this be fine for now given that they can't change the model config in remote mode anyway at the moment?

aramesh7: I suppose the proper thing to do would be to check with `get_triton_metrics_gpus` during the model config generation?

Tabrizian: > True but would this be fine for now given that they can't change the model config in remote mode anyway at the moment?

The problem is that this config can exist anywhere. The model may have been loaded in the remote Triton server that doesn't have value for `instance_group`. It is not a problem with remote/local.

> I suppose the proper thing to do would be to check with get_triton_metrics_gpus during the model config generation?

I didn't follow this. Can you explain more?

I don't want to insist on this change. If you think it is ok, then feel free to go ahead.

Tabrizian (approved):

PR #100, opened 2021-03-18
Update docs for 21.03 release

aramesh7:
- Update quick start
- Update install 
- Add caveat for Use with triton sdk
- Report Generation with image


docs/config.md
```diff
@@ -42,20 +36,24 @@ batch_sizes:
     step: 2
 ```
 
-This YAML represents the array `[2, 4, 6]`.
-* Other types are described in the end of this file.
+The above YAML declares the value of batch_sizes to be an array `[2, 4, 6]`.
+
+## CLI and YAML Config Options
+
+A list of all the configuration options supported by both the CLI and YAML config file are shown below.
```
Tabrizian, line 43: Can you add this before the scheme?

"The scheme for the configuration file is described below. Brackets indicate
that a parameter is optional. For non-list and non-object parameters the
value is set to the specified default."

Tabrizian: Can you explain how they can convert the field names to CLI flags?

"You need to convert the snake case to kebab case for the config field to become the CLI flag."

docs/config.md
```diff
@@ -122,14 +126,16 @@ model_names: <comma-delimited-string|list|model>
...
+[ triton_launch_mode: <string> ]
 
-# List of constraints placed on the config search results.
-[ constrains: <constraint> ]
+# Logging level
+[ log_level: <string> | default: INFO ]
 
-# List of objectives that user wants to sort the results by it.
-[ objectives: <objective|list> ]
+# List of GPU UUIDs to be used for the profiling. Use 'all' to profile all the GPUs visible by CUDA."
```
Tabrizian, line 137: Remove the double quote at the end of sentence.

docs/config.md
```diff
@@ -253,18 +281,27 @@ objectives:
...
 objectives:
 - perf_latency
 - perf_throughput
 ```
 
-The above config will multiply obtained throughput and latency for each
-measurement by 0.5 (i.e. 0.5 * latency + 0.5 * throughput) and use this as
-the "score" to sort the measurements according to. This mode can be useful
-if a more advanced criteria needs to be used for sorting.
+The above config is telling model analyzer to compare two measurements by finding relative gain from one measurement to the other, and computing the weighted average of this gain across all listed metrics. In the above example, the relative weights for each metric is equal by default. So if we have two measurements of latency and throughput, model analyzer employs the following logic.
```
Tabrizian, line 293: . -> :

docs/config.md
```diff
@@ -253,18 +281,27 @@ objectives:
...
+
+```python
+measurement_A = (latency_A, throughput_A)
+measurement_B = (latency_B, throughput_B)
+
+gain_A_B = (latency_A - latency_B, throughput_A - throughput_B)
+
+weighted_average_gain = 0.5*(latency_A - latency_B) + 0.5(throughput_A - throughput_B)
```
Tabrizian, line 301: weighted_average_gain = 0.5 * (latency_A - latency_B) + 0.5 * (throughput_A - throughput_B)

Some minor formatting issues.

docs/config.md
```diff
@@ -411,9 +447,27 @@ sizes and two instance group combinations). If both
...
+    perf_analyzer_flags:
+        percentile: 95
+        latency-report-file: /path/to/latency/report/file
+```
+**Important Notes**: 
+1. The section name contains underscores `perf_analyzer_flags` as seen in the example above. However the arguments themselves should contain hyphens as seen with `latency-report-file`.
+2. The Model Analyzer also provides certain arguments to the `perf_analyzer` instances it launches. These ***cannot*** be overriden by providing those arguments in this section. An example of this is `perf_measurement_window`, which is an argument to Model Analyzer itself.
+
+## The `model-names` field and `<model>`
+The `model-names` argument can be provided as a list of strings (names of models) from the CLI interface, or as a more complex `<model>` object but only through the YAML configuration file. The model object can contain `<constraint>`, `<objective>`,
```
Tabrizian, line 469: "The `--model-names` argument..."

docs/config.md
```diff
@@ -411,9 +447,27 @@ sizes and two instance group combinations). If both
...
+`<model-config-parameters>`, `<parameter>` and `<perf_analyzer_flags>`.
```
Tabrizian, line 470: `<perf-analyzer-flags>`

docs/config.md
```diff
@@ -478,7 +561,34 @@ Model Analyzer can be started using the `-f`, or `--config-file` flag.
...
 
-This config will analyze the models using batch sizes equal
-to `[4, 5, 6, 7, 8, 9]` and concurrency values equal to
-`[2, 4, 8]`.  `step` value
-for batch size can also be specified to change the step size.
+It will run the model `vgg_19_graphdef` over combinations of batch sizes `[1,2,3]`, `concurrency` `[2,4,8]` (taken from the global concurrency section), with dynamic batching enabled and preferred batch sizes `2` and `3` and a single CPU instance.
+
+It will also run the model `vgg_16_graphdef` over combinations of batch sizes `[4,5,6,7,8,9]`(taken from the global `batch_sizes` section), concurrency `[2,10,18,26,34,42,50,58]`, with dynamic batching enabled and preferred batch sizes `1` and `2` and a single GPU instance.
+
+## <triton_server_flags>
```
Tabrizian, line 568: `<triton-server-flags>`

docs/config.md
```diff
@@ -142,8 +148,30 @@ model_names: <comma-delimited-string|list|model>
 
 # Disables automatic config search
 [ run_config_search_disable: <bool> | default: false ]
+
+# Specify path to config yaml file
+[ config_file: <string> ]
```
Tabrizian, line 153: Can you also add the `triton_docker_image`?

docs/install.md
```diff
@@ -14,15 +14,54 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-# Install 
+# Installation
 
 There are three ways to use Triton Model Analyzer:
 
-1. Using the `pip`:
+1. Building the Dockerfile:
```
Tabrizian, line 21: I think we can encourage them to use the Triton SDK container instead of building Docker.

docs/install.md
```diff
@@ -14,15 +14,54 @@ See the License for the specific language governing permissions and
...
+   $ git clone https://github.com/triton-inference-server/model_analyzer
+   $ docker build --pull -t model-analyzer .
+   ```
+
+   The above command will pull all the containers that model analyzer needs to run. The Model Analyzer's Dockerfile bases the container on the latest `tritonserver` containers from NGC. Now you can run the container with:
+
    ```
-   $ pip3 install git+https://github.com/triton-inference-server/model_analyzer
+   $ docker run -it --privileged --rm --gpus all \
+        -v /var/run/docker.sock:/var/run/docker.sock \
```
Tabrizian, line 35: I think mounting `docker` makes things very complicated (the path issues, etc). If they use Docker they are encouraged to use the `local` or `remote` mode. But if they want to use `docker` mode, they can use `dind`  or mount the `docker.sock`  (as the last option) with the list of issues that they need to be aware of.

docs/install.md
```diff
@@ -14,15 +14,54 @@ See the License for the specific language governing permissions and
...
+        -v <model_repo_abs_path>:<model_repo_abs_path> \
```
Tabrizian, line 36: \<path-to-triton-model-repository\>

docs/install.md
```diff
@@ -14,15 +14,54 @@ See the License for the specific language governing permissions and
...
+        -v <workspace_path>:/opt/triton-model-analyzer \
```
Tabrizian, line 37: Is this needed?

docs/install.md
```diff
@@ -14,15 +14,54 @@ See the License for the specific language governing permissions and
...
+        -v <output_repo_abs_path>:<output_repo_abs_path> \
```
Tabrizian, line 38: An explanation about why this is needed would be helpful.

docs/install.md
```diff
@@ -14,15 +14,54 @@ See the License for the specific language governing permissions and
...
+        --net=host --name model-analyzer 
+   
+   root@hostname:/opt/triton-model-analyzer# 
    ```
 
+2. Using `pip3`:
+
+   You can install pip using:
+   ```
+   $ sudo apt-get update && sudo apt-get install python3-pip
```
Tabrizian, line 48: They need to install whktmlpdf

docs/install.md
```diff
@@ -33,23 +72,22 @@ There are three ways to use Triton Model Analyzer:
     dpkg -i datacenter-gpu-manager_${DCGM_VERSION}_amd64.deb
    ```
 
-2. Building from source:
+3. Building from source:
 
-   Building from source is similar to installing using pip. You need
-   to install the same dependencies mentioned in the "Using pip section".
-   After that, you can use the following commands:
+   You may sometimes want to build from source, if for example you want to use a
```
Tabrizian, line 77: I think the first sentence is not necessary.

docs/quick_start.md
```diff
@@ -16,64 +16,77 @@ limitations under the License.
 
 # Quick Start
 
-The goal of this section is to analyze a simple libtorch model.
-
-1. Build Model Anlayzer Container:
+The steps below will guide you through using model analyzer to analyze a simple PyTorch model. If you are not using the docker installation, you may skip the first step. The instructions below assume a directory structure like the following:
```
Tabrizian, line 19: It would be great if this doc can be modified to suggest using the SDK container.

docs/report.md
```diff
@@ -0,0 +1,37 @@
+<!--
+Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 2: 2021

docs/report.md
```diff
@@ -0,0 +1,37 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+# Model Analyzer Reports
+
+The Model analyzer can run inferences with models across multiple parameters. It can also generate reports containing common statistics of interest such as the throughput versus latency curve obtained from these inference runs, as well as the best performing model configuration for each model.
```
Tabrizian, line 19: "common statistics of interests" -> "useful plots or common plots"

docs/report.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Requirements
```
Tabrizian, line 21: I think this needs to be moved to install section

Tabrizian (changes requested):

docs/config.md
aramesh7: Ah yes I think I meant to cut and paste that first paragraph but I forgot. Good catch thanks.

docs/config.md
aramesh7: It is already present. The first argument after the perf analyzer related arguments

aramesh7: I don't remember adding it myself

docs/install.md
aramesh7: Does this include if they use the Triton SDK? Is it not the case that the recommended way to use docker in docker is to mount the sock file? We'll be specific about the issues.

docs/config.md
Tabrizian: Yeah, I added it in my PR. I forgot about it :)

docs/install.md
Tabrizian: I think you addressed it correctly in your PR. Thanks!

PR #107, opened 2021-03-31
Find top N model across multiple models

aramesh7:


model_analyzer/analyzer.py
```diff
@@ -149,3 +137,31 @@ def _process_top_results(self):
 
             # Send top_n measurements to report manager
             self._report_manager.add_result(result=result)
+
+    def _save_top_models(self):
```
Tabrizian, line 141: Is this in addition to the output_model_repository that we had?

model_analyzer/config/run/run_config_generator.py
```diff
@@ -13,181 +13,69 @@
...
-                                                    )
-
-                sweeped_dict[key] = sweep_parameter_list
+        return self._run_configs
 
-            # Generate parameter combinations for this field.
-            return self._generate_parameter_combinations(sweeped_dict)
+    def next_config(self):
+        """
+        Returns list of run configs
```
Tabrizian, line 57: Does this return a list of run configs or only the next run config?

model_analyzer/model_manager.py
```diff
@@ -0,0 +1,269 @@
...
+            self._run_model_with_search(model)
+
+        # Sort the results for this model
+        self._result_manager.sort_results()
+
+        # Save the best result
+        [top_result] = self._result_manager.top_n_results(n=1)
+        heapq.heappush(self._model_results, (top_result, model.model_name()))
+
+    def top_n_models(self, n):
```
Tabrizian, line 99: Is this `top_n_models` or `top_n_configs` acrosss all the models.

qa/L0_results/test.sh
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Copyright year

tests/common/test_utils.py
```diff
@@ -69,12 +69,16 @@ def construct_result(avg_gpu_metric_values, avg_non_gpu_metric_values,
     gpu_metric_values = {}
     for gpu_id, metric_values in avg_gpu_metric_values.items():
         gpu_metric_values[gpu_id] = {
-            key: list(range(val - num_vals, val + num_vals))
+            key: list(
+                range(val - value_step * num_vals, val + value_step * num_vals,
```
Tabrizian, line 73: what is the `value_step`? Can you add a comment explaining this?

tests/test_run_search.py
```diff
@@ -44,33 +70,45 @@ def test_run_search(self):
...
-        max_preferred_batch_size = 16
-        max_instance_count = 5
-        run_search = RunSearch(max_concurrency, max_instance_count,
-                               max_preferred_batch_size)
+        args = [
+            'model-analyzer', '--model-repository', 'cli_repository', '-f',
+            'path-to-config-file', '--model-names', 'vgg11'
+        ]
+
+        # Empty yaml
```
Tabrizian, line 95: The YAML doesn't look empty ;)

model_analyzer/model_manager.py
Tabrizian: I think the Bermuda team requested the top n configs across all the models.

model_analyzer/analyzer.py
aramesh7: Yes, this saves the best model configs (across all models) in the export path.

model_analyzer/model_manager.py
aramesh7: From a list of the best configs from each model, this chooses the top n configs.

model_analyzer/config/run/run_config_generator.py
aramesh7: Ah yes I need to change the docstring

docs/config.md
```diff
@@ -73,6 +73,12 @@ model_names: <comma-delimited-string>
 # Generate summary of results
 [ summarize: <boolean>  | default: true ]
 
+# Number of top configs to show in summary plots
+[ configs_shown_per_model: <int> | default: 3]
+
+# Number of top configs to save across ALL models, none saved by default
+[ top_n_configs: <int> | default: 0 ]
```
Tabrizian, line 80: I think it would be better to be backward compatible. As far as I remember, `top_n_configs` used to have different functionality in the previous release. We can use the new parameter (i.e. `configs_per_model`) for the new functionality and keep the old one for the previous functionality.

aramesh7: Technically, this is not yet a feature that we were supposed to use in any stable version, which was why I had hidden it in the autofill. I think the fact that we're just now adding it to the docs gives us some leeway in this regard? Im getting rid of `top_n_configs` altogether and just having `configs_shown_per_model`  and `top_n_models` which is by default across all models.

Tabrizian (changes requested):
As far as I have understood, the current PR it doesn't update the report if the `top_n_models` is specified right? It just saves the top models in the new directory. The bermuda team in their ticket requested the top configs across the models to be included in the summary report. We should add a flag that switches between the top configs per model and across the model. I don't think it is possible to support both of them in a single report.

docs/config.md
```diff
@@ -73,6 +73,12 @@ model_names: <comma-delimited-string>
 # Generate summary of results
 [ summarize: <boolean>  | default: true ]
 
+# Number of top configs to show in summary plots
+[ configs_shown_per_model: <int> | default: 3]
```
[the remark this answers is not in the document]
aramesh7: Understood

docs/config.md
```diff
@@ -73,6 +73,12 @@ model_names: <comma-delimited-string>
...
+
+# Number of top model configs to save across ALL models, none saved by default
+[ top_n_models: <int> | default: 0 ]
```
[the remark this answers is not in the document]
aramesh7: Yes I will do this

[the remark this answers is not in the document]
aramesh7: Yes I will do this

docs/config.md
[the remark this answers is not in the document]
aramesh7: Will do

Tabrizian (approved):
Just to understand, now both the per model and across the models' report will be generated simultaneously?

aramesh7:
> Just to understand, now both the per model and across the models' report will be generated simultaneously? Also, is it possible to remove the directory now. I don't think it is required anymore.

Will remove this functionality, and add tests to check for top models in PR for the report bug fix.

PR #118, opened 2021-04-15
Fix log message

Tabrizian:


aramesh7 (approved):

aramesh7:
Good catch! This was quite annoying to see, but I didn't know the fix was this easy,

PR #121, opened 2021-04-17
Enabled state management

aramesh7:
* Save and Load checkpoints after each model
* Interrupt handling, early exit
* Profiling code decoupled from report generation code
* Report plot monotonicity and table bug fixes


model_analyzer/state/analyzer_state_manager.py
```diff
@@ -0,0 +1,162 @@
...
+import pickle
+import os
+import glob
+
+
+class AnalyzerStateManager:
+    """
+    Maintains the state of the Model Analyzer
+    """
+    def __init__(self, config):
```
Tabrizian, line 30: Can you just pass the config variables required instead of the whole config object?

model_analyzer/result/result_table.py
```diff
@@ -42,6 +42,13 @@ def __init__(self, headers, title=None):
         ]
         self._rows = []
 
+    @staticmethod
+    def from_csvs(filename_server_table, filename_gpu_table,
+                  filename_inference_table):
+        """
+        Create result table from the CSV
+        """
```
Tabrizian, line 50: Is this method empty?

Tabrizian (commented):
It looks like we are not storing the state inside each model sweep. So for example, if a user is in a middle of a very long-running sweep for a single model and MA exits for some reason, the user will still lose all the progress. Just making sure that this is the intended behavior. 

Also, what will happen if the user starts the checkpoint with a different MA config? We can store the config inside the checkpoint too so that MA can be resumed without requiring any additional flag or input.

model_analyzer/state/analyzer_state_manager.py
aramesh7: Sure but is there a particular reason for this?

model_analyzer/result/result_table.py
aramesh7: Oh good catch. I will fix this.

model_analyzer/state/analyzer_state_manager.py
Tabrizian: I think it makes the object's dependencies clear but not a very big deal.

model_analyzer/model_manager.py
```diff
@@ -182,9 +196,17 @@ def _execute_run_configs(self):
...
+            # Check if exiting
+            if self._state_manager.exiting():
+                return measurements
+
             # Remove one run config from the list
             run_config = self._run_config_generator.next_config()
 
+            # If this run config was already run, do not run again
+            if self._already_ran_config(run_config):
+                continue
```
Tabrizian, line 208: I think it should not continue. It should extract the result from the state manager and append it to the list of measurements.

model_analyzer/model_manager.py
```diff
@@ -245,3 +268,23 @@ def _create_and_load_model_variant(self, original_name, variant_config):
...
+
+        results = self._state_manager.get_state_variable(
+            'ResultManager.results')
+
+        # check whether perf config string is a key in result dict
+        if model_name not in results:
+            return False
+        if model_config_name not in results[model_name]:
+            return False
+        return (perf_config_str in results[model_name][model_config_name][1])
```
Tabrizian, line 290: return the result that was found in state manager and return None if it doesn't exist.

model_analyzer/result/result_manager.py
```diff
@@ -229,39 +227,56 @@ def add_measurement(self, run_config, measurement):
...
+        constraints = {}
+        for model in self._config.model_names:
+            comparators[model.model_name()] = ResultComparator(
+                metric_objectives=model.objectives())
+            constraints[model.model_name()] = model.constraints()
+
+        # Construct and add results to individual result heaps as well as global result heap
+        results = self._state_manager.get_state_variable(
+            'ResultManager.results')
+        for model_name, result_dict in results.items():
```
Tabrizian, line 268: Just trying to understand, in the new version, the results will not be added to the result manager until the very end right?

Tabrizian (changes requested):

model_analyzer/result/result_manager.py
aramesh7: Yes, only measurements are added during profiling, and the results are then constructed and sorted after all the runs have finished.

model_analyzer/model_manager.py
aramesh7: Ah okay, this was a dumb error on my part, the run search needs these measurements.

model_analyzer/model_manager.py
aramesh7: Will address

Tabrizian (dismissed):

model_analyzer/analyzer.py
```diff
@@ -114,7 +128,7 @@ def write_and_export_results(self):
             self._plot_manager.compile_and_export_plots()
 
             # Export individual model summaries
-            for model in self._config.model_names:
+            for model in self._config.model_names[:self._model_index]:
```
dzier, line 131: Why does the top level analyzer need a model index? Shouldn't it be analyzing all models?

model_analyzer/config/input/config.py
```diff
@@ -399,7 +398,7 @@ def objective_list_output_mapper(objectives):
                 'max_retries',
                 flags=['-r', '--max-retries'],
                 field_type=ConfigPrimitive(int),
-                default_value=100,
+                default_value=1000,
```
dzier, line 401: It is not a good idea to have random "magic" numbers. A better way is to define a const somewhere. In C++, that may be in a top level `constants.h` file. But in python, there should be a simpilar way. One file that has all important constants. This both encourages reuse of the constant and provides just one place to make the configuration change.

Also, this would fit the pattern needed for the modes. A single parent class/interface that has default values `default.get_max_retries()`, then a different child class for offline and online modes. Then use that instantiated class for default values through out the project.

model_analyzer/constants.py
```diff
@@ -27,3 +27,6 @@
 
 # Reports
 TOP_MODELS_REPORT_KEY = "Best Configs Across All Models"
+
+# State Management
+MAX_NUMBER_OF_INTERRUPTS = 3
```
dzier, line 32: Looks like you have a constants file already. Just add `MAX_RETRIES = 1000` here

model_analyzer/entrypoint.py
```diff
@@ -12,27 +12,23 @@
...
 from .triton.client.client_factory import TritonClientFactory
 from .device.gpu_device_factory import GPUDeviceFactory
 from .config.input.config import AnalyzerConfig
+from .state.analyzer_state_manager import AnalyzerStateManager
 
-logger = logging.getLogger(__name__)
-MAX_NUMBER_OF_INTERRUPTS = 3
+import sys
+import os
+from prometheus_client.parser import text_string_to_metric_families
```
dzier, line 27: Might want to add an empty line before `from` to separate from the imports.

model_analyzer/model_manager.py
```diff
@@ -157,6 +168,9 @@ def _run_model_config_sweep(self,
 
         next_model = model
         while True:
+            if self._state_manager.exiting():
```
dzier, line 171: Will this `while true` conflict with the top level while true loop in the analyze.pl file

model_analyzer/result/result_manager.py
```diff
@@ -314,15 +341,16 @@ def _compile_measurements(self, result):
...
         for (measurements, passes) in [(passing_measurements, True),
                                        (failing_measurements, False)]:
             while measurements:
                 next_best_measurement = heapq.heappop(measurements)
-                self._compile_measurement(model_name=model_name,
-                                          instance_group=instance_group,
-                                          dynamic_batching=dynamic_batching,
-                                          measurement=next_best_measurement,
-                                          passes=passes)
+                self._tabulate_measurement(model_name=model_name,
```
dzier, line 349: By my count, this a quadruple nested loop, if you follow the method call. That doesn't seem algorithmically performant.

dzier (changes requested):

model_analyzer/analyzer.py
aramesh7: Yes, but that keeps track of how many models were profiled, so that only the results for those models are queried in the report generation part.

model_analyzer/model_manager.py
aramesh7: Conflict? How do you mean?

model_analyzer/constants.py
aramesh7: Will fix this.

model_analyzer/result/result_manager.py
aramesh7: There is no repeated work. Do you mean inefficient because of simply the overhead of adding an extra nested loop?

model_analyzer/model_manager.py
dzier: Are you sure you will get an "exiting()" statement? Any `while true` should be a code smell and may result in an infinite while loop. Can this be better served as `while !self._state_manager.exiting():` for example.

model_analyzer/result/result_manager.py
dzier: Right. Is there a way to cache results, for example, from a separate while loop so you don't have to nest. Can you use a sorted list, or hash table? Algorithmically, deeply nested loops are a bad in worst case scenarios. They can also accidentally lead to O(n^2) issues.

model_analyzer/model_manager.py
aramesh7: There is no `do while` construct in python so usually the `while True: if ... break` is what is used. Here you may not get an `exiting()` but the  `get_model_sweep` will eventually return an empty list, which indicates the end of the sweep.

model_analyzer/result/result_manager.py
aramesh7: During collection, each measurement is filled into a nested dictionary 

{model_name : {model_config_name: (model_config, list of measurements)}}

So the data is 3 dimensional inherently. The 4th dimension below is just for convenience to split up passing failing results. 

During report generation we do:
for each model:
    for each config:
          for each of passing/failing:
                for each measurement:
                      ... write to table

dzier (approved):

Tabrizian (approved):

PR #122, opened 2021-04-19
Avoid copying models multiple times

Tabrizian:


aramesh7 (commented):
Is there anyway we could use `os.symlink` here?

Tabrizian:
If I symlink here, I'm not sure whether it will work inside the container too or not.

aramesh7 (approved):

PR #127, opened 2021-04-29
Fixed key error during collect and sort if load_model fails 

aramesh7:


model_analyzer/result/result_manager.py
```diff
@@ -275,18 +275,24 @@ def collect_and_sort_results(self, num_models):
...
-                                     model_config=model_config,
-                                     comparator=comparators[model_name],
-                                     constraints=constraints[model_name])
-                for measurement in measurements.values():
-                    measurement.set_result_comparator(
-                        comparator=comparators[model_name])
-                    result.add_measurement(measurement)
-                self._per_model_sorted_results[model_name].add_result(result)
-                self._across_model_sorted_results.add_result(result)
+            if model_name not in results:
```
Tabrizian, line 278: Why is the `model_name` not profiled until this stage? I think this bug was introduced in the checkpoint PR.

Tabrizian (approved):

PR #131, opened 2021-05-03
Use libtorch models instead of clara models

Tabrizian:


qa/L0_analyze/test.sh
```diff
@@ -20,16 +20,17 @@ rm -rf results && mkdir -p results
...
+QA_MODELS="vgg19_libtorch resnet50_libtorch"
 MODEL_NAMES="$(echo $QA_MODELS | sed 's/ /,/g')"
 BATCH_SIZES="1,2"
 CONCURRENCY="1,2"
 EXPORT_PATH="`pwd`/results"
 FILENAME_SERVER_ONLY="server-metrics.csv"
 FILENAME_INFERENCE_MODEL="model-metrics-inference.csv"
 FILENAME_GPU_MODEL="model-metrics-gpu.csv"
-TRITON_LAUNCH_MODE=${TRITON_LAUNCH_MODE:="docker"}
+TRITON_LAUNCH_MODE=${TRITON_LAUNCH_MODE:="local"}
```
aramesh7, line 33: Is there a reason for this change?

Tabrizian: When using docker mode,  we will run 21.04 by default which will have problems but the local version is 21.03 for now. We can use docker too but we have to fix the container to 21.03.

aramesh7: Ah okay

aramesh7 (approved):

PR #139, opened 2021-05-13
Use count based stabilization for perf_analyzer

Tabrizian:
I added the Triton `format.py` file here too so that we can unify formatting across the files.


model_analyzer/perf_analyzer/perf_config.py
```diff
@@ -142,8 +143,7 @@ def __getitem__(self, key):
         elif key in self._input_to_verbose:
             return self._verbose[self._input_to_verbose[key]]
         else:
-            raise TritonModelAnalyzerException(
-                f"'{key}' Key not found in config")
+            raise KeyError
```
aramesh7, line 146: Do we want to raise a key error if someone uses an improper argument to perf analyzer via `perf_analyzer_flags`?

aramesh7 (changes requested):
It was a little tough to spot the relevant changes. Is there anything that can be done about this?

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -147,17 +143,34 @@ def run(self, metrics):
...
-
                 if process.returncode != 0:
                     if self._output.find(
-                            "Please use a larger time window.") > 0:
-                        self._config['measurement-interval'] += INTERVAL_DELTA
-                        logger.info(
-                            "perf_analyzer's measurement window is too small, "
-                            f"increased to {self._config['measurement-interval']} ms."
-                        )
+                            "Failed to obtain stable measurement") > 0:
```
aramesh7, line 148: This section is tested by `L0_perf_analyzer` , Is there anything could you add to the tests there?

aramesh7 (changes requested):

model_analyzer/perf_analyzer/perf_analyzer.py
Tabrizian: I'll add a unittest for it.

model_analyzer/perf_analyzer/perf_config.py
Tabrizian: Good point. Will revert this change.

aramesh7 (approved):

PR #141, opened 2021-05-14
Fixed snake_case/--kebab-case conversion for custom flags

aramesh7:


Tabrizian:
I understand why this PR is created and I like the unification of case across different config parameters, but I was wondering what can we do if Triton/Perf Analyzer for some reason decides to use snake_case for one of the flags? IMHO, I think we should just pass the flags as is without doing these conversions.

Tabrizian (approved):

PR #142, opened 2021-05-17
Fixing triton server flags and adding plots unittest

aramesh7:


model_analyzer/triton/server/server_config.py
```diff
@@ -23,52 +23,52 @@ class TritonServerConfig:
     """
     server_arg_keys = [
         # Logging
-        'log_verbose',
-        'log_info',
-        'log_warning',
-        'log_error',
+        'log-verbose',
```
Tabrizian, line 26: Why do we need to change the flags again?

aramesh7: I decided to just go with what it was originally, i shouldve made this change in the last one, im sorry. I think its better to have the internal case be kebab case so it doesn't need to be changed 2 times, potentially once when reading the key value in, and once when writing to the command line. It doesn't matter, but i see no reason to have it be snake cased internally.

Tabrizian (approved):

PR #148, opened 2021-05-20
Added LFS and fixed some config defaults and flags

aramesh7:


Tabrizian (approved):

Tabrizian (changes requested):
In the screen shot of the summary report I can see your taskbar. Also, can't we include the PDF of the report instead of the screen shot? Then, we can modify the docs and say that "An example detailed report can be found here".

Tabrizian (approved):

Tabrizian (approved):

PR #150, opened 2021-05-24
Update the note for tracking development

Tabrizian:


README.md
```diff
@@ -18,9 +18,11 @@ limitations under the License.
 
 # Triton Model Analyzer
 
-**NOTE: You are currently on main branch. This branch is tracking 
-development for the new version of Model Analyzer and is not currently 
-usable.**
+**LATEST RELEASE: You are currently on the main branch which tracks
```
dzier, line 21: Is there a reason we need to be this specific in the note? Has this lead to confusion from customers? I was intentionally keeping it more generic for all releases so we do not have to update the note after every release.

Tabrizian: Mahan mentioned to me that it implies Model Analyzer is not stable so I updated this note to avoid the confusion.

/cc @msalehiNV

dzier (approved):

PR #151, opened 2021-05-25
Unify qa tests

kthui:
Incorporated check_results.py and test_config_generator.py into all tests, except L0_server_launch_modes because it is testing the tritonserver and the two py files will do nothing if incorporated.


qa/L0_analyze/check_results.py
```diff
@@ -0,0 +1,142 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Can you please move this file to `qa/common` folder? We can have a generalized version of this file that allows us to check the model analyzer logs and CSV files of any of the tests. It should accept the arguments below:

* --analyzer-log
* --server-metrics-file
* --config-file
* --model-metrics-gpu-file
* --model-metrics-inference-file
* --expected-number-of-row-inference
* --expected-number-of-row-gpu
* --expected-number-of-row-server
* --expected-number-of-column-inference
* --expected-number-of-column-gpu
* --expected-number-of-column-server

Then, each of the tests can call this Python script with different arguments to make sure that the results are correct. Also, it would be good to use https://docs.python.org/3/library/unittest.html for the checks instead of `check_` functions.

If the test requires additional checks, we can use another `py` file that contains additional unit tests.

Tabrizian (changes requested):

qa/L0_analyze/check_results.py
aramesh7: I think the reason `check_` functions are used instead of the usual `unittest` flow is because they require the arguments you described above. Its not idiomatic to the unittest framework to have arguments. However, if we really want to we can turn the arguments into class variables as in [this post](https://stackoverflow.com/questions/11380413/python-unittest-passing-arguments/20702984)

Tabrizian: I see. So, it looks like it is better to stick with the current structure and not use unittests.

kthui: There are some other test cases which don't use the `check_log_table_row_column` and `check_csv_table_row_column`, such as the `L0_perf_analyzer`. Do you mean refactor patterns like 
```
        def print_verification_fail(filename):
            print("\n***\n*** Test Output Verification Failed for " + filename +
                  ".\n***")
            with open(self._analyzer_log, "r") as f:
                print(f.read())

        passed = True

        if not check_log_table_row_column(os.path.abspath(
                self._analyzer_log), server_num_columns, server_num_rows,
                                          "Server\\\\ Only:"):
            print_verification_fail(self._analyzer_log)
            passed = False

        if not check_log_table_row_column(
                os.path.abspath(self._analyzer_log), metrics_num_columns,
                metrics_num_rows, "Models\\\\ \\\\(GPU\\\\ Metrics\\\\):"):
            print_verification_fail(self._analyzer_log)
            passed = False

        if not check_csv_table_row_column(
                os.path.abspath(self._model_metrics_inference_file),
                inference_num_columns, metrics_num_rows, output_tag):
            print_verification_fail(self._model_metrics_inference_file)
            passed = False

        return passed
```
into qa/common and only leave behind things like 
```
        return check1(...) and check2(...) and check3(...)
```
in their respective `check_results.py` files?

Tabrizian: @kthui If the test doesn't use `check_csv_table_row_column`, then I think it can be converted to a unittest. For the tests that do use the `check_csv_table_row_column` or similar the test will we converted to a Python script and it can called like below in the `test.sh`:

```
python3 ../qa/common/check_results.py --analyzer-log <...> ..
if [ $? -ne 0 ]; then
  echo "\n\n\n Test Failed.\n\n\n"
fi
```

qa/L0_config_range/check_results.py
```diff
@@ -0,0 +1,135 @@
...
+
+        def print_verification_fail(filename):
+            print("\n***\n*** Test Output Verification Failed for " + filename +
+                  ".\n***")
+            with open(self._analyzer_log, "r") as f:
+                print(f.read())
+
+        passed = True
+
+        if not check_csv_table_row_column(
```
aramesh7, line 74: Yes seems like these 3 lines can be refactored into a single function call that checks all 3 files.

qa/L0_output_fields/test_config_generator.py
```diff
@@ -0,0 +1,96 @@
...
+    model_config['total_param_server'] = len(
+        model_config['server_output_fields'])
+    model_config['total_param_gpu'] = len(model_config['gpu_output_fields'])
+    model_config['total_param_inference'] = len(
+        model_config['inference_output_fields'])
+    sweep_configs.append(model_config)
+    return sweep_configs
+
+
+def get_all_configurations():
```
aramesh7, line 88: You can delete this and just use `_get_sweep_configs` from `generate_output_fields_test`

qa/common/util.py
```diff
@@ -0,0 +1,117 @@
...
+            "bash", "-c", "source " + util_sh_path +
+            "; get_number_of_rows_csv \"" + str(csv_file) + "\""
+        ],
+                       check=True)
+    except subprocess.CalledProcessError:
+        return False
+    return True
+
+
+def check_log_table_row_column(log_file, expected_num_columns,
```
aramesh7, line 105: These are fine for now, but there must be more readable pythonic ways to implement each of these?

aramesh7 (changes requested):

kthui:
The issues are fixed. The more readable pythonic ways to implement /qa/common/util.py will be implemented in the next PR.

qa/common/util.py
```diff
@@ -0,0 +1,127 @@
...
+            "source " + util_sh_path + "; check_log_table_row_column \"" +
+            str(log_file) + "\" \"" + str(expected_num_columns) + "\" \"" +
+            str(expected_num_rows) + "\" \"" + str(tag) + "\""
+        ],
+                       check=True)
+    except subprocess.CalledProcessError:
+        print("\n***\n*** Test Output Verification Failed for " +
+              os.path.basename(log_file) + ".\n***")
+        return False
+    return True
```
Tabrizian, line 127: Can you add another function here that looks like below:

```
def check_test_results(log_file, expected_num_columns_inference, expected_num_columns_gpu, expected_num_columns_server, expected_num_rows_inference, expected_num_rows_server, expected_num_rows_gpu, gpu_filename_csv, inference_filename_csv, server_filename_csv):
```

This function will use the other functions that are present in this file to make sure that the results are correct. In all other tests, instead of calling `check_csv_table_row_column` and `check_log_table_row_column` separately this new function should be called.

qa/L0_unit_tests/test_config_generator.py
```diff
@@ -0,0 +1,89 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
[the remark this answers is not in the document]
Tabrizian: Also, I think it would be better to keep the `count_tests.py` file because this file doesn't seem to "generate" a unit test. It is only checking the results.

qa/L0_unit_tests/count_tests.py
```diff
@@ -1,58 +0,0 @@
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Please keep this file.

qa/L0_output_fields/check_results.py
```diff
@@ -0,0 +1,121 @@
...
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_GPU_file), metrics_num_columns,
+            test_output_num_rows, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_inference_file),
+            inference_num_columns, test_output_num_rows, output_tag)
+        passed &= check_log_table_row_column(
+            os.path.abspath(self._analyzer_log), server_metrics_num_columns, 1,
+            "Server\\\\ Only:")
+        passed &= check_csv_table_row_column(
```
Tabrizian, line 85: You should only use `check_test_results` instead of all above separate checks. I described the `check_test_results` in the `qa/common`.

qa/L0_export_metrics/check_results.py
```diff
@@ -0,0 +1,106 @@
...
+
+        passed = check_csv_table_row_column(
+            os.path.abspath(self._server_metrics_file),
+            server_metrics_num_columns, metrics_num_rows, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_GPU_file), metrics_num_columns,
+            metrics_num_rows, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_inference_file),
+            inference_num_columns, metrics_num_rows, output_tag)
```
Tabrizian, line 72: You should only use `check_test_results` instead of all above separate checks. I described the `check_test_results` in the `qa/common`.

qa/L0_doa/test.sh
```diff
@@ -17,18 +17,41 @@ source ../common/util.sh
...
+    echo -e "\n***\n*** Test Failed. No config file exists. \n***"
     RET=1
+    exit $RET
 fi
-set -e
+
+# Run the analyzer and check the results
+RET=0
+
+for CONFIG_FILE in ${LIST_OF_CONFIG_FILES[@]}; do
```
Tabrizian, line 36: This test really doesn't need a list of config files. Please revert this to the previous state.

qa/L0_doa/check_results.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Please remove this file.

qa/L0_config_range/check_results.py
```diff
@@ -0,0 +1,117 @@
...
+
+        passed = check_csv_table_row_column(
+            os.path.abspath(self._server_metrics_file),
+            server_metrics_num_columns, 1 * self._num_gpus, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_GPU_file), metrics_num_columns,
+            self._test_output_num_rows * self._num_gpus, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_inference_file),
+            inference_num_columns, self._test_output_num_rows, output_tag)
```
Tabrizian, line 74: You should only use `check_test_results` instead of all above separate checks. I described the `check_test_results` in the `qa/common`.

qa/L0_config/check_results.py
```diff
@@ -0,0 +1,116 @@
...
+
+        passed = check_csv_table_row_column(
+            os.path.abspath(self._server_metrics_file),
+            server_metrics_num_columns, 1 * self._num_gpus, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_GPU_file), metrics_num_columns,
+            self._test_output_num_rows * self._num_gpus, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_inference_file),
+            inference_num_columns, self._test_output_num_rows, output_tag)
```
Tabrizian, line 73: You should only use `check_test_results` instead of all above separate checks. I described the `check_test_results` in the `qa/common`.

qa/L0_analyze/check_results.py
```diff
@@ -0,0 +1,116 @@
...
+            metrics_num_rows, "Models\\\\ \\\\(Inference\\\\):")
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._server_metrics_file), server_num_columns,
+            server_num_rows, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_GPU_file), metrics_num_columns,
+            metrics_num_rows, output_tag)
+        passed &= check_csv_table_row_column(
+            os.path.abspath(self._model_metrics_inference_file),
+            inference_num_columns, metrics_num_rows, output_tag)
```
Tabrizian, line 82: You should only use `check_test_results` instead of all above separate checks. I described the `check_test_results` in the `qa/common`.

Tabrizian (changes requested):

qa/common/util.py
```diff
@@ -0,0 +1,127 @@
...
+            "bash", "-c", "source " + util_sh_path + "; check_test_results \"" +
+            str(test_log) + "\" \"" + str(expected_number_of_test) + "\""
+        ],
+                       check=True)
+    except subprocess.CalledProcessError:
+        print("\n***\n*** Test Output Verification Failed for " +
+              os.path.basename(test_log) + ".\n***")
+        return False
+    return True
+    """
```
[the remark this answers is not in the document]
kthui: It not really supposed to be there, because the functions are to be converted to Python script, as you described above, in the next PR. Maybe I could just skip the next PR and convert with this one.

qa/common/util.py
kthui: Can you please elaborate a bit more on the commonality you see between each L0 tests that uses `check_csv_table_row_column` and/or `check_log_table_row_column` ?

It's because each tests are somewhat distinct from each other. For example, the `L0_export_metrics/check_results.py` calls 
```
check_csv_table_row_column(
    os.path.abspath(self._model_metrics_GPU_file), metrics_num_columns,
    metrics_num_rows, output_tag)
```
while `L0_config/check_results.py` calls 
```
check_csv_table_row_column(
    os.path.abspath(self._model_metrics_GPU_file), metrics_num_columns,
    self._test_output_num_rows * self._num_gpus, output_tag)
```
They are both calling the same `check_csv_table_row_column` function on `model_metrics_GPU_file`, but notice the 3rd parameter is different. Thus, it's unlikely that those tests calling `check_csv_table_row_column` and/or `check_log_table_row_column` can be unified into one function without moving all the logics from each respective L0 tests into the single function. If that's the case, maybe it's better to simply leave the logic in their respective L0 tests?

Tabrizian: @kthui The third argument is taken care of in the `expected_num_columns_gpu` in the `check_results` function that I posted above. Each of these tests will provide a different value for that argument in the check_results. For example, `L0_export_metrics` will use `metrics_num_rows` for `expected_num_columns_gpu` argument and `self._test_output_num_rows * self._num_gpus` when the `check_results` is being used in `L0_config`. So, `check_results` will make sure that both the logs and CSV files have the expected number of rows and columns in each of the tests.

Does it make sense?

kthui: I see, so you are just "combining" `check_csv_table_row_column` and `check_log_table_row_column` functions. Codes like 
```
        passed = check_log_table_row_column(...)
        passed &= check_csv_table_row_column(...)
        return passed
```
will become
```
        passed = check_results(...)
        passed &= check_results(...)
        return passed
```

qa/common/util.py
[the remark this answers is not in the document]
Tabrizian: Doing it in the next PR is ok too.

qa/common/util.py
Tabrizian: @kthui Yes, exactly. Also, I think you don't need two calls to `check_results`. It will be a single call:

```
passed = check_results(...)
return passed.
```

kthui: I see.  I guess the fix is like you initially proposed that a lot of the `check_results.py`, those involving `check_csv_table_row_column` and `check_log_table_row_column`, should be refactored into `qa/common`, and variables like
```
        metrics_num_columns = 11
        inference_num_columns = 10
        server_metrics_num_columns = 5
        output_tag = "Model"
```
should remain in the `test.sh` where they originally came from.
Also their respective `test_config_generator.py`, since there are no different cases to generate, should remain in their original `config_generator.py` form, if configs need to be generated.

qa/L0_unit_tests/test_config_generator.py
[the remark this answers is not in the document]
kthui: Agree

qa/L0_unit_tests/test.sh
```diff
@@ -12,30 +12,49 @@
...
+# Run the unittest and check the results
 RET=0
 
-set +e
-python3 -m unittest discover -v -s ../../tests  -t ../../ > $TEST_LOG 2>&1
-if [ $? -ne 0 ]; then
-    RET=1
-else
-    check_test_results $TEST_LOG $EXPECTED_NUM_TESTS
+for CONFIG_FILE in ${LIST_OF_CONFIG_FILES[@]}; do
```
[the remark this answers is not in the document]
kthui: Agree

qa/L0_unit_tests/check_results.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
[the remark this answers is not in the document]
kthui: Agree

qa/L0_unit_tests/count_tests.py
kthui: Agree

qa/L0_doa/test_config_generator.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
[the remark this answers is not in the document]
kthui: Agree

qa/L0_doa/test.sh
kthui: Agree

qa/L0_doa/check_results.py
kthui: Agree

qa/common/util.py
Tabrizian: Exactly, except that the `output_tag` should also be put into `qa/common`.  The `check_results` function will check all the model analyzer tables and CSV files and use the appropriate tag to find them. 

> Also their respective test_config_generator.py, since there are no different cases to generate, should remain in their original config_generator.py form, if configs need to be generated.

The `test_config_generator` changes look good to me except the tests that don't require it (like L0_doa and L0_unit_tests).

qa/L0_analyze/test_config_generator.py
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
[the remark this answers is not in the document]
kthui: Agree

qa/common/util.py
kthui: I found an easier way to implement this. Since it's only a single function, it can be written as bash inside `qa/common/util.sh`. Then, the python part can be eliminated, and allow the bash scripts to simply call each other.

qa/L0_output_fields/check_results.py
kthui: Refactored

qa/L0_export_metrics/check_results.py
kthui: Refactored

qa/L0_config_range/check_results.py
kthui: Refactored

qa/L0_config/check_results.py
kthui: Refactored

qa/L0_analyze/check_results.py
kthui: Refactored

qa/common/util.py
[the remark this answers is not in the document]
kthui: This part will be moved into `util.sh`. This file will be removed.

qa/common/util.py
kthui: Will not rely on Python script, and complete solely using Bash script. This file will be removed.

qa/L0_output_fields/test_config_generator.py
kthui: Revert to use the original `config_generator.py`

qa/L0_config_range/check_results.py
kthui: Revert to use the original `config_generator.py`

qa/L0_config/test_config_generator.py
```diff
@@ -16,6 +16,50 @@
...
+                get_all_configurations(self.profile_models)):
+            total_param = configuration['total_param']
+            del configuration['total_param']
+            with open(f'./config-{i}.yml', 'w') as file:
+                yaml.dump(configuration, file)
+            with open(f'./config-{i}.txt', 'w') as file:
+                file.write(str(total_param))
+
+
 def _get_sweep_configs(profile_models):
```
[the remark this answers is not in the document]
kthui: Revert to use the original `config_generator.py`

qa/L0_config/check_results.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    Functions that validate the output
+    of the test
+    """
+
+    def __init__(self, config, config_file, analyzer_log, server_metrics_file,
+                 model_metrics_GPU_file, model_metrics_inference_file,
+                 test_output_num_rows, num_gpus):
+        self._config = config
+        self._config_file = config_file
```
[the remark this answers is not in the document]
kthui: Will use Bash script only to check result, so this file will be removed.

kthui:
The issues are fixed.

qa/L0_analyze/test.sh
```diff
@@ -102,4 +75,4 @@ else
     echo -e "\n***\n*** Test FAILED\n***"
 fi
 
-exit $RET
+exit $RET
```
Tabrizian, line 78: Add a new line to the end of file.

qa/L0_config/config_generator.py
```diff
@@ -101,4 +101,4 @@ def get_all_configurations(profile_models):
         with open(f'./config-{i}.yml', 'w') as file:
             yaml.dump(configuration, file)
         with open(f'./config-{i}.txt', 'w') as file:
-            file.write(str(total_param))
+            file.write(str(total_param))
```
Tabrizian, line 104: New line.

qa/L0_config/test.sh
```diff
@@ -119,4 +110,4 @@ else
     echo -e "\n***\n*** Test FAILED\n***"
 fi
 
-exit $RET
+exit $RET
```
Tabrizian, line 113: New line.

qa/L0_config_range/config_generator.py
```diff
@@ -123,4 +123,4 @@ def get_all_configurations():
             if isinstance(config['profile_models'], str):
                 file.write(config['profile_models'])
             else:
-                file.write(','.join(config['profile_models']))
+                file.write(','.join(config['profile_models']))
```
Tabrizian, line 126: New line.

qa/L0_config_range/test.sh
```diff
@@ -120,4 +111,4 @@ else
     echo -e "\n***\n*** Test FAILED\n***"
 fi
 
-exit $RET
+exit $RET
```
Tabrizian, line 114: New line.

qa/L0_config_search/config_generator.py
```diff
@@ -212,4 +212,4 @@ def get_all_configurations(profile_models):
         with open(f'./config-{i}-models-remote.txt', 'w') as file:
             file.write(str(total_models_remote))
         with open(f'./config-{i}.yml', 'w') as file:
-            yaml.dump(configuration, file)
+            yaml.dump(configuration, file)
```
Tabrizian, line 215: New line.

Tabrizian: Please keep the new line at the end of all the files.

qa/common/util.sh
```diff
@@ -213,6 +213,73 @@ function check_log_table_row_column() {
     return 0
 }
 
+# Check the table(s) by calling check_log_table_row_column and check_csv_table_row_column
+# See if the logic generalizes to the specify testing needs
+# Call check_log_table_row_column and check_csv_table_row_column directly if not
+function check_table_row_column() {
```
Tabrizian, line 219: Can you rename this to `check_analyzer_results`?

Tabrizian (changes requested):
Minor comments. Rest LGTM.

qa/L0_analyze/test.sh
kthui: Adjusted

qa/L0_config/config_generator.py
kthui: Adjusted

qa/L0_config/test.sh
kthui: Adjusted

qa/L0_config_range/config_generator.py
kthui: Adjusted

qa/L0_config_range/test.sh
kthui: Adjusted

qa/L0_config_search/config_generator.py
kthui: Adjusted

qa/common/util.sh
kthui: Yes

kthui:
The issues are fixed.

kthui:
The issues are fixed.

Tabrizian (dismissed):

aramesh7 (commented):
You want the commits for the PR to be few in number and descriptive. Can you rebase and squash all the commits into a few descriptive commits? I suppose now you should probably squash them into 1 commit.

kthui:
```
> git diff 2bdf2d7bbc4d6ac59649d69d55cb833196d9ac55 6d1c62ff9dfc56ac25a680c6ad9bc0b52ee79784
diff --git a/Dockerfile b/Dockerfile
index 46609eb..f1efe41 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ ENV DCGM_VERSION=2.0.13
 ENV DEBIAN_FRONTEND=noninteractive

 RUN apt-get update && \
-    apt-get install -y python3-dev wkhtmltopdf
+    apt-get install -y python3-dev

 RUN mkdir -p /opt/triton-model-analyzer

@@ -49,6 +49,7 @@ WORKDIR /opt/triton-model-analyzer
 RUN rm -fr *
 COPY --from=sdk /usr/local/bin/perf_analyzer .
 RUN chmod +x ./perf_analyzer
+
 COPY . .
 RUN chmod +x /opt/triton-model-analyzer/nvidia_entrypoint.sh
 RUN chmod +x build_wheel.sh && \
@@ -58,6 +59,9 @@ RUN python3 -m pip install --upgrade pip && \
     python3 -m pip install nvidia-pyindex && \
     python3 -m pip install wheels/triton_model_analyzer-*-manylinux1_x86_64.whl

+RUN apt-get install -y wkhtmltopdf
+
 ENTRYPOINT ["/opt/triton-model-analyzer/nvidia_entrypoint.sh"]
 ENV MODEL_ANALYZER_VERSION ${MODEL_ANALYZER_VERSION}
 ENV MODEL_ANALYZER_CONTAINER_VERSION ${MODEL_ANALYZER_CONTAINER_VERSION}
+
```

kthui:
```
> git diff 6d1c62ff9dfc56ac25a680c6ad9bc0b52ee79784 3d3fc1bb582b3a99b5361dcbf08715a8af5108b4
>
```

aramesh7 (approved):

PR #154, opened 2021-06-02
Update report manager unit test

kthui:


tests/test_report_manager.py
```diff
@@ -44,6 +53,29 @@ def _evaluate_config(self, args, yaml_content):
         mock_config.stop()
         return config
 
+    def _setUp_managers(self,
```
aramesh7, line 56: If you have snake_case, you don't want to mix in camelCase. You can change this to `_init_managers`

tests/test_report_manager.py
```diff
@@ -55,98 +87,112 @@ def setUp(self):
...
-        """
-        config = self._evaluate_config(args, yaml_content)
-        self.report_manager = ReportManager(config=config)
+        self.perf_config = PerfAnalyzerConfig()
 
     def test_add_results(self):
-        objective_spec = {'perf_throughput': 10}
-        self.result_comparator = ResultComparator(
-            metric_objectives=objective_spec)
+        self._setUp_managers("test_model1,test_model2")
```
aramesh7, line 103: There are functions in `tests/common/test_utils.py` that construct measurements and results for you. Are you unable to use those for some reason?

tests/test_report_manager.py
```diff
@@ -55,98 +87,112 @@ def setUp(self):
...
+            model_config = ModelConfig.create_from_dictionary(self.model_config)
+            run_config = RunConfig("test_model2", model_config,
+                                   self.perf_config)
+
+            avg_non_gpu_metrics = [
+                PerfThroughput(200 + 10 * i),
+                PerfLatency(4000),
+                CPUUsedRAM(1000)
+            ]
+            measurement = Measurement(gpu_data=avg_gpu_metrics,
```
aramesh7, line 138: Here as well

tests/test_report_manager.py
```diff
@@ -55,98 +87,112 @@ def setUp(self):
...
+                                      non_gpu_data=avg_non_gpu_metrics,
+                                      perf_config=self.perf_config)
+            measurement.set_result_comparator(result_comparator)
+
+            self.result_manager.add_measurement(run_config, measurement)
+
+        self.result_manager.compile_and_sort_results()
+        with unittest.mock.patch(
```
aramesh7, line 146: Look into using the `mock_io.py` classes. You can instantiate the io_mock in the setUp function, just like we do with all the other mocks. Look at `test_state_manager.py` to see if you can use that here as well.

tests/test_report_manager.py
```diff
@@ -55,98 +87,112 @@ def setUp(self):
...
-            model_config = ModelConfig.create_from_dictionary(
-                self.model_config)
-            self.report_manager.add_result(
-                report_key='test_report',
-                result=construct_result(
-                    avg_gpu_metric_values=avg_gpu_metrics,
-                    avg_non_gpu_metric_values=avg_non_gpu_metrics,
-                    comparator=self.result_comparator,
-                    model_config=model_config))
+            self.model_config["name"] = f"model_{i}"
```
aramesh7, line 168: Even if you cannot use the `test_utils.py` functions, there seems to be a lot of repeated code here, perhaps you can refactor the measurement creation/addition into a single function.

aramesh7 (changes requested):

tests/test_report_manager.py
kthui: Changed

tests/test_report_manager.py
kthui: There was a problem with perf_config, but I found a way to get around it. I will use the `test_utils.py`.

tests/test_report_manager.py
kthui: Will use the `test_utils.py`.

tests/test_report_manager.py
kthui: `test_utils.py` function will be called

tests/test_report_manager.py
kthui: Updated

tests/test_report_manager.py
```diff
@@ -12,141 +12,174 @@
...
-            self.report_manager.add_result(
-                report_key='test_report1',
-                result=construct_result(
-                    avg_gpu_metric_values=avg_gpu_metrics,
-                    avg_non_gpu_metric_values=avg_non_gpu_metrics,
-                    comparator=self.result_comparator))
+            self.model_config["name"] = f"test_model1_report_{i}"
+
+            measurement = construct_measurement("test_model1", avg_gpu_metrics,
+                                                avg_non_gpu_metrics,
```
aramesh7, line 114: Can you refactor the creation of this and the next report out into one function called `_add_report_measurements` or something like that. Let it take the model name and any other important paramters as arguments.

aramesh7 (changes requested):
Last thing, otherwise it looks really good!

tests/test_report_manager.py
kthui: Sure

kthui:
Updated

aramesh7 (approved):

PR #156, opened 2021-06-03
Updated helm chart

aramesh7:


dzier (changes requested):

docs/kubernetes_deploy.md
```diff
@@ -130,12 +182,12 @@ REVISION: 1
 TEST SUITE: None
 ```
 
-Check that the model analyzer pod is running and that it has two containers.
+Check that the model analyzer pod is running.
```
Tabrizian, line 185: Why did it use to require to containers? Is that one of them was used for Triton Server and the other one was used for Triton Server?

helm-chart/templates/config-map.yaml
```diff
@@ -0,0 +1,64 @@
...
+    run_config_search_disable: True
+    triton_http_endpoint: localhost:8000
+    triton_grpc_endpoint: localhost:8001
+    triton_metrics_url: http://localhost:8002/metrics
+
+    concurrency: 1,2
+    batch_sizes: 1
+
+    profile_models: 
+      resnet50_libtorch:
```
Tabrizian, line 36: I think it would be better to use the example add sub model so that the users can really get started with helm deployment based on the models currently available to them.

Tabrizian (changes requested):

docs/kubernetes_deploy.md
aramesh7: Yes

docs/kubernetes_deploy.md
```diff
@@ -51,28 +51,9 @@ kube-system   kube-scheduler-user.nvidia.com                   1/1     Running
 kube-system   nvidia-device-plugin-1607379880-dblhc            1/1     Running   0          11m
 ```
 
-Before deploying the model analyzer, you must specify the directories that the model analyzer container will mount and use in `helm-chart/values.yaml`.
+Before deploying the model analyzer, the directories that the container will mount and use must be specified in `helm-chart/values.yaml`.
```
Tabrizian, line 54: "mount and use" is redundant. "that the container will mount" should be sufficient.

Tabrizian (changes requested):
Minor comment, rest LGTM.

Tabrizian (approved):

dzier (approved):

PR #157, opened 2021-06-03
Shorten long CI tests

aramesh7:


qa/L0_config_search/config_generator.py
```diff
@@ -113,77 +126,9 @@ def _get_sweep_configs(profile_models):
     model_config['total_models_remote'] = 1
     sweep_configs.append(model_config)
 
-    model_config = {
```
Tabrizian, line 116: Why do you remove the test? Each of them is covering one corner case in the L0_config_search. I don't think we should remove tests to make them shorter.

Tabrizian (changes requested):
I think the plan for shortening the CI tests was to use a checkpoint so that the run times are reduced. Removing tests does not seem like a good idea for shortening the test time.

qa/L0_config_search/config_generator.py
aramesh7: When the run config search is disabled, the code path only forks based on launch mode, which is covered in test.sh. I think those tests were redundant. I only removed tests where the run config search was disabled.

Tabrizian (commented):
Can we move the checkpoints file to the model analyzer folder in dlcluster? I think it is better to not store binary files inside git.

model_analyzer/device/gpu_device_factory.py
```diff
@@ -150,30 +150,44 @@ def create_device_by_uuid(uuid, dcgmPath=None):
...
+
+        Returns
+        -------
+        dict
+            keys are gpu uuids
+            values are device ids on this machine
         """
 
-        model_analyzer_gpus = []
+        cuda_visible_gpus = []
```
Tabrizian, line 184: Why not make this a dictionary from the beginning?

model_analyzer/record/metrics_manager.py
```diff
@@ -160,10 +172,15 @@ def _start_monitors(self, cpu_only=False):
...
-            self._dcgm_monitor.start_recording_metrics()
+            try:
+                self._dcgm_monitor = DCGMMonitor(
+                    self._gpus, self._config.monitoring_interval,
+                    self._dcgm_metrics)
+                self._check_triton_and_model_analyzer_gpus()
+                self._dcgm_monitor.start_recording_metrics()
+            except TritonModelAnalyzerException as e:
+                self._destroy_monitors()
+                raise TritonModelAnalyzerException(e)
```
Tabrizian, line 183: I think you can just use `raise` to reraise the exception.

Tabrizian (changes requested):

model_analyzer/entrypoint.py
```diff
@@ -92,9 +90,14 @@ def get_server_handle(config):
...
         triton_config['grpc-port'] = config.triton_grpc_endpoint.split(':')[-1]
-        triton_config['metrics-port'] = urlparse(config.triton_metrics_url).port
+        triton_config['metrics-port'] = urlparse(
+            config.triton_metrics_url).port
         triton_config['model-control-mode'] = 'explicit'
         logging.info('Starting a local Triton Server...')
+        if len(config.gpus) >= 1 and config.gpus[0] != 'all':
+            visible_gpus = GPUDeviceFactory.get_cuda_visible_gpus()
+            os.environ['CUDA_VISIBLE_DEVICES'] = ','.join(
+                [visible_gpus[uuid] for uuid in config.gpus])
```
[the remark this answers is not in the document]
aramesh7: yes I will do this

model_analyzer/record/metrics_manager.py
aramesh7: ah thank you!

model_analyzer/perf_analyzer/perf_config.py
```diff
@@ -96,16 +96,46 @@ def update_config(self, params=None):
...
 
-    def to_cli_string(self, remove_url=False):
+    def representation(self):
         """
-        Utility function to convert a config into a
-        string of arguments to the perf_analyzer with CLI.
+        Returns
+        -------
+        str
+            a string representation that Does not include the url
```
Tabrizian, line 104: Does -> does

model_analyzer/model_manager.py
```diff
@@ -276,7 +275,11 @@ def _get_measurement_if_config_duplicate(self, run_config):
         if model_config_name not in results[model_name]:
             return False
         measurements = results[model_name][model_config_name][1]
-        if perf_config_str in measurements:
+
+        # For backward compatibility
```
Tabrizian, line 279: Can you elaborate on the backward computability that is being covered here?

aramesh7: Resolved in comment

Tabrizian (approved):

PR #163, opened 2021-06-10
Update README and versions for 21.06 branch

dzier:
REVIEW ONLY, DO NOT MERGE.


aramesh7 (changes requested):
One of the main places that need updating is
[here](https://github.com/triton-inference-server/model_analyzer/blob/64aba2e3b46dfa730706afa8537da8ec2e282b1e/model_analyzer/config/input/config_defaults.py#L40)

PR #164, opened 2021-06-11
Offline mode for analyze and report

aramesh7:
Offline summary
[result_summary_offline.pdf](https://github.com/triton-inference-server/model_analyzer/files/6636427/result_summary_offline.pdf)

Offline detailed report
[detailed_report_libtorch_offline.pdf](https://github.com/triton-inference-server/model_analyzer/files/6640632/detailed_report_libtorch_offline.pdf)
 
 online equivalents
[detailed_report_online.pdf](https://github.com/triton-inference-server/model_analyzer/files/6640923/detailed_report_online.pdf)
[result_summary_online.pdf](https://github.com/triton-inference-server/model_analyzer/files/6640924/result_summary_online.pdf)


model_analyzer/config/input/config_command_analyze.py
```diff
@@ -276,6 +276,14 @@ def _add_shorthand_configs(self):
                 description=
                 "Shorthand flag for specifying a maximum latency in ms."))
 
+        self._add_config(
+            ConfigField(
+                'min_throughput',
+                flags=['--min-throughput'],
+                field_type=ConfigPrimitive(int),
+                description=
+                "Shorthand flag for specifying a maximum latency in ms."))
```
Tabrizian, line 285: Description is incorrect

model_analyzer/config/input/config_command_analyze.py
```diff
@@ -369,6 +380,18 @@ def _autofill_values(self):
                         'max': self.latency_budget
                     }})
 
+        # Set global constraints if latency budget is specified
+        if self.min_throughput:
```
Tabrizian, line 384: if `min_throughput` is specified and equal to zero this will not work.

model_analyzer/analyzer.py
```diff
@@ -136,6 +141,7 @@ def analyze(self):
         if not gpu_info:
             gpu_info = {}
         self._report_manager = ReportManager(
```
Tabrizian, line 143: Why do we need `ReportManager` in analyze mode?

model_analyzer/analyzer.py
```diff
@@ -118,12 +118,17 @@ def profile(self, client):
             f"Finished profiling. Obtained measurements for models: {profiled_model_list}."
         )
 
-    def analyze(self):
+    def analyze(self, mode):
```
Tabrizian, line 121: I don't think `mode` is required for `analyze`.

Tabrizian (changes requested):

dzier (approved):

model_analyzer/analyzer.py
dzier: It is. There are different set of presets and parameters depending on mode.

Tabrizian:
Also, in the detailed report for the offline mode, I think the batch size should be included otherwise the difference is not clear between each of the rows.

model_analyzer/analyzer.py
Tabrizian: Yes, I understood that. But I think `mode` doesn't change anything in the `analyze` mode so I was wondering why it is required.

model_analyzer/analyzer.py
aramesh7: The analyze mode is what generates the summaries.

model_analyzer/analyzer.py
aramesh7: It changes the summary a little bit. Shows different plots.

model_analyzer/config/input/config_command_analyze.py
aramesh7: Im thinking about it, and if the min throughput is 0, and this branch is not taken, its equivalent to not having the constraint since the throughput must be positive.

Tabrizian (dismissed):

Tabrizian (dismissed):

dzier (approved):

PR #167, opened 2021-06-17
Improve triton server log capture methods

aramesh7:


model_analyzer/triton/server/server_local.py
```diff
@@ -67,10 +69,14 @@ def start(self):
...
                     [visible_gpus[uuid] for uuid in self._gpus])
 
+            if self._log_path:
+                self._log_file = open(self._log_path, 'w+')
+            else:
+                self._log_file = StringIO()
             self._tritonserver_process = Popen(cmd,
-                                               start_new_session=True,
-                                               stdout=PIPE,
+                                               stdout=DEVNULL,
```
Tabrizian, line 77: When we output everything to `DEVNULL` when are we going to use the `self._log_file`?

aramesh7: Yep thats leftover, will remove it.

Tabrizian (dismissed):

tests/test_triton_server.py
```diff
@@ -172,6 +173,7 @@ def test_start_stop_gpus(self):
         self.server.stop()
         self.server_local_mock.assert_server_process_terminate_called()
 
+    @skip('May not be valid')
```
Tabrizian, line 176: Why are you skipping this test?

aramesh7: Ah I had skipped it because the logs function was not unit testable. It is tested in L0_custom_flags and L0_server_launch_modes so perhaps this can be deleted

Tabrizian (approved):

PR #171, opened 2021-06-21
Changed serialization library to json for security

aramesh7:


model_analyzer/perf_analyzer/perf_config.py
```diff
@@ -95,6 +95,10 @@ def update_config(self, params=None):
             for key in params:
                 self[key] = params[key]
 
+    def deserialize(self, perf_config_dict):
```
Tabrizian, line 98: What about using a static `from_dict` method that returns an object using the state dictionary?

aramesh7: Yes I think this makes sense.

model_analyzer/perf_analyzer/perf_config.py
```diff
@@ -95,6 +95,13 @@ def update_config(self, params=None):
             for key in params:
                 self[key] = params[key]
 
+    @classmethod
+    def from_dict(cls, perf_config_dict):
+        perf_config = PerfAnalyzerConfig()
+        for key, val in perf_config_dict.items():
+            setattr(perf_config, key, val)
```
Tabrizian, line 102: Also, I think it would be good to not use `setattr`. You can extract the fields that you need from the dictionary and construct the PerfAnalyzerConfig using them.

aramesh7: Is there a reason you think I should't use setattr? It might be more explicit but it also introduces one more place to make changes if I we need to edit the PerAnalyzerConfig.

Tabrizian (approved):

PR #175, opened 2021-06-30
Added parameters to allow control over perf_analyzer retries

aramesh7:


docs/config.md
```diff
@@ -106,6 +106,12 @@ profile_models: <comma-delimited-string-list>
 # Enables writing the output from the perf_analyzer to stdout.
 [ perf_output: <bool> | default: false ]
 
+# Minimum number of requests to be collected in each measurement window when 'count_windows' mode is used.
+[ perf_analyzer_minimum_request_count: <int> | default: 100 ]
```
Tabrizian, line 110: I believe they can modify this inside the `perf_analyzer_flags`.

aramesh7: Yes, they can modify it but I thought it best to force a default from the model analyzer side. if you think its unnecessary I will remove it.

Tabrizian: Does it work properly when they provide it in the `perf_analyzer_flags` too? I think it would be good to not have duplicate parameters controlling the same setting.

aramesh7: Understood It will be removed

Tabrizian (approved):

PR #176, opened 2021-07-01
Remove dependency dir being set to empty string

jishminor:
Not sure the intention of the original, but dependency dir was being set to an empty string rather than `pwd`/wheel as is specified by `build_wheel.sh`. This resulted in failure to copy over perf_analyzer binary in setup.py


jishminor:
Did not see: https://github.com/triton-inference-server/model_analyzer/pull/174. Closing.

PR #178, opened 2021-07-08
Updated default monitor frequency to 1

kthui:
The default monitor frequency is changed to 1, to reduce the load of cpu-monitor on the processor. It is because higher load on the processor might affect the result of the perf_analyzer.


Tabrizian (approved):

kthui:
Waiting for CI. Will merge after it is completed.

PR #181, opened 2021-07-13
Fix model analyzer to launch server with no GPU when empty gpus config is used

kthui:
If the user sets the `gpus: []` config for model-analyzer, then this implies the model-analyzer should not use any GPU. This is a fix that the model-analyzer will launch the triton server without any GPU, when that config is present. 

This pull request also unifies how the `gpus` config is processed between local and docker mode. Previously, they are scattered across `entrypoint.py`, `server_local.py` and `server_docker.py`. This unification will process the `gpus` config inside `server_local.py` and `server_docker.py` using the `GPUDeviceFactory` class in `gpu_device_factory.py` which will return a list of GPU UUIDs that should be used based on the `gpus` config. Then, the `server_local.py` and `server_docker.py` will launch the server according to the returned list of GPU UUIDs.

### DIFF
--- a/model_analyzer/entrypoint.py
+++ b/model_analyzer/entrypoint.py
@@ -18,7 +18,6 @@ from .model_analyzer_exceptions import TritonModelAnalyzerException
 from .triton.server.server_factory import TritonServerFactory
 from .triton.server.server_config import TritonServerConfig
 from .triton.client.client_factory import TritonClientFactory
-from .device.gpu_device_factory import GPUDeviceFactory
 from .state.analyzer_state_manager import AnalyzerStateManager
 from .config.input.config_command_profile import ConfigCommandProfile
 from .config.input.config_command_analyze import ConfigCommandAnalyze
@@ -73,8 +72,8 @@ def get_server_handle(config):
         logging.info('Using remote Triton Server...')
         server = TritonServerFactory.create_server_local(path=None,
                                                          config=triton_config,
-                                                         gpus=None,
-                                                         log_path=None)
+                                                         gpus=[],
+                                                         log_path="")
         logging.warning(
             'GPU memory metrics reported in the remote mode are not'
             ' accuracte. Model Analyzer uses Triton explicit model control to'
@@ -92,8 +91,7 @@ def get_server_handle(config):
         triton_config['model-repository'] = config.output_model_repository_path
         triton_config['http-port'] = config.triton_http_endpoint.split(':')[-1]
         triton_config['grpc-port'] = config.triton_grpc_endpoint.split(':')[-1]
-        triton_config['metrics-port'] = urlparse(
-            config.triton_metrics_url).port
+        triton_config['metrics-port'] = urlparse(config.triton_metrics_url).port
         triton_config['model-control-mode'] = 'explicit'
         logging.info('Starting a local Triton Server...')
         server = TritonServerFactory.create_server_local(
@@ -108,14 +106,13 @@ def get_server_handle(config):
             config.output_model_repository_path)
         triton_config['http-port'] = config.triton_http_endpoint.split(':')[-1]
         triton_config['grpc-port'] = config.triton_grpc_endpoint.split(':')[-1]
-        triton_config['metrics-port'] = urlparse(
-            config.triton_metrics_url).port
+        triton_config['metrics-port'] = urlparse(config.triton_metrics_url).port
         triton_config['model-control-mode'] = 'explicit'
         logging.info('Starting a Triton Server using docker...')
         server = TritonServerFactory.create_server_docker(
             image=config.triton_docker_image,
             config=triton_config,
-            gpus=GPUDeviceFactory.verify_requested_gpus(config.gpus),
+            gpus=config.gpus,
             log_path=config.triton_output_path)
     else:
         raise TritonModelAnalyzerException(
@@ -178,10 +175,9 @@ def get_cli_and_config_options():
             help=
             'Collect and sort profiling results and generate data and summaries.',
             config=config_analyze)
-        cli.add_subcommand(
-            cmd='report',
-            help='Generate detailed reports for a single config',
-            config=config_report)
+        cli.add_subcommand(cmd='report',
+                           help='Generate detailed reports for a single config',
+                           config=config_report)
         return cli.parse()
 
     except TritonModelAnalyzerException as e:
--- a/model_analyzer/triton/server/server_docker.py
+++ b/model_analyzer/triton/server/server_docker.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020,21 NVIDIA CORPORATION. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -41,8 +41,8 @@ class TritonServerDocker(TritonServer):
             The tritonserver docker image to pull and run
         config : TritonServerConfig
             the config object containing arguments for this server instance
-        gpus : list
-            list of GPUs to be used
+        gpus : list of str
+            List of GPU UUIDs to be mounted in the container
         log_path: str
             Absolute path to the triton log file
         """
@@ -62,11 +62,9 @@ class TritonServerDocker(TritonServer):
         Starts the tritonserver docker container using docker-py
         """
 
-        if len(self._gpus) == 1 and self._gpus[0] == 'all':
-            devices = [
-                docker.types.DeviceRequest(count=-1, capabilities=[['gpu']])
-            ]
-        else:
+        # List GPUs to be mounted inside docker container
+        devices = []
+        if len(self._gpus):
             devices = [
                 docker.types.DeviceRequest(device_ids=self._gpus,
                                            capabilities=[['gpu']])
--- a/model_analyzer/triton/server/server_docker.py
+++ b/model_analyzer/triton/server/server_docker.py
@@ -42,7 +42,7 @@ class TritonServerDocker(TritonServer):
         config : TritonServerConfig
             the config object containing arguments for this server instance
         gpus : list of str
-            List of GPU UUIDs to be mounted in the container
+            List of GPU UUIDs to be mounted and used in the container
         log_path: str
             Absolute path to the triton log file
         """
@@ -62,13 +62,16 @@ class TritonServerDocker(TritonServer):
         Starts the tritonserver docker container using docker-py
         """
 
-        # List GPUs to be mounted inside docker container
+        # List GPUs to be mounted and used inside docker container
         devices = []
         if len(self._gpus):
             devices = [
                 docker.types.DeviceRequest(device_ids=self._gpus,
                                            capabilities=[['gpu']])
             ]
+        environment = {
+            'CUDA_VISIBLE_DEVICES': ','.join([uuid for uuid in self._gpus])
+        }
 
         # Mount required directories
         volumes = {
@@ -97,6 +100,7 @@ class TritonServerDocker(TritonServer):
                 name='tritonserver',
                 image=self._tritonserver_image,
                 device_requests=devices,
+                environment=environment,
                 volumes=volumes,
                 ports=ports,
                 publish_all_ports=True,
--- a/model_analyzer/triton/server/server_local.py
+++ b/model_analyzer/triton/server/server_local.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020,21 NVIDIA CORPORATION. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from .server import TritonServer
-from model_analyzer.device.gpu_device_factory import GPUDeviceFactory
 from model_analyzer.constants import SERVER_OUTPUT_TIMEOUT_SECS
 from model_analyzer.model_analyzer_exceptions \
     import TritonModelAnalyzerException
@@ -31,6 +30,7 @@ class TritonServerLocal(TritonServer):
     Concrete Implementation of TritonServer interface that runs
     tritonserver locally as as subprocess.
     """
+
     def __init__(self, path, config, gpus, log_path):
         """
         Parameters
@@ -40,7 +40,7 @@ class TritonServerLocal(TritonServer):
         config : TritonServerConfig
             the config object containing arguments for this server instance
         gpus: list of str
-            List of strings of GPU UUIDs that should be made visible to Triton
+            List of GPU UUIDs to be made visible to Triton
         log_path: str
             Absolute path to the triton log file
         """
@@ -62,13 +62,11 @@ class TritonServerLocal(TritonServer):
         if self._server_path:
             # Create command list and run subprocess
             cmd = [self._server_path]
-            cmd += self._server_config.to_cli_string().replace('=',
-                                                               ' ').split()
+            cmd += self._server_config.to_cli_string().replace('=', ' ').split()
+            # List GPUs to be used by tritonserver
             triton_env = os.environ.copy()
-            if len(self._gpus) >= 1 and self._gpus[0] != 'all':
-                visible_gpus = GPUDeviceFactory.get_cuda_visible_gpus()
-                triton_env['CUDA_VISIBLE_DEVICES'] = ','.join(
-                    [visible_gpus[uuid] for uuid in self._gpus])
+            triton_env['CUDA_VISIBLE_DEVICES'] = ','.join(
+                [uuid for uuid in self._gpus])
 
             if self._log_path:
                 try:
--- a/qa/L0_server_launch_modes/test.sh
+++ b/qa/L0_server_launch_modes/test.sh
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020,21 NVIDIA CORPORATION. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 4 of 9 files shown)


model_analyzer/entrypoint.py
```diff
@@ -92,8 +91,7 @@ def get_server_handle(config):
         triton_config['model-repository'] = config.output_model_repository_path
         triton_config['http-port'] = config.triton_http_endpoint.split(':')[-1]
         triton_config['grpc-port'] = config.triton_grpc_endpoint.split(':')[-1]
-        triton_config['metrics-port'] = urlparse(
-            config.triton_metrics_url).port
+        triton_config['metrics-port'] = urlparse(config.triton_metrics_url).port
```
kthui, line 94: This is updated by the styling script.

model_analyzer/entrypoint.py
```diff
@@ -108,14 +106,13 @@ def get_server_handle(config):
             config.output_model_repository_path)
         triton_config['http-port'] = config.triton_http_endpoint.split(':')[-1]
         triton_config['grpc-port'] = config.triton_grpc_endpoint.split(':')[-1]
-        triton_config['metrics-port'] = urlparse(
-            config.triton_metrics_url).port
+        triton_config['metrics-port'] = urlparse(config.triton_metrics_url).port
```
kthui, line 109: This is updated by the styling script.

model_analyzer/entrypoint.py
```diff
@@ -178,10 +175,9 @@ def get_cli_and_config_options():
             help=
             'Collect and sort profiling results and generate data and summaries.',
             config=config_analyze)
-        cli.add_subcommand(
-            cmd='report',
-            help='Generate detailed reports for a single config',
-            config=config_report)
+        cli.add_subcommand(cmd='report',
+                           help='Generate detailed reports for a single config',
+                           config=config_report)
```
kthui, line 180: This is updated by the styling script.

model_analyzer/triton/server/server_local.py
```diff
@@ -62,13 +64,11 @@ def start(self):
         if self._server_path:
             # Create command list and run subprocess
             cmd = [self._server_path]
-            cmd += self._server_config.to_cli_string().replace('=',
-                                                               ' ').split()
+            cmd += self._server_config.to_cli_string().replace('=', ' ').split()
```
kthui, line 67: This is updated by the styling script.

model_analyzer/triton/server/server_local.py
```diff
@@ -62,13 +64,11 @@ def start(self):
...
+            # List GPUs to be used by tritonserver
             triton_env = os.environ.copy()
-            if len(self._gpus) >= 1 and self._gpus[0] != 'all':
-                visible_gpus = GPUDeviceFactory.get_cuda_visible_gpus()
-                triton_env['CUDA_VISIBLE_DEVICES'] = ','.join(
-                    [visible_gpus[uuid] for uuid in self._gpus])
+            triton_env['CUDA_VISIBLE_DEVICES'] = ','.join(
```
Tabrizian, line 70: Does CUDA_VISIBLE_DEVICES work with GPU UUIDs?

Tabrizian:
Can you add a test for this in https://github.com/triton-inference-server/model_analyzer/blob/main/qa/L0_server_launch_modes/test.sh?

tests/mocks/mock_server_docker.py
```diff
@@ -30,15 +30,15 @@ class MockServerDockerMethods(MockServerMethods):
...
-                }
-            })
+                Mock(return_value={
+                    'memory_stats': {
+                        'usage': 0.0,
+                        'max_usage': 0.0,
+                        'limits': 0.0
+                    }
+                })
         }
```
kthui, line 42: This is updated by the styling script.

tests/mocks/mock_server_local.py
```diff
@@ -34,8 +35,7 @@ def __init__(self):
             'virtual_memory': Mock(return_value=Mock(**virtual_memory_attrs))
         }
         Popen_attrs = {
-            'communicate.return_value':
-            ('Triton Server Test Log', 'Test Error')
+            'communicate.return_value': ('Triton Server Test Log', 'Test Error')
         }
```
kthui, line 39: This is updated by the styling script.

tests/test_triton_server.py
```diff
@@ -100,23 +106,21 @@ def test_server_config(self):
 
             # Make sure parsed value is the one from dict, check type too
             test_value = CLI_TO_STRING_TEST_ARGS[arg]
-            self.assertEqual(
-                test_value,
-                type(test_value)(value),
-                msg=f"CLI string contained unknown value: {value}")
+            self.assertEqual(test_value,
+                             type(test_value)(value),
+                             msg=f"CLI string contained unknown value: {value}")
```
kthui, line 111: This is updated by the styling script.

model_analyzer/triton/server/server_local.py
kthui: It should work, because the previous code will set the `CUDA_VISIBLE_DEVICES` if the user provides UUIDs of GPU using the `gpus` config (case 1).
There are 3 cases of `gpus` config: 
   Case 1: User provides `["UUID(s) of GPU"]`
      This update does not modify the flow.
   Case 2: User provides `[]`
      This update will set the `CUDA_VISIBLE_DEVICES` to empty, while the previous code will not set it which causes a discrepancy between the GPU used by model analyzer and triton server.
   Case 3: User provides `all`
      This update will set the `CUDA_VISIBLE_DEVICES` to includes all GPU UUIDs visible by CUDA, while the previous code will not set it. This should not change the behavior.

kthui:
> Can you add a test for this in https://github.com/triton-inference-server/model_analyzer/blob/main/qa/L0_server_launch_modes/test.sh?

It is added

tests/test_triton_server.py
```diff
@@ -100,23 +106,21 @@ def test_server_config(self):
...
 
-    def test_create_server(self):
+    def test_create_server(self, gpus=["all"]):
```
Tabrizian, line 113: Same as above

tests/test_triton_server.py
```diff
@@ -132,13 +136,25 @@ def test_create_server(self):
...
+
+    def test_create_server_no_gpu(self):
+        self.test_create_server(gpus=[])
+
+    def test_create_server_select_gpu(self):
+        gpu_uuids = get_gpus_uuids(["all"])
+        if len(gpu_uuids):
+            self.test_create_server(gpus=gpu_uuids[:1])
+
+    def test_start_stop_gpus(self, gpus=["all"]):
```
Tabrizian, line 149: Same

tests/test_triton_server.py
```diff
@@ -189,20 +217,25 @@ def test_get_logs(self):
...
         self.server = TritonServerFactory.create_server_local(
-            path=TRITON_LOCAL_BIN_PATH, config=server_config, gpus=['all'])
+            path=TRITON_LOCAL_BIN_PATH, config=server_config, gpus=gpus)
         self.server.start()
         self.server.stop()
         self.server_local_mock.assert_server_process_terminate_called()
         self.assertEqual(self.server.logs(), "Triton Server Test Log")
 
-    def test_cpu_stats(self):
+    def test_cpu_stats(self, gpus=["all"]):
```
Tabrizian, line 226: Same as above

tests/test_triton_server.py
```diff
@@ -39,7 +40,12 @@
 }
 
 
+def get_gpus_uuids(gpus_config):
+    return GPUDeviceFactory.verify_requested_gpus(gpus_config)
```
Tabrizian, line 44: I think you need to mock the `GPUDeviceFactory`.

Tabrizian (changes requested):

model_analyzer/triton/server/server_local.py
```diff
@@ -40,15 +41,16 @@ def __init__(self, path, config, gpus, log_path):
...
+            Use ["all"] to include all GPUs
         log_path: str
             Absolute path to the triton log file
         """
 
         self._tritonserver_process = None
         self._server_config = config
         self._server_path = path
-        self._gpus = gpus
+        self._gpus = GPUDeviceFactory.verify_requested_gpus(gpus)
```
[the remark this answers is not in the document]
kthui: It is moved.

tests/mocks/mock_server_docker.py
```diff
@@ -88,6 +88,7 @@ def assert_server_process_start_called_with(self,
                                                 cmd,
                                                 model_repository_path,
                                                 triton_image,
+                                                device_requests=[0],
```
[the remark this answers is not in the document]
kthui: It is updated.

tests/test_triton_server.py
kthui: It is updated.

tests/test_triton_server.py
kthui: It is updated.

tests/test_triton_server.py
kthui: It is updated.

tests/test_triton_server.py
kthui: Good suggestion. This speeds up the test. It is updated.

qa/L0_server_launch_modes/test.sh
```diff
@@ -156,6 +156,9 @@ run_server_launch_modes "$CURRENT_GPUS"
 CURRENT_GPUS=${GPUS[@]:1}
 run_server_launch_modes "$CURRENT_GPUS"
 
+CURRENT_GPUS=""
+run_server_launch_modes "$CURRENT_GPUS"
```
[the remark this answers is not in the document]
kthui: I think this should trigger. 
It is because the `gpus` array is passed to the `run_server_launch_modes` function, and then converted to the `--gpus` flag for the model analyzer by the `convert_gpu_array_to_flag` function. It will result in the `run_server_launch_modes` function launching the model analyzer with an empty `--gpus` flag for different launch modes.

kthui:
Fix an issue where remote mode could check `len(None)`.

qa/L0_server_launch_modes/test.sh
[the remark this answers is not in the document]
Tabrizian: I don't think it does this. https://github.com/triton-inference-server/model_analyzer/blob/main/qa/L0_server_launch_modes/test.sh#L60 If the "CURRENT_GPUS" is not specified it will remove the `--gpus` flag altogether.

tests/test_triton_server.py
```diff
@@ -100,23 +104,24 @@ def test_server_config(self):
...
             test_value = CLI_TO_STRING_TEST_ARGS[arg]
-            self.assertEqual(
-                test_value,
-                type(test_value)(value),
-                msg=f"CLI string contained unknown value: {value}")
+            self.assertEqual(test_value,
+                             type(test_value)(value),
+                             msg=f"CLI string contained unknown value: {value}")
+
+    def test_create_server(self, gpus=None):
```
Tabrizian, line 111: Perhaps you can rename this function to `_test_create_server` so that it is not called by the `unittest` and then you can add another variant that says `test_create_server_all_gpu` and explicitly call it with `gpus=['all']`

tests/test_triton_server.py
```diff
@@ -100,23 +104,24 @@ def test_server_config(self):
...
+        if gpus == None:
+            gpus = ["all"]  # Avoid strange default args behavior
```
Tabrizian, line 113: If the comment above is applied you can remote this function as well.

tests/test_triton_server.py
```diff
@@ -132,13 +137,28 @@ def test_create_server(self):
...
+                path=TRITON_LOCAL_BIN_PATH, config=server_config, gpus=gpus)
+
+    def test_create_server_no_gpu(self):
+        self.test_create_server(gpus=[])
+
+    def test_create_server_select_gpu(self):
+        self.test_create_server(
+            gpus=self.gpu_device_factory_mock.get_cuda_visible_gpus()[:1])
+
+    def test_start_stop_gpus(self, gpus=None):
```
Tabrizian, line 149: Same

tests/mocks/mock_gpu_device_factory.py
```diff
@@ -0,0 +1,35 @@
...
+        super().__init__()
+
+    def _fill_patchers(self):
+        self._patchers.append(patch("model_analyzer.device.gpu_device_factory.GPUDeviceFactory.get_cuda_visible_gpus", 
+                                    MagicMock(return_value=dict.fromkeys(self._cuda_visible_gpus))))
+    
+    def get_cuda_visible_gpus(self):
+        return self._cuda_visible_gpus[:]
+    
+    def verify_requested_gpus(self, requested_gpus):
```
Tabrizian, line 32: You can mock this function as well similar to what you did for "model_analyzer.device.gpu_device_factory.GPUDeviceFactory.get_cuda_visible_gpus". If you mock this properly, you can just call `model_analyzer.device.gpu_device_factory.GPUDeviceFactory.verify_requested_gpus` and expect it to run your code.

tests/test_triton_server.py
Tabrizian: Also, I think the function name can be updated as well.

tests/test_triton_server.py
```diff
@@ -189,20 +224,29 @@ def test_get_logs(self):
...
 
-    def test_cpu_stats(self):
+    def test_cpu_stats(self, gpus=None):
+        if gpus == None:
+            gpus = ["all"]  # Avoid strange default args behavior
+
+        # Device request for docker mode
+        device_requests = [0]
+        if len(gpus) == 0 or len(
+                self.gpu_device_factory_mock.get_cuda_visible_gpus()) == 0:
```
Tabrizian, line 240: You should just call the `GPUDeviceFactory.get_cuda_visible_gpus()` and expect it to run your mock. For example, in this test, we are not calling `TritonServerFactory` from the mock. We just call the original version and the mock is executed.

Tabrizian (changes requested):

Tabrizian:
Also, it would be a good idea to not force-push after the PR is reviewed once. You can just add new commits to your PR containing the review edits. This way only the new changes can be easily reviewed.

model_analyzer/triton/server/server_local.py
```diff
@@ -62,13 +62,11 @@ def start(self):
...
             # Create command list and run subprocess
             cmd = [self._server_path]
-            cmd += self._server_config.to_cli_string().replace('=',
-                                                               ' ').split()
+            cmd += self._server_config.to_cli_string().replace('=', ' ').split()
+            # List GPUs to be used by tritonserver
             triton_env = os.environ.copy()
-            if len(self._gpus) >= 1 and self._gpus[0] != 'all':
-                visible_gpus = GPUDeviceFactory.get_cuda_visible_gpus()
-                triton_env['CUDA_VISIBLE_DEVICES'] = ','.join(
```
aramesh7, line 70: DId you find this to be redundant?

model_analyzer/triton/server/server_docker.py
```diff
@@ -62,11 +62,9 @@ def start(self):
...
         """
 
-        if len(self._gpus) == 1 and self._gpus[0] == 'all':
-            devices = [
-                docker.types.DeviceRequest(count=-1, capabilities=[['gpu']])
-            ]
-        else:
+        # List GPUs to be mounted inside docker container
+        devices = []
+        if len(self._gpus):
```
aramesh7, line 67: Is this case now taken care of after calling verify_requested_gpus?

model_analyzer/triton/server/server_local.py
kthui: You mean if it includes all GPUs, then just don't bother setting the flag?

model_analyzer/triton/server/server_docker.py
kthui: Yes. The `verify_requested_gpus` will be called by the `server_factory.py` before calling into `server_docker.py` and `server_local.py`.

tests/test_triton_server.py
kthui: Updated.

tests/test_triton_server.py
kthui: Updated.

tests/test_triton_server.py
kthui: Updated.

tests/mocks/mock_gpu_device_factory.py
kthui: Actually, what I want to do is to fake 2 GPUs using the mock. The `verify_requested_gpus` can be tested using this unit test, so I won't be mocking that function. I removed the `get_cuda_visible_gpus` and `verify_requested_gpus` functions, and just use the expected answer for each test.

tests/test_triton_server.py
kthui: I'm a bit skeptical on using `GPUDeviceFactory` directly. Calling it directly certainly works, but the goal of getting the `device_requests` variable is to test the `server_docker.py` actually sets the `device_requests` correctly, so it is a bit like testing itself with itself.
I refactored setting visible GPUs to `setUp`, so the `mock_gpu_device_factory.py` no longer decides visible GPUs, which will be decided by the `test_triton_server.py`. Then, this should no longer be an issue.

qa/L0_server_launch_modes/test.sh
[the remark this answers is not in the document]
kthui: Oh no. This is fixed, with a special case for empty flag

Tabrizian (dismissed):

Tabrizian:
Make sure it passes the CI before merging.

model_analyzer/triton/server/server_local.py
aramesh7: Im just asking why it was removed.

aramesh7 (dismissed):

model_analyzer/triton/server/server_local.py
kthui: It is moved to `server_factory.py`.

kthui:
There was an issue with the docker mode where setting the argument `device_requests = []` in the `self._docker_client.containers.run()` might still allow the docker to mount the GPUs.
This is happening during CI, but not during local test.
The fix is to set the `CUDA_VISIBLE_DEVICES` environment variable inside the docker container in case the GPUs are mounted.

Tabrizian (approved):

PR #189, opened 2021-08-03
Add options for cpu metric collection

kthui:


model_analyzer/config/input/config_command_profile.py
```diff
@@ -178,6 +178,13 @@ def _fill_config(self):
                 default_value=DEFAULT_DURATION_SECONDS,
                 description=
                 'Specifies how long (seconds) to gather server-only metrics'))
+        self._add_config(
```
aramesh7, line 181: I think that this option exposes too much of the internals of model analyzer. Ideally, we do not want the user to know about the CPU monitor. We want them to only deal in the level of CPU metrics. What if we moved the metric tags from the MetricsManager and made it user configurable with a reasonable default set of metrics (no CPU metrics). Then in the metrics manager, we use that option to determine whether we need to run the CPU monitor or not. @Tabrizian would like your input on this as well.

aramesh7 (changes requested):

model_analyzer/config/input/config_command_profile.py
Tabrizian: Agreed. I think using the metric tags to determine whether the CPU monitor should be enabled or not is a good idea. Having this option can complicate things. For example, if the CPU monitor is disabled and cpu metrics are in the metrics tag what should the MA do?

kthui: I see, so ultimately the CPU monitor follows the metric tags, but the auto mode logic should be used to display warning(s) if there is a discrepancy between what is specified and what is supposed to do.

kthui: @aramesh7 @Tabrizian Please confirm if this setup is what you are expecting. 
I will be working on the analyze part for this change.

docs/config.md
```diff
@@ -86,9 +86,8 @@ profile_models: <comma-delimited-string-list>
 # Duration of waiting time between each metric measurement in seconds
 [ monitoring_interval: <float> | default: 1 ]
 
-# Specifies whether to disable cpu monitor or not.
-# "auto" will disable cpu monitor when GPU(s) are used for profiling.
-[ cpu_monitor_disable: ("auto"|True|False) | default: "auto" ]
+# Specifies which metric(s) are to be collected.
+[ metrics: comma-delimited-list-string | default: `all non CPU metrics` ]
```
aramesh7, line 90: Set the default here to say `default: See Config Defaults section`. Look at the argument `server_output_fields` for example.

model_analyzer/config/input/config_command_profile.py
```diff
@@ -594,6 +595,10 @@ def _autofill_values(self):
         config values.
         """
 
+        min_metric_tags = DEFAULT_METRICS
+        extra_metric_tags = list(set(self.metrics) - set(min_metric_tags))
+        self._fields["metrics"].set_value(min_metric_tags + extra_metric_tags)
```
aramesh7, line 600: This is good for now. Ideally we want flexibility on what the user is capturing, but the model analyzer requires there be some metrics like throughput. This makes things tricky because in analysis we have to make sure that the user does not try to prioritize the results by a metric that was not captured. 

One solution I can think of for this is to store the captured metric list to the checkpoint as profile metadata. This must somehow then be used to check the objectives and constraints sections of the ConfigCommandAnalyze.

To make this easier for 21.08, maybe we don't give the user full control over the metrics to be collected, but instead going back to what you did, add an argument `--collect-cpu-metrics`.  The logic you have in this PR can remain the same. This flag will be false by default, but if true or there is a model that is cpu_only, `cpu_used_ram` and `cpu_available_ram` will be added to the metric tags internally.

model_analyzer/record/metrics_manager.py
```diff
@@ -164,44 +150,46 @@ def profile_model(self, run_config, perf_output_writer=None):
             The gpu specific and non gpu metrics
         """
 
-        collect_gpu_metrics = not run_config.model_config().cpu_only()
-        collect_cpu_metrics = not self._config.cpu_monitor_disable
```
aramesh7: You may have this be set to the value of the argument `--collect-cpu-metrics`

aramesh7 (changes requested):

docs/config.md
kthui: Updated.

model_analyzer/record/metrics_manager.py
kthui: The logic here is changed. The `--metric` flag will be inspected for whether the CPU monitor will run or not.

model_analyzer/config/input/config_command_profile.py
kthui: I agree this is good for now, for 21.08.

The user actually does not have the ability to control flags other than `cpu_used_ram` and `cpu_available_ram`. It is because the default metrics - the same as previously collected metrics except `cpu_used_ram` and `cpu_available_ram` - will always be collected regardless of what the user specifies. If the user adds `cpu_used_ram` and/or `cpu_available_ram`, then CPU monitor will run and collect the requested CPU metrics.

I think this is good placeholder for future changes. Currently, if a model is `cpu_only` or no GPU is visible, then a `logging.info` will be printed to remind the user when none of the CPU metric flag is set. (Instead of automatically adding the flag.)

For the `ConfigCommandAnalyze`, instead of capturing the metric list, the metric list is determined by checking the collected measurements. This removes a possible corner case where the user runs `profile` twice with a different `--metrics` flag.

kthui:
I will add a unit test for the running of CPU monitor.

model_analyzer/result/result_manager.py
```diff
@@ -160,9 +170,50 @@ def create_tables(self, gpu_specific_metrics, non_gpu_specific_metrics):
             The metrics that do not have a GPU id associated with them
         """
 
+        gpu_specific_metrics_from_measurements = {}
```
aramesh7, line 173: Would like to see a qa test for this. Also please add a couple of lines of comments explaining why it is necessary for us to get the metrics from the checkpoint here.

aramesh7 (changes requested):

model_analyzer/result/result_manager.py
kthui: Explanation is added.

kthui: Could the `L0_export_metrics` test this case? It runs the `analyze` command on a pre-profiled checkpoint file, which includes this logic during the run. If the output metrics are correct, then this logic should have captured the metric columns correctly.

model_analyzer/record/metrics_manager.py
```diff
@@ -160,6 +153,23 @@ def profile_model(self, run_config, perf_output_writer=None):
...
+                "CPU metric(s) are not being collected, while this profiling will run on CPU(s)."
+            )
+        # Warn user about CPU monitor performance issue
+        if collect_cpu_metrics_actual:
+            logging.warning("CPU metric(s) are being collected.")
+            logging.warning(
+                "Collecting CPU metric(s) during profiling might affect performance!"
+            )
+            logging.info(
+                "CPU metric(s) collection can be disabled by changing the `--metrics` flag."
```
Tabrizian, line 170: "CPU metric(s) collection can be disabled by removing the CPU metrics (e.g. cpu_used_ram) from the `--metrics` flag."

model_analyzer/reports/report_manager.py
```diff
@@ -461,12 +458,11 @@ def _build_summary_table(self,
...
                     model_config.dynamic_batching_string(), instance_group_str,
-                    measurement.get_metric('perf_latency').value(),
-                    measurement.get_metric('perf_throughput').value(),
-                    measurement.get_metric('cpu_used_ram').value(),
-                    measurement.get_metric('gpu_used_memory').value(),
-                    round(
-                        measurement.get_metric('gpu_utilization').value(), 1)
+                    measurement.get_metric_value('perf_latency'),
+                    measurement.get_metric_value('perf_throughput'),
+                    measurement.get_metric_value('cpu_used_ram'),
```
Tabrizian, line 463: If there is no `cpu_used_ram` will the report still work fine?

model_analyzer/result/measurement.py
```diff
@@ -160,6 +157,28 @@ def get_metric(self, tag):
                             "measurements across devices.")
             return None
 
+    def get_metric_value(self, tag, not_found_value=0):
```
Tabrizian, line 160: `not_found_value` -> `default_value`.

model_analyzer/result/result_manager.py
```diff
@@ -83,23 +85,25 @@ def _init_state(self):
...
                 server_output_headers.append(self.headers[server_output_field])
             elif server_output_field in self._gpu_metrics_to_headers:
                 server_output_headers.append(
                     self._gpu_metrics_to_headers[server_output_field])
             else:
-                raise TritonModelAnalyzerException(
-                    f'Server output field "{server_output_field}", does not exist'
-                )
+                logging.warning(
+                    f'Server output field "{server_output_field}", has no data')
```
Tabrizian, line 103: Is it because the CPU metrics are disabled? If I understood correctly, it is going to be displayed on every run run of MA using default arguments, right?

model_analyzer/result/result_manager.py
```diff
@@ -108,20 +112,22 @@ def _create_server_table(self):
...
             if inference_output_field in self.headers:
                 inference_output_headers.append(
                     self.headers[inference_output_field])
             elif inference_output_field in self._non_gpu_metrics_to_headers:
                 inference_output_headers.append(
                     self._non_gpu_metrics_to_headers[inference_output_field])
             else:
-                raise TritonModelAnalyzerException(
-                    f'Inference output field "{inference_output_field}", does not exist'
+                logging.warning(
```
Tabrizian, line 124: Can we modify the `server_output_fields` and `gpu_output_fields` based on the `metrics` config variable so that we don't have to ignore the headers in case they don't exist?

Tabrizian (changes requested):

docs/config.md
```diff
@@ -86,6 +86,9 @@ profile_models: <comma-delimited-string-list>
 # Duration of waiting time between each metric measurement in seconds
 [ monitoring_interval: <float> | default: 1 ]
 
+# Specifies which metric(s) are to be collected.
+[ metrics: <comma-delimited-string-list> | default: See Config Defaults section ]
```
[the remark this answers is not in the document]
kthui: Good idea. Added.

model_analyzer/record/metrics_manager.py
```diff
@@ -28,22 +29,15 @@
 import numba
 import requests
```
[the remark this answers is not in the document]
kthui: Updated.

model_analyzer/record/metrics_manager.py
```diff
@@ -160,6 +153,23 @@ def profile_model(self, run_config, perf_output_writer=None):
         cpu_only = run_config.model_config().cpu_only()
         perf_config = run_config.perf_config()
 
+        # Inform user CPU metric(s) are not being collected under CPU mode
+        collect_cpu_metrics_expect = cpu_only or len(self._gpus) == 0
```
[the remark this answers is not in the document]
kthui: This variable is to detect whether the user has set the `--metrics` flag to collect CPU metric(s) "correctly" under CPU mode. It is basically to inform the user that the model is running under CPU mode and CPU metric(s) are not being collected.

model_analyzer/record/metrics_manager.py
```diff
@@ -160,6 +153,23 @@ def profile_model(self, run_config, perf_output_writer=None):
...
+        collect_cpu_metrics_actual = len(self._cpu_metrics) > 0
+        if collect_cpu_metrics_expect and not collect_cpu_metrics_actual:
+            logging.info(
+                "CPU metric(s) are not being collected, while this profiling will run on CPU(s)."
+            )
+        # Warn user about CPU monitor performance issue
+        if collect_cpu_metrics_actual:
+            logging.warning("CPU metric(s) are being collected.")
+            logging.warning(
+                "Collecting CPU metric(s) during profiling might affect performance!"
```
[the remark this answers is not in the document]
kthui: Updated.

model_analyzer/record/metrics_manager.py
kthui: Updated.

model_analyzer/reports/report_manager.py
kthui: Yes. It will just return `0` by default. The behavior is similar to running the CPU monitor under remote mode.

model_analyzer/result/measurement.py
kthui: Updated.

model_analyzer/result/result_manager.py
kthui: Yes. The `profile` command does not include CPU metrics, but the `analyze` command includes CPU metrics by default. This will cause it to print.

Tabrizian: Is there any way to prevent this? I think using this warning we are pushing the user toward two different contradictory directions. If they have CPU metrics inside the `--metrics` flag, we are going to display a warning that it affects the performance. If they do not use CPU metrics they are going to receive the warning above when running analyze. I think the MA run using the default configurations should be warning free (unless something has gone wrong).

model_analyzer/result/result_manager.py
kthui: This is tricky because the user can specify different metrics during `profile` and `analyze`. 

If we ask the user to input the `--metrics` flag during `analyze`, the flag is redundant with `server_output_fields`, `inference_output_fields` and `gpu_output_fields`. If we capture the `--metrics` flag during `profile`, it is not necessary since we are finding out exactly which metrics were collected.

The warning is due to a discrepancy between metrics during `profile` and `analyze`. I think it is a good idea we keep the default metrics during `analyze` phase a superset of `profile` phase, to avoid cases like the user includes `cpu_used_ram` during `profile` but forget to includes it during `analyze`.

Can we print the opposite of the warning? We inform the user there are metrics profiled but not analyzed, instead of there are not profiled metrics being analyzed.

model_analyzer/result/result_manager.py
aramesh7: Yes we don't want warnings on default runs. If we are changing the default profile to exclude CPU metrics, the default analyze will also have to exclude them from all the tables and reports.

model_analyzer/result/result_manager.py
aramesh7: I think its best if analyze is able to automatically detect what metrics were profiled. Unfortunately this would likely make it necessary for analyze to find out the largest superset of metrics that were collected during profile. This is because it is possible to run a profile with metrics X,Y,Z then run another profile with metrics X,Y or Y,Z, or W,X,Y. When analyze is asked to export the data, it must somehow unify these.

aramesh7: If the user is requesting metrics be in the output tables that were not collect, we must warn them and exclude those from the output.

model_analyzer/reports/report_manager.py
aramesh7: There is currently very few tests that check the reports. It would be a good idea to generate two examples and link them in this PR so that we can look at them.

qa/L0_config/config_generator.py
```diff
@@ -20,6 +20,7 @@ def _get_sweep_configs(profile_models):
 
     sweep_configs = []
     model_config = {
+        'metrics': ['cpu_available_ram', 'cpu_used_ram'],
```
aramesh7, line 23: So in the actual config, this list of metrics works out to be just the "extra metrics" that are being collected. I don't know if there is a good case to have this be called `metrics` anymore.

model_analyzer/config/input/config_command_profile.py
```diff
@@ -178,6 +178,14 @@ def _fill_config(self):
                 default_value=DEFAULT_DURATION_SECONDS,
                 description=
                 'Specifies how long (seconds) to gather server-only metrics'))
+        self._add_config(
+            ConfigField(
+                'metrics',
```
aramesh7, line 183: Would this end up cleaner if we made this a boolean flag that only determines whether we collect cpu metrics or not? Much of the logic here would stay the same, except the challenge would only be removing CPU metrics from the reports and tables etc unless this value is true.

aramesh7 (changes requested):

qa/L0_config/config_generator.py
kthui: I could use the full list, but since the minimum list is added on to the provided list, I just keep it short.
Maybe we can call it `optional_metrics`?

model_analyzer/config/input/config_command_profile.py
kthui: If we are going back to the on/off mode, should we also bring back the `auto` mode?
`True`: Collect CPU metrics
`False`: Do not collect CPU metrics
`auto`: Collect CPU metrics iff the model will run on CPU. (default to)
Then, typical users would not need to deal with this flag, except they want or do not want CPU metrics.

model_analyzer/result/result_manager.py
kthui: Maybe we can just print an `info` instead of `warning`?

model_analyzer/result/result_manager.py
kthui: > I think its best if analyze is able to automatically detect what metrics were profiled. Unfortunately this would likely make it necessary for analyze to find out the largest superset of metrics that were collected during profile. This is because it is possible to run a profile with metrics X,Y,Z then run another profile with metrics X,Y or Y,Z, or W,X,Y. When analyze is asked to export the data, it must somehow unify these.

This is achieved by line 173-205.

kthui: > If the user is requesting metrics be in the output tables that were not collected, we must warn them and exclude those from the output.

I will not update the behavior, but change the default. Then, the user should remember to update metrics in the `analyze` part if the `profile` part includes CPU metrics.

model_analyzer/config/input/config_command_profile.py
aramesh7: @Tabrizian @kthui what are your thoughts on this. If we don't want to go back to a boolean flag `--collect-cpu-metrics`, then the rest of this PR looks good to me.

kthui: A summary of the changes so far:
  - `--metrics` flag of the `profile` stage specifies what metrics are to be collected.
  - A minimum set of metrics is enforced similar to `inference_output_fields`, ... of the `analyze` stage.
  - The 2 CPU metrics are currently optional.
  - `analyze` stage will find the union of metrics collected during `profile`, instead of using a fixed list of metrics.
  - By default, CPU metrics are not included in both `profile` and `analyze` stages.
  - A unit test is added to check the CPU monitor will not run if non of the CPU metrics are requested during `profile` stage.
  - L0_metrics test is added to check requesting different metrics works.

I think this change is more extensible than a boolean flag, in the future.

Tabrizian: Looks good to me as well. If the default run of MA is warning free, then the changes look good to me.

docs/config.md
```diff
@@ -215,13 +218,13 @@ analysis_models: <comma-delimited-string-list>
 [ filename_server_only: <string> | default: metrics-server-only.csv ]
 
 # Specifies columns keys for model inference metrics table
-[ inference_output_fields: <comma-delimited-string-list> | default: See Config Defaults section]
+[ inference_output_fields: <comma-delimited-string-list> | default: See [Config Defaults](#config-defaults) section]
```
Tabrizian, line 221: I think it would be better if we don't exposes these parameters inside the documentation. Having `--metrics` flag, appropriately set these parameters is a better option.

Tabrizian (changes requested):

model_analyzer/config/input/config_command_profile.py
kthui: CPU metrics are not included by default during `analyze` stage (and they are not collected by default), so it is error free.

docs/config.md
kthui: The `--metrics` flag is replaced with `--collect-cpu-metrics` flag in flavor for simplicity, so less parameters are exposed.

aramesh7 (approved):

Tabrizian (approved):

docs/config.md
Tabrizian: `--collect-cpu-metrics` sounds good as well.

PR #192, opened 2021-08-06
Quiet table output with global -q option

aramesh7:


model_analyzer/plots/detailed_plot.py
```diff
@@ -14,10 +14,13 @@
 
 import os
 import matplotlib.pyplot as plt
+import logging
 from matplotlib import patches as mpatches
 from collections import defaultdict
 from model_analyzer.record.metrics_manager import MetricsManager
 
+logging.getLogger('matplotlib').setLevel(logging.ERROR)
```
Tabrizian, line 22: Why do you need to set matplotlib logging to error?

aramesh7: Theres a lot of info level logging that matplotlib writes when we run `model-analyzer report`. This is to silence that.

Tabrizian (dismissed):

Tabrizian (approved):

PR #193, opened 2021-08-10
Allow user to set triton container environment and volume mounts

aramesh7:


docs/config.md
```diff
@@ -173,7 +177,11 @@ profile_models: <comma-delimited-string-list|list|profile_model>
 [ triton_server_flags: <dict> ]
 
 # Allows custom configuration of perf analyzer instances used by model analyzer
-[ perf_analyzer_flags: <dict>]
+[ perf_analyzer_flags: <dict> ]
+
+# Allows custom configuration of the environment variables for tritonserver instances
+# launched by model analyzer
+[ triton_server_environment: <dict> ]
```
Tabrizian, line 184: "triton_server_environment_variables"

docs/config.md
```diff
@@ -680,6 +688,56 @@ profile_models:
   arguments in this section. An example of this is `http-port`, which is an
   argument to Model Analyzer itself.
 
+### `<triton-server-environment>`
+
+This section enables setting environment variables for the tritonserver
+instances launched by model-analyzer. This is useful, for examples, when
```
Tabrizian, line 694: model-analyzer -> Model Analyzer
"For example, when a ..."

docs/config.md
```diff
@@ -680,6 +688,56 @@ profile_models:
...
+a custom operation is required by a model and Triton requires the LD_PRELOAD and
```
Tabrizian, line 695: Perhaps link to the PyTorch custom ops page?

docs/config.md
```diff
@@ -680,6 +688,56 @@ profile_models:
...
+LD_LIBRARY_PATH environment variables to be set. The section is a dictionary where the
```
Tabrizian, line 696: "The section is a dictionary where the" -> "The value for this section is a dictionary"..

docs/config.md
```diff
@@ -680,6 +688,56 @@ profile_models:
...
+keys are the environment variable names and their values are the values to be
+set.
+
+#### Example
+
+```yaml
+model_repository: /path/to/model/repository/
+profile_models:
+  model_1:
+    parameters:
```
Tabrizian, line 706: I think having parameters section is not necessary here.

model_analyzer/config/input/config_command_profile.py
```diff
@@ -249,6 +253,13 @@ def _add_profile_models_configs(self):
                 description=
                 'Allows custom configuration of the triton instances used by model analyzer.'
             ))
+        self._add_config(
+            ConfigField(
+                'triton_server_environment',
+                field_type=triton_server_environment_scheme,
+                description=
+                'Allows setting environment variables for tritonserver server instances launched by model analyzer'
```
Tabrizian, line 261: "model analyzer" -> "Model Analyzer."

model_analyzer/triton/server/server_docker.py
```diff
@@ -69,16 +71,39 @@ def start(self):
...
+            env.update({
+                'CUDA_VISIBLE_DEVICES':
+                ','.join([uuid for uuid in self._gpus])
+            })
+
+            # Filter out variables that need to be set within container
+            # If the variable is LD_PRELOAD, it cannot be set in docker env
+            # If the variable contains a lookup, it must be done inside the container
+            for env_variable in list(env):
+                if env_variable == "LD_PRELOAD" or env[env_variable].find(
```
Tabrizian, line 88: What happens to the variables that do not satisfy this if condition?

model_analyzer/triton/server/server_docker.py
```diff
@@ -69,16 +71,39 @@ def start(self):
...
+        # Supports only strings, and value lookups/concats
+        env_cmds = []
+        if env:
+            env.update({
+                'CUDA_VISIBLE_DEVICES':
+                ','.join([uuid for uuid in self._gpus])
+            })
+
+            # Filter out variables that need to be set within container
+            # If the variable is LD_PRELOAD, it cannot be set in docker env
```
Tabrizian, line 85: Is `LD_PRELOAD` the only variable that needs to be handled in this way? Can we be flexible to support both modes for all the variables? For example, `X=$VALUE` indicates that it should be set from the outside container and `X=$$VALUE` indicates that it should be set from inside the container.

Tabrizian (changes requested):

model_analyzer/triton/server/server_docker.py
aramesh7: They are able to be passed into docker run's "environment" parameter.

model_analyzer/triton/server/server_docker.py
aramesh7: Yes, LD_PRELOAD seems to need to be set immediately before the program being run, and its best to set it only for that program. 

Regards to your suggestion, I am wondering which case is not covered in the code as is? If the variables value contains a lookup it will need to be set from inside the container. 

I suppose there are cases where the you may need to set the variable to a value containing "$" is that the case you are referring to?

model_analyzer/triton/server/server_docker.py
```diff
@@ -32,8 +32,11 @@ class TritonServerDocker(TritonServer):
     Concrete Implementation of TritonServer interface that runs
     triton in a docker container.
     """
+<<<<<<< HEAD
```
Tabrizian, line 35: Merge/rebase issues

model_analyzer/triton/server/server_docker.py
Tabrizian: Given that the name of the option is "triton_server_environment" why not set all the parameters like LD_PRELOAD?

> I suppose there are cases where the you may need to set the variable to a value containing "$" is that the case you are referring to?

Right, I was wondering which environment will be used for the resolution of that env variable? Is it going to use the environment variables inside the container or it is going to use the outside environment to resolve that variable? For example, if I set "X=$PATH", is going to be resolved to the PATH variable inside the container or outside the container.

docs/config.md
aramesh7: My logic here was just that wherever we see being able to pass environment variables, the "variables" is always left out. For example docker environment, Popen environment etc. Unless you feel strongly that it should be more specific, I'd rather keep it shorter.

model_analyzer/triton/server/server_docker.py
aramesh7: That will be resolved inside the container. Anything with a $ will be resolved inside. The reason i opted to split some up was in case the command needs to be debugged, i didn't want the tritonserver command to be inundated with environment variables. If you think this is unnecessary I can set them all the same way.

docs/config.md
Tabrizian: That makes sense. Let's keep it this way.

model_analyzer/triton/server/server_docker.py
Tabrizian: I think it is a bit strange to single out `LD_PRELOAD` as a special environment variable. Can you elaborate on what you mean by the command needs to be debugged?

aramesh7: The tritonserver command that ultimately gets run inside the container will be somethign like `X=1 Y=2 tritonserver ....` If we have a lot of environment variables, i think this will get messy. This was my only concern. Otherwise i think you're probably right that we're better off setting all the variables in the command.

aramesh7: I will make this change and rerun the CI.

Tabrizian (dismissed):

Tabrizian (approved):

PR #196, opened 2021-08-11
Added L0_stability_perf

aramesh7:
* Added perf stability test to compare model analyzer against perf analyzer standalone
* Fixed bug where early exit from model analyzer might leave tritonserver running.

### DIFF
--- a/qa/L0_stability_perf/check_results.py
+++ b/qa/L0_stability_perf/check_results.py
@@ -1,0 +1,123 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import sys
+import glob
+import os
+from collections import defaultdict
+import re
+
+
+class TestOutputValidator:
+    """
+    Functions that validate the output
+    of the test
+    """
+    def __init__(self, test_name, threshold):
+        self._tolerance_percent = threshold
+
+        check_function = self.__getattribute__(f'check_{test_name}')
+
+        if check_function():
+            sys.exit(0)
+        else:
+            sys.exit(1)
+
+    def check_perf_stability(self):
+        """
+        Get the perf results from each model analyzer run
+        as well as perf_analyzer run and compare them
+        """
+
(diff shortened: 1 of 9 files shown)


qa/L0_stability_perf/check_results.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def check_perf_stability(self):
+        """
+        Get the perf results from each model analyzer run
+        as well as perf_analyzer run and compare them
+        """
+
+        log_paths = glob.glob(os.path.join(os.getcwd(), f"*.log"))
+
+        # Split the logs up into log pairs model_name: [model_name.test.log, model_name.perf.log]
```
Tabrizian, line 46: Could you please explain what's the difference between `model_name.test.log` and `model_name.perf.log`?

qa/L0_stability_perf/check_results.py
```diff
@@ -0,0 +1,123 @@
...
+                    latencies.append(float(p99_latency.group(1)))
+                else:
+                    print(f"\n***\n*** No throughput found in {log} \n***")
+                    return False
+
+            # Once all throughputs and latencies are collected, compute diff percentage
+            throughput_diff = 100 * (throughputs[0] -
+                                     throughputs[1]) / throughputs[0]
+
+            latency_diff = 100 * (latencies[1] - latencies[0]) / latencies[1]
```
Tabrizian, line 81: Why throughput diff is calculated different than latency diff?

qa/L0_stability_perf/check_results.py
```diff
@@ -0,0 +1,123 @@
...
+
+            if throughput_diff > self._tolerance_percent and latency_diff > self._tolerance_percent:
+                print(
+                    f"\n***\n*** Model Analyzer throughput and latency differ "
+                    f"greatly from perf analyzer standalone for model {model_name}. "
+                    f"\n*** perf analyzer values: Throughput={throughputs[0]} infer/sec, p99 Latency={latencies[0]} usec. \n***"
```
Tabrizian, line 87: Maybe include the MA results here as well?

Tabrizian (approved):

qa/L0_stability_perf/check_results.py
aramesh7: The logic here is that the model-analyzer value can be higher (better than) the perf_analyzer value, but should not be very much worse.

qa/L0_stability_perf/check_results.py
aramesh7: The MA results are appended on error

qa/L0_stability_perf/check_results.py
aramesh7: .test.log is the model-analyzer log, same as all the other tests. .perf.log is the perf analyzer log.

qa/L0_stability_perf/check_results.py
aramesh7: Better values for latency are lower, which is why the sign is flipped.

qa/L0_stability_perf/check_results.py
Tabrizian: Would be great to add a comment but I was just wondering which files they refer to.

PR #198, opened 2021-08-11
Handle missing metrics from perf_analyzer output gracefully

kthui:


model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -242,7 +242,7 @@ def get_records(self):
         last perf_analyzer run
         """
 
-        if self._perf_records:
+        if self._output:
```
aramesh7, line 245: How does this change the behaviour? Are we guaranteed that when there is output, there are records?

aramesh7 (changes requested):
What happens when a metric is requested in the output tables but is not available in the perf_analyzer output?

model_analyzer/perf_analyzer/perf_analyzer.py
kthui: There is no guarantee, but because of #189 change, the `analyze` part will gracefully handle no records output. If this is a concert, we can revert it back to checking `self._perf_records`.

kthui: Reverted.

aramesh7 (approved):

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020-2021 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Tabrizian, line 1: Remove "(c)" to avoid copyright wrap around.

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -276,16 +276,14 @@ def _parse_output(self, metrics):
             if metric not in self.perf_metrics:
                 raise TritonModelAnalyzerException(
                     f"Perf metric : {metric} not found or supported.")
+            parse_func = getattr(self, self.perf_metrics[metric])
+            output = None
```
Tabrizian, line 280: This line is not required.

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -302,8 +300,9 @@ def _parse_perf_client_send_recv(self, section):
         if client_send_recv:
             client_send_recv = float(client_send_recv.group(1)) / 1e3
             return PerfClientSendRecv(value=client_send_recv)
-        raise TritonModelAnalyzerException(
-            'perf_analyzer output did not client send/recv time.')
+        logging.error(
+            'perf_analyzer output did not contain client send/recv time.')
+        return None
```
Tabrizian, line 305: Why is this return required? I think `None` is the default return value.

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -302,8 +300,9 @@ def _parse_perf_client_send_recv(self, section):
         if client_send_recv:
             client_send_recv = float(client_send_recv.group(1)) / 1e3
             return PerfClientSendRecv(value=client_send_recv)
-        raise TritonModelAnalyzerException(
-            'perf_analyzer output did not client send/recv time.')
+        logging.error(
```
Tabrizian, line 303: Maybe use warning instead of error?

Tabrizian (changes requested):

model_analyzer/perf_analyzer/perf_analyzer.py
kthui: Updated.

model_analyzer/perf_analyzer/perf_analyzer.py
kthui: Updated.

model_analyzer/perf_analyzer/perf_analyzer.py
kthui: It is not required. Removed.

model_analyzer/perf_analyzer/perf_analyzer.py
kthui: Changed to warning. @aramesh7 what do you think?

aramesh7: Currently `-q` option on model analyzer silences everything below error, but I was wondering whether we should change that to allow warnings?

aramesh7 (commented):
Could you make the changes in the commit explicit? Doesn't need that much context but "Some small fixes" is too vague

Tabrizian (approved):

model_analyzer/perf_analyzer/perf_analyzer.py
kthui: Since this is not a critical error because the model analyzer can continue to function afterward, maybe it is okay not printing the warning if the user elects to silent it?

aramesh7: Its fine as a warning for now.

PR #200, opened 2021-08-17
Fixed CI issues with docker launch mode

aramesh7:


model_analyzer/config/input/config_command_profile.py
```diff
@@ -604,19 +612,20 @@ def _preprocess_and_verify_arguments(self):
...
-                        raise TritonModelAnalyzerException(
-                            "triton_docker_mounts needs to be a list of strings. Each string "
-                            " should be of the format <host path>:<container dest>:<access mode>"
-                        )
+                if self.triton_docker_mounts:
+                    for volume_str in self.triton_docker_mounts:
+                        if volume_str.count(':') != 2:
+                            raise TritonModelAnalyzerException(
+                                "triton_docker_mounts needs to be a list of strings. Each string "
+                                " should be of the format <host path>:<container dest>:<access mode>"
```
Tabrizian, line 627: Why require the access mode and not make it optional?

qa/L0_server_launch_modes/test.sh
```diff
@@ -64,70 +71,70 @@ function convert_gpu_array_to_flag() {
...
-                SERVER_HTTP_PORT=${http_port}
-                
-                run_server
-                if [ "$SERVER_PID" == "0" ]; then
-                    echo -e "\n***\n*** Failed to start $SERVER\n***"
-                    cat $SERVER_LOG
-                    exit 1
-                fi
+    for CONFIG_FILE in ${LIST_OF_CONFIG_FILES[@]}; do
+        CONFIG_PARAMETERS=${CONFIG_FILE%".yaml"}    # e.g. config-docker-http.yaml -> config-docker-http
```
Tabrizian, line 75: Move the comment to the top of the line.

qa/L0_server_launch_modes/test.sh
```diff
@@ -24,24 +24,31 @@ REPO_VERSION=${NVIDIA_TRITON_SERVER_VERSION}
 MODEL_REPOSITORY=${MODEL_REPOSITORY:="/mnt/dldata/inferenceserver/$REPO_VERSION/libtorch_model_store"}
 CHECKPOINT_REPOSITORY=${CHECKPOINT_REPOSITORY:="/mnt/dldata/inferenceserver/model_analyzer_checkpoints"}
 MODEL_NAMES="vgg19_libtorch"
-BATCH_SIZES="4"
-CONCURRENCY="4"
+TRITON_LAUNCH_MODES="local docker remote"
```
Tabrizian, line 27: Just wondering what were the issues with the Docker launch mode.

qa/L0_state_management/test.sh
```diff
@@ -32,7 +32,6 @@ CLIENT_PROTOCOL="http"
 PORTS=(`find_available_ports 3`)
 GPUS=(`get_all_gpus_uuids`)
 OUTPUT_MODEL_REPOSITORY=${OUTPUT_MODEL_REPOSITORY:=`get_output_directory`}
-rm -rf $OUTPUT_MODEL_REPOSITORY
```
Tabrizian, line 35: Why remove this line?

model_analyzer/triton/server/server_docker.py
```diff
@@ -180,7 +177,17 @@ def stop(self):
             self._tritonserver_container.stop()
             self._tritonserver_container.remove(force=True)
             self._tritonserver_container = None
-            self._docker_client.close()
+            logger.info('Stopped Triton Server.')
+        else:
+            try:
+                # Possible that an error left a stopped triton container
+                self._tritonserver_container = self._docker_client.containers.get(
```
Tabrizian, line 184: Isn't this going to mistakenly remove other containers named `tritonserver`? I believe we do not set the name of the container to be `tritonserver`. It is automatically chosen using the Docker API.

Tabrizian (changes requested):

model_analyzer/triton/server/server_docker.py
aramesh7: We set the name of our triton server container to be `tritonserver`. The goal of this is to remove any containers named `tritonserver`. We've been running into some issues where if there is an error launching the container, there is a stopped container named `tritonserver` that prevents other tests from running on that node.

qa/L0_state_management/test.sh
aramesh7: It is unnecessary because we are using `--override-output-model-repository`

model_analyzer/triton/server/server_docker.py
Tabrizian: I see. I was wrong about not setting the container name. But still if the user has another container named `tritonserver` (unrelated to MA) and tries to run the MA, is this going to remove that container?

model_analyzer/config/input/config_command_profile.py
aramesh7: It simplifies processing this argument and verifying the argument (check for 2 ':' in the string). I don't see a huge advantage in affording the user the flexibility to leave the mode unspecified and to use a default. Do you think this is worth it?

qa/L0_server_launch_modes/test.sh
aramesh7: There were instances where model-analyzer would be killed but triton server remains running. If there was an exception thrown by the docker API, there would be a stopped (but not removed) container named tritonserver.

model_analyzer/triton/server/server_docker.py
aramesh7: Yes, perhaps the exact name should be changed? Or maybe We should ignore this problem altogether and not name our container? The downside to that approach is that we won't know if we're improperly clearning up stopped containers.

aramesh7: Let me try removing the name for now. With the labels we are adding to the CI tests (RUNNER_ID=$RUNNER_ID), the cleanup daemon will remove our stopped containers.

Tabrizian: I think we used to not name the container.
> The downside to that approach is that we won't know if we're improperly clearning up stopped containers. 

I think if we are able to find the root cause for when the dangling container is created, we wouldn't need this try block. Alternatively, we can extract the name of the container assigned by docker and delete it here.

model_analyzer/config/input/config_command_profile.py
Tabrizian: Yeah, I agree it is fine in its current mode too. It was only a suggestion.

model_analyzer/triton/server/server_docker.py
aramesh7: The root cause I've found is what this try block aims to handle. So in certain cases (such as if the image is empty), the docker API throws an exception. In this case, it seems the container is started, exits, and is stopped. Now we don't have a handle (`self._tritonserver_container` is `None`), and there is a stopped container named `tritonserver`. 

We can add checks that prevent these types of scenarios from occuring. In this PR I added a null check for the image, but this try/block is to handle the case where something goes wrong. 

The only reason to keep a named container is to expose cases where we are hanging containers it seems like.

tests/mocks/mock_server_docker.py
```diff
@@ -92,7 +92,9 @@ def assert_server_process_start_called_with(self,
                                                 gpu_uuids,
                                                 http_port=8000,
                                                 grpc_port=8001,
-                                                metrics_port=8002):
+                                                metrics_port=8002,
+                                                mounts=[],
```
Tabrizian, line 96: Using dictionary and lists in the default arguments can lead to strange issues. See here: https://docs.python-guide.org/writing/gotchas/

Tabrizian (commented):
Except one minor comment the rest LGTM.

tests/mocks/mock_server_docker.py
aramesh7: Ah wow I didn't know this. Thanks for pointing this out.

Tabrizian (dismissed):

model_analyzer/config/input/config_command_profile.py
```diff
@@ -217,7 +221,8 @@ def _add_repository_configs(self):
                 field_type=ConfigPrimitive(str),
                 default_value=DEFAULT_OUTPUT_MODEL_REPOSITORY,
                 flags=['--output-model-repository-path'],
-                description='Output model repository path used by Model Analyzer.'
+                description=
+                'Output model repository path used by Model Analyzer.'
```
Tabrizian, line 225: Are you using the `format.py` for formatting the code? It looks like you make some formatting changes and Jacky again reverts them. Can you please check whether the code is formatted using `format.py` or not?

Tabrizian (approved):

model_analyzer/config/input/config_command_profile.py
aramesh7: Yes, I thought that the yapf version used was the same but its not, I will address this.

aramesh7: The best thing to do for this might be to have a separate pull request that formats `model_analyzer/*` and `test/*` with format.py, because we'll end up with far more files changed than this PR.

PR #208, opened 2021-09-08
Added RemoteMonitor to query metrics directly from Triton metrics

aramesh7:


docs/config.md
```diff
@@ -81,7 +81,10 @@ profile_models: <comma-delimited-string-list>
 [ max_retries: <int> | default: 100 ]
 
 # Specifies how long (seconds) to gather server-only metrics
-[ duration_seconds: <int> | default: 5 ]
+[ duration_seconds: <int> | default: 2 ]
+
+# Specify whether DCGM should be used by Model Analyzer to collect GPU metricss
+[ use_local_gpu_monitor: <bool> | default: False ]
```
Tabrizian, line 87: Can we control the monitoring interval for remote monitor? If no, we can perhaps not expose this option until the full support for remote mode is added.

model_analyzer/record/metrics_manager.py
```diff
@@ -140,6 +144,7 @@ def profile_server(self):
 
         cpu_only = (not numba.cuda.is_available())
         self._start_monitors(cpu_only=cpu_only)
+        time.sleep(self._config.duration_seconds)
```
Tabrizian, line 147: Does this mean that previously we were not waiting enough for collecting Server only metrics?

tests/test_dcgm_monitor.py
```diff
@@ -61,7 +61,7 @@ def test_record_memory(self):
 
         # Assert instance types
         for record in records:
-            self.assertIsInstance(record.device(), GPUDevice)
```
Tabrizian, line 64: Why the device is no longer a GPUDevice?

Tabrizian (changes requested):

docs/config.md
aramesh7: Im working on adding it on the server side, perhaps I should have done it first.

model_analyzer/record/metrics_manager.py
aramesh7: There were intermittent failures. This became apparent when switching to the remote monitor.

tests/test_dcgm_monitor.py
aramesh7: The records now just store the UUID. This makes it so that only the GPUDeviceFactory creates and keeps track of GPUDevices, and to prevent needing to get extra information in the remote monitor/dcgm monitor.

docs/config.md
Tabrizian: I mean if we are tight on schedule we can just checkout the code and keep the functionality the same as previous releases.

aramesh7: I think we might be able to make it hopefully within the hour/before EOD

Tabrizian (approved):

Tabrizian (approved):

PR #211, opened 2021-09-10
Added metrics/monitor stability tests

aramesh7:


qa/L0_stability_metrics/check_results.py
```diff
@@ -0,0 +1,118 @@
...
+    def check_metrics_stability(self):
+        """
+        Makes sure that the same configuration
+        appears as the best across iterations for
+        each model
+        """
+
+        # There should be 4 csv files the results path
+        pathname = os.path.join(self._result_path, 'result_*.csv')
+        csv_contents = []
```
Tabrizian, line 50: Maybe add a check to make sure that 4 CSV files are generated?

Tabrizian (changes requested):

qa/L0_stability_metrics/check_results.py
aramesh7: Sorry that comment is out of place. I will remove it. The number of csv files is set by the test.sh

Tabrizian (approved):

PR #214, opened 2021-09-13
Updated docs and setup.py for publishing wheel to PyPI

aramesh7:


docs/install.md
```diff
@@ -16,138 +16,136 @@ limitations under the License.
...
+## Using `pip3`
+
+You can install pip using:
+```
+$ sudo apt-get update && sudo apt-get install python3-pip
+```
+
+Model analyzer can be installed with: 
+```
+$ pip3 install nvidia-pyindex
```
Tabrizian, line 101: When we publish MA on PyPi, this line is no longer required.

docs/install.md
```diff
@@ -16,138 +16,136 @@ limitations under the License.
...
+
+You can also build the Model Analyzer's dockerfile yourself. First, clone the
+Model Analyzer's git repository, then build the docker image.
+
+```
+$ git clone https://github.com/triton-inference-server/model_analyzer
+$ docker build --pull -t model-analyzer .
+```
+
+The above command will pull all the containers that model analyzer needs to run.
```
Tabrizian, line 78: model analyzer -> Model Analyzer

docs/install.md
```diff
@@ -16,138 +16,136 @@ limitations under the License.
...
+```
+$ sudo apt-get update && sudo apt-get install wkhtmltopdf
+```
+
+Once you do this, Model Analyzer will able to use `pdfkit` to generate reports.
+
+## Building the Dockerfile
+
+You can also build the Model Analyzer's dockerfile yourself. First, clone the
+Model Analyzer's git repository, then build the docker image.
```
Tabrizian, line 71: Could you add a sentence that this is the recommended installation for local mode?

Tabrizian (changes requested):

docs/install.md
aramesh7: The recommended is not the Triton SDK? Why are we recommending it by mode?

Tabrizian: > The recommended is not the Triton SDK?

Yes, but I was saying that if someone wants to use the local mode what is the easiest way to do so? I think "Building the Dockerfile" is the easiest way to use the local mode. Maybe not the "recommended way" but the easiest way to use the local mode.

aramesh7: Gotcha. I will make the note!

aramesh7: In my opinion this is the easiest way for all launch modes. Do you find that to be the case as well?

Tabrizian: Yes, I agree. It doesn't have the problems of Docker mode. There is a ticket to install MA in the main Triton container. If that ticket is done, then we can change the "recommended way" to using the main Triton container.

docs/install.md
```diff
@@ -16,138 +16,144 @@ limitations under the License.
...
+```
+
+Once you do this, Model Analyzer will able to use `pdfkit` to generate reports.
+
+## Building the Dockerfile
+
+You can also build the Model Analyzer's dockerfile yourself. First, clone the
+Model Analyzer's git repository, then build the docker image. This is the
+recommended installation method if you mainly intend to use
+`--triton-launch-mode=docker`, as all the dependencies will be available.
```
Tabrizian, line 80: `triton-launch-mode=local`

Tabrizian (changes requested):

Tabrizian (approved):

PR #238, opened 2021-11-09
Test all record types

cvinson830:
The record types in the model-analyzer.record.types package is contextual when it uses 'less than' (<) and 'greater than' (>) operators.  The 'less than' and 'greater than' operators are overloaded to mean 'worse than' and 'better than' respectively.

- Some record types treat MORE as better (eg, gpu_free_memory, cpu_available_ram)
- Other record types treat LESS as better (eg, gpu_used_memory, cpu_used_ram)

So, when comparing two objects of type cpu_used_ram  12 > 13 is actually true, since 12 'is better than' 13.
Based on this context, the code is working fine.

-------------------------------------------------------------------------------------------------------

However, not all record types were being tested.  
This PR closes the gap so that all record types are tested in two ways:

1. Make sure that the count of 'less is better tests' plus the count of 'more is better tests' equals the total number of record types.
2. Update tests that use 'less is better' and 'more is better' to fail if it's not accounted for

### DIFF
--- a/tests/test_record_types.py
+++ b/tests/test_record_types.py
@@ -19,6 +19,21 @@ from .common import test_result_collector as trc
 
 
 class TestRecordAggregatorMethods(trc.TestResultCollector):
+    """
+    The record types in the model-analyzer.record.types package are contextual 
+        when it uses 'less than' (<) and 'greater than' (>) operators. 
+    
+    The 'less than' and 'greater than' operators are overloaded to 
+        mean 'worse than' and 'better than' respectively.
+
+    Some record types treat MORE as better 
+        (eg, gpu_free_memory, cpu_available_ram)
+    Other record types treat LESS as better 
+        (eg, gpu_used_memory, cpu_used_ram)
+        
+    So, when comparing two objects of type 'cpu_used_ram' 
+        12 > 13 is actually true, since 12 'is better than' 13.
+    """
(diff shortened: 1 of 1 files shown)


tgerdesnv (changes requested):
Please add a high level description of the file/class near line 21

cvinson830:
> Please add a high level description of the file/class near line 21

Done.

tgerdesnv (approved):

tests/test_record_types.py
```diff
@@ -104,12 +134,16 @@ def test_compare(self):
                 self.assertTrue(metric1 < metric2)
             elif record_type in self.more_is_better_types:
                 self.assertTrue(metric2 < metric1)
+            else:
+                fail
```
Tabrizian, line 138: How about using `self.assertTrue(False, 'An error message')` instead of an undefined variable?

tests/test_record_types.py
```diff
@@ -64,6 +92,8 @@ def test_sub(self):
                 self.assertEqual(metric3.value(), -7)
             elif record_type in self.more_is_better_types:
                 self.assertEqual(metric3.value(), 7)
+            else:
+                fail
```
Tabrizian, line 96: @cvinson830 Same as below.

PR #241, opened 2021-11-11
Print help when no args

matthewkotila:
model analyzer should print the help message when no arguments are provided, rather than raising an exception like before

before:
```
$ model-analyzer
Traceback (most recent call last):
  File "/usr/local/bin/model-analyzer", line 33, in <module>
    sys.exit(load_entry_point('triton-model-analyzer', 'console_scripts', 'model-analyzer')())
  File "/opt/triton-model-analyzer/model_analyzer/entrypoint.py", line 274, in main
    args, config = get_cli_and_config_options()
  File "/opt/triton-model-analyzer/model_analyzer/entrypoint.py", line 211, in get_cli_and_config_options
    return cli.parse()
  File "/opt/triton-model-analyzer/model_analyzer/cli/cli.py", line 137, in parse
    config = self._subcommand_configs[args.subcommand]
KeyError: None
```

after:
```
$ model-analyzer
usage: model-analyzer [-h] [-q] [-v] [-m {online,offline}] {profile,analyze,report} ...

positional arguments:
  {profile,analyze,report}
                        Subcommands under Model Analyzer
    profile             Run model inference profiling based on specified CLI or config options.
    analyze             Collect and sort profiling results and generate data and summaries.
    report              Generate detailed reports for a single config

optional arguments:
  -h, --help            show this help message and exit
  -q, --quiet           Suppress all output except for error messages.
  -v, --verbose         Show detailed logs, messags and status.
  -m {online,offline}, --mode {online,offline}
                        Choose a preset configuration mode.
```

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,73 @@
+# Copyright (c) 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+from argparse import ArgumentParser
+
+from .common import test_result_collector as trc
+
+from model_analyzer.config.input.config_command_profile import ConfigCommandProfile
+from model_analyzer.config.input.config_command_analyze import ConfigCommandAnalyze
+from model_analyzer.config.input.config_command_report import ConfigCommandReport
+from model_analyzer.cli.cli import CLI
+
+
+class ArgumentParserSubclass(ArgumentParser):
+
+    def print_help(self, file=None):
+        super().print_help(file)
+        self._did_print_help = True
+
+    _did_print_help = False
+
+
+class CLISubclass(CLI):
+
+    def __init__(self):
+        self._parser = ArgumentParserSubclass()
+        self._add_global_options()
+        self._subparsers = self._parser.add_subparsers(
+            help='Subcommands under Model Analyzer', dest='subcommand')
+
+        # Store subcommands, and their configs
+        self._subcommand_configs = {}
+
+
+class TestCLI(trc.TestResultCollector):
+
+    def test_help_message_no_args(self):
+        sys.argv = ['/usr/local/bin/model-analyzer']
+
+        config_profile = ConfigCommandProfile()
+        config_analyze = ConfigCommandAnalyze()
+        config_report = ConfigCommandReport()
+
+        cli = CLISubclass()
+        cli.add_subcommand(
+            cmd='profile',
+            help=
+            'Run model inference profiling based on specified CLI or config options.',
+            config=config_profile)
+        cli.add_subcommand(
+            cmd='analyze',
+            help=
+            'Collect and sort profiling results and generate data and summaries.',
+            config=config_analyze)
+        cli.add_subcommand(cmd='report',
+                           help='Generate detailed reports for a single config',
+                           config=config_report)
+
+        self.assertRaises(SystemExit, cli.parse)
+        self.assertTrue(cli._parser._did_print_help == True)
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -25,6 +25,10 @@ from model_analyzer.cli.cli import CLI
 
 
 class ArgumentParserSubclass(ArgumentParser):
+    """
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -0,0 +1,73 @@
...
+        self._subparsers = self._parser.add_subparsers(
+            help='Subcommands under Model Analyzer', dest='subcommand')
+
+        # Store subcommands, and their configs
+        self._subcommand_configs = {}
+
+
+class TestCLI(trc.TestResultCollector):
+
+    def test_help_message_no_args(self):
```
tgerdesnv, line 50: Please comment the test, as well as the above classes

Also, is "subclass" the appropriate term?  Didn't we use Test* in DLProf:   PytorchIterationTracker vs TestPytorchIterationTracker.  Does that not work because this testing framework looks for anything that starts with Test* to run?

tgerdesnv (changes requested):

tests/test_cli.py
```diff
@@ -0,0 +1,73 @@
...
+    def print_help(self, file=None):
+        super().print_help(file)
+        self._did_print_help = True
+
+    _did_print_help = False
+
+
+class CLISubclass(CLI):
+
+    def __init__(self):
```
tgerdesnv, line 38: Instead of re-implementing the function, I think you should use super().__init__():

```
def __init__(self):
    super().__init__()
    self._parser = ArgumentParserSubclass()
```

tgerdesnv (changes requested):

debermudez (commented):
Other than Tim's changes, this seems good to go.

tests/test_cli.py
matthewkotila: added comments 👍

to use python [unittest](https://docs.python.org/3/library/unittest.html#basic-example), you basically create a subclass of `unittest.TestCase`. that subclass has to have a name, of course, and in this case, the file I added is supposed to test the CLI, so the files name is `test_cli.py` and the subclass of `unittest.TestCase` has the same name (TestCLI), to match the pattern of other unit tests (e.g. `test_run_search.py` has `TestRunSearch` as the subclass of `unittest.TestCase`)

that left me in a conundrum, because we used to do the thing where if we were trying to write a test for `SomeSpecificClass`, but we needed to change functionality of a particular method to test it, we'd make a subclass of `SomeSpecificClass` called `TestSomeSpecificClass`, but that clearly can't work with the above naming pattern

hence i used `SomeSpecificClassSubclass`

any suggestions on alternative ways of handling this naming pattern conflict?

tests/test_cli.py
matthewkotila: this doesn't work

```
Traceback (most recent call last):
  File "/opt/triton-model-analyzer/tests/test_cli.py", line 88, in test_help_message_no_args
    self.assertRaises(SystemExit, cli.parse)
  File "/usr/lib/python3.8/unittest/case.py", line 816, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/usr/lib/python3.8/unittest/case.py", line 202, in handle
    callable_obj(*args, **kwargs)
  File "/opt/triton-model-analyzer/model_analyzer/cli/cli.py", line 137, in parse
    if args.subcommand is None:
AttributeError: 'Namespace' object has no attribute 'subcommand'
```

if i call `super().__init__()`, it will already create a regular ArgumentParser object and do things with it. by the time i can reassign the _parser data member to my own custom subclass, it's too late, and a future call to `cli.parse()` have its state messed up

tgerdesnv: You're right :(.  Hardcoded class inside of constructor is tough to work around.

tgerdesnv: As mentioned on the phone, lets try to use mock/patch for this. I think it fits in seamlessly and results in both of your subclasses going away

tests/test_cli.py
jbkyang-nvi: Maybe you can mock with google mock (Tim mentioned this in another comment) so that `ArgumentParserSubclass` can just be a mocked internal member?

jbkyang-nvi (commented):
Were you going to be inheriting from unittest?

tests/test_cli.py
matthewkotila: switched to using [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) 👍

tests/test_cli.py
matthewkotila: switched to using [unittest.mock](https://docs.python.org/3/library/unittest.mock.html) 👍

tests/test_cli.py
```diff
@@ -0,0 +1,88 @@
...
+            cmd='analyze',
+            help=
+            'Collect and sort profiling results and generate data and summaries.',
+            config=config_analyze)
+        cli.add_subcommand(cmd='report',
+                           help='Generate detailed reports for a single config',
+                           config=config_report)
+
+        self.assertRaises(SystemExit, cli.parse)
+        self.assertTrue(cli._parser._did_print_help == True)
```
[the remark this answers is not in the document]
matthewkotila: this is now `mock_print_help.assert_called()` 👍

tgerdesnv (approved):

PR #242, opened 2021-11-11
Update quick start

matthewkotila:
check out the new quick start here:

https://github.com/triton-inference-server/model_analyzer/blob/matthewkotila-update-docs/docs/quick_start.md

### DIFF
--- a/docs/quick_start.md
+++ b/docs/quick_start.md
@@ -34,6 +34,7 @@ $ docker build --pull -t model-analyzer .
 $ docker run -it --rm --gpus all \
     -v /var/run/docker.sock:/var/run/docker.sock \
     -v $HOME/model_analyzer/examples/quick-start:/quick_start_repository \
+    -w /quick_start_repository \
     --net=host --name model-analyzer \
     model-analyzer /bin/bash
 ```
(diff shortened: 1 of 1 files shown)


docs/quick_start.md
```diff
@@ -45,35 +46,40 @@ libtorch model which calculates the sum and difference of two inputs. Run the
 Model Analyzer `profile` subcommand inside the container with:
 
 ```
-$ model-analyzer profile -m /quick_start_repository/ \
-    --profile-models add_sub \
-    --run-config-search-max-concurrency 2 \
-    --run-config-search-max-instance-count 2 \
-    --run-config-search-preferred-batch-size-disable true
+$ model-analyzer profile -m . --profile-models add_sub
```
dzier, line 49: Why do we need `-m .`, that is even worse. It obfuscates what it is asking. Should use the full option as well if you are using it.

docs/quick_start.md
```diff
@@ -45,35 +46,40 @@ libtorch model which calculates the sum and difference of two inputs. Run the
...
 ```
 
 If you already ran this earlier in the container, you can use the `--override-output-model-repository` option to overwrite the earlier results.
 
 This will perform a search across various config parameters on the `add_sub`
```
dzier, line 54: Remove `various` and say "limited configurable model parameters"

docs/quick_start.md
```diff
@@ -45,35 +46,40 @@ libtorch model which calculates the sum and difference of two inputs. Run the
...
-model. This should take no more than a few minutes to finish. Here is some sample output:
+model. This will take around 60 minutes to finish. If you want a shorter run (1-2 minutes) for example purposes, you can run with the below additional options. Note that these options are not intended to find the best configuration:
```
dzier, line 55: "This can take up to 60 minutes..."

docs/quick_start.md
```diff
@@ -45,35 +46,40 @@ libtorch model which calculates the sum and difference of two inputs. Run the
...
 
 ```
+--run-config-search-max-concurrency 2 \
+--run-config-search-max-instance-count 2 \
+--run-config-search-preferred-batch-size-disable true
+```
+
```
dzier, line 62: You should state here what those parameters do and how may configs will be tested.

dzier (changes requested):

docs/quick_start.md
matthewkotila: `model-analyzer profile` requires the user to supply the model option (`-m`):

```
$ model-analyzer profile --profile-models add_sub
2021-11-11 21:56:29.213 ERROR[entrypoint.py:214] Model Analyzer encountered an error: Config for model_repository is not specified. You need to specify it using the YAML config file or using the -m, --model-repository flags in CLI.
```

docs/quick_start.md
matthewkotila: reworded 👍

docs/quick_start.md
matthewkotila: reworded 👍

docs/quick_start.md
matthewkotila: added 👍

docs/quick_start.md
matthewkotila: reverted back to using `-m /quick_start_repository` 👍

dzier (approved):

PR #245, opened 2021-11-12
Add testing for model_manager

tgerdesnv:
Added testing of all config combinations for model_manager

### DIFF
--- a/model_analyzer/config/run/run_search.py
+++ b/model_analyzer/config/run/run_search.py
@@ -139,6 +139,7 @@ class RunSearch:
         # Reset the measurements after each init
         self._measurements = []
         if len(concurrency) != 0 and search_model_config_parameters:
+            logger.info('Will sweep only through model config parameters...')
             self._model_config_parameters = {'instance_count': 0}
             self._sweep_mode_function = self._sweep_model_config_only
         elif len(concurrency) == 0 and search_model_config_parameters:
@@ -151,9 +152,9 @@ class RunSearch:
             logger.info('Will sweep only through the concurrency values...')
             self._sweep_mode_function = self._sweep_concurrency_only
 
-    def get_model_sweep(self, config_model):
+    def get_next_model_sweep(self, config_model):
         """
-        Get the next iteration of the sweeps.
+        Get the next iteration of the sweep
 
         Parameters
         ----------
@@ -181,6 +182,7 @@ class RunSearch:
             # Only log message if there is new runs.
             if model_sweep:
                 self._log_message(new_model)
+
             return new_model, model_sweep
         return new_model, []
 
--- a/tests/test_model_manager.py
+++ b/tests/test_model_manager.py
@@ -1,0 +1,294 @@
+# Copyright (c) 2021 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from .common import test_result_collector as trc
+
+from .mocks.mock_config import MockConfig
+from .mocks.mock_os import MockOSMethods
+from .mocks.mock_model_config import MockModelConfig
+from .mocks.mock_client import MockTritonClientMethods
+
+from model_analyzer.cli.cli import CLI
+from model_analyzer.config.input.config_command_profile import ConfigCommandProfile
+from model_analyzer.config.run.run_search import RunSearch
+from model_analyzer.config.run.run_config_generator import RunConfigGenerator
+from model_analyzer.constants import LOGGER_NAME
+from model_analyzer.model_manager import ModelManager
+from model_analyzer.state.analyzer_state_manager import AnalyzerStateManager
+
+import itertools
+import logging
+import sys
+
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+
+
+class ModelManagerSubclass(ModelManager):
+    """ 
+    Overrides execute_run_configs() to gather a list of tuples that contain 
+    the main values of each 'executed' run_config
+    """
+
+    def __init__(self, config, client, server, metrics_manager, result_manager, state_manager):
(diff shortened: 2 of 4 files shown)


tests/test_model_manager.py
```diff
@@ -0,0 +1,295 @@
...
+
+            instances = None
+            if model_config.get("instanceGroup") is not None:
+                instances = model_config["instanceGroup"][0]["count"]
+
+            dynamic_batching = None
+            if model_config.get("dynamicBatching") is not None:
+                dynamic_batching = model_config["dynamicBatching"].get("preferredBatchSize", [0])[0]
+
+            # FIXME __getitem__
```
tgerdesnv, line 61: Fixme -- I don't have any good way of accessing data in PerfConfig.  I'll likely add functions or extend the class here as a Subclass for testing

tests/test_model_manager.py
```diff
@@ -0,0 +1,295 @@
...
+            batch_size = perf_config.__getitem__("batch-size")
+            concurrency = perf_config.__getitem__("concurrency-range")
+
+            self._configs.append((instances, dynamic_batching, batch_size, concurrency))
+
+    def get_run_configs(self):
+        # FIXME not sure why this is necessary
```
tgerdesnv, line 68: Fixme -- not sure why I have to manually reset this. I would expect each test to have its own object that starts out empty?

tests/test_model_manager.py
```diff
@@ -0,0 +1,295 @@
...
+        tmp = self._configs.copy()
+        self._configs.clear()
+        return tmp
+
+
+@patch('model_analyzer.config.run.run_search.RunSearch.add_measurements', MagicMock())
```
tgerdesnv, line 74: Here is the magic that removes the add_measurements function for every test.  It would otherwise cause problems due to so many other things being mocked

tests/test_run_search.py
```diff
@@ -57,6 +57,105 @@ def setUp(self):
     def tearDown(self):
         self.mock_os.stop()
 
+
+    def test_full_sweep(self):
```
tgerdesnv, line 61: Not sure whether to keep these tests or not. The tests in model_manager are a superset of these. Also, the tests in model_manager are better because they verify the full set of config combinations instead of just the expected count

tests/test_model_manager.py
```diff
@@ -0,0 +1,295 @@
...
+
+            dynamic_batching = None
+            if model_config.get("dynamicBatching") is not None:
+                dynamic_batching = model_config["dynamicBatching"].get("preferredBatchSize", [0])[0]
+
+            # FIXME __getitem__
+            batch_size = perf_config.__getitem__("batch-size")
+            concurrency = perf_config.__getitem__("concurrency-range")
+
+            self._configs.append((instances, dynamic_batching, batch_size, concurrency))
```
tgerdesnv, line 65: How bad is it that there is an assumed order here (instances first, dynamic_batching second, etc) that has to match the order of expected_values set for every test?

tests/test_run_search.py
debermudez: I would remove them. It makes for less to maintain

tests/test_model_manager.py
tgerdesnv: I solved this (but haven't uploaded yet), and it is important for us to understand. I haven't gotten to the class section of the python tutorial, so maybe you guys already know this.
https://docs.python.org/2/tutorial/classes.html#class-and-instance-variables

tests/test_model_manager.py
tgerdesnv: I'm going to just remove the FIXME. I don't see the point in creating a fake test class to extend this just to 'expose' the function

tests/test_run_search.py
tgerdesnv: removed

tests/test_model_manager.py
tgerdesnv: ok, leaving as-is for now

tests/test_model_manager.py
```diff
@@ -0,0 +1,294 @@
...
+            cmd='profile',
+            help=
+            'Run model inference profiling based on specified CLI or config options.',
+            config=config)
+        cli.parse()
+        mock_config.stop()
+        return config    
+
+    def setUp(self):
+        # Use mock model config or else TritonModelAnalyzerException will be thrown as it tries to read from disc
```
matthewkotila, line 288: *disk

debermudez (approved):
discussed all changes offline. good to go

tests/test_model_manager.py
```diff
@@ -0,0 +1,294 @@
...
+import itertools
+import logging
+import sys
+
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+
+
+class ModelManagerSubclass(ModelManager):
```
matthewkotila, line 40: this should be renamed to `MockModelManager` and be put in `mock_model_manager.py` in `mocks` directory

tests/test_model_manager.py
```diff
@@ -0,0 +1,294 @@
...
+
+            batch_size = perf_config.__getitem__("batch-size")
+            concurrency = perf_config.__getitem__("concurrency-range")
+
+            self._configs.append((instances, dynamic_batching, batch_size, concurrency))
+
+    def get_run_configs(self):
+        return self._configs
+
+
```
matthewkotila, line 72: can't comment on the line below because Elias already put a thread on there and GitHub is weird

no need to specify `MagicMock()` as the next argument--it's the default:

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

model_analyzer/config/run/run_search.py
```diff
@@ -139,6 +139,7 @@ def init_model_sweep(self, concurrency, search_model_config_parameters):
         # Reset the measurements after each init
```
matthewkotila, line 139: did all of these changes get formatted with `format.py` in the repo root?

matthewkotila:
did all of these changes get formatted with format.py in the repo root?

matthewkotila:
oh, just noticed it was merged--whoops

tests/test_model_manager.py
tgerdesnv: This isn't a generic mocked ModelManager that should be used by any other tests. This is very specifically part of testing in this file.

tests/test_model_manager.py
tgerdesnv: I'll keep that in mind next time. While we are still learning this all I think it is best to be explicit

tgerdesnv: Also, this doesn't work as-is.  If I omit the `MagicMock()` I would then need to pass in an extra arg into every test function (I confirmed that all the tests break without further edits):

_If patch() is used as a decorator and new is omitted, the created mock is passed in as an extra argument to the decorated function._

PR #248, opened 2021-11-18
Validate triton_server_path inside local launch mode

cvinson830:
- config_command_profile.py: Disengage `binary_path_validator` validator from `triton-server-path`'s ConfigField object
- entrypoint.py: Within launch_mode `remote` call `binary_path_validator` directly.


model_analyzer/config/input/config_command.py
```diff
@@ -96,6 +96,11 @@ def set_config_values(self, args):
         else:
             yaml_config = None
 
+        # Remove 'triton_server_path' validator when in 'remote' mode
```
tgerdesnv, line 99: I've started to dig into the CLI/config code for a different story and wow is it involved.  This file you are touching is a very generic high level class that isn't intended to have any knowledge of any specific options.  I don't think this is the correct place to make the change.  I'm going to dig around and try to figure out how all this works.

tgerdesnv (changes requested):

cvinson830:
I want to run a real "remote" run too to make sure the intended behavior works in production

model_analyzer/config/input/config_command.py
cvinson830: Ack...  I'll look into those.  The tricky part is finding a place where `args` and `config` are both in scope.

cvinson830: After review of suggestion: The exception gets thrown during `set_value`.  These set_value calls are inside the loop immediately following my addition to config_command.py.  So the removal of the validation needs to be ahead of those `set_value` calls.  Removing the validation at the time of launching triton will be too late.  Still open for suggestions.

cvinson830: [slacked offline] The new approach is to, by default, opt-out validation for the field 'triton_server_path' for all launch modes.  For 'local' launch mode, opt-in validation using `binary_path_validator`.

cvinson830: 'local' launch mode code now calls binary_path_validator() directly, circumventing the config's internal validator.

model_analyzer/config/input/config_value.py
```diff
@@ -237,4 +237,4 @@ def set_name(self, name):
             New name to be set for this field.
         """
 
-        self._name = name
+        self._name = name
```
tgerdesnv, line 240: lets just revert the file. shouldn't be any changes.

model_analyzer/entrypoint.py
```diff
@@ -92,6 +93,11 @@ def get_server_handle(config, gpus):
             ' Model Analyzer does not have access to the model repository of'
             ' the remote Triton Server.')
     elif config.triton_launch_mode == 'local':
+        path = config.get_config()['triton_server_path'].value()
```
tgerdesnv, line 96: I know all of the existing code desperately needs a refactor for cleanliness, but lets make sure the new code is clean. Put all of this added code into a function called validate_triton_server_path()

tests/test_run_triton_server_path.py
```diff
@@ -0,0 +1,105 @@
...
+            '--triton-launch-mode', 'remote', 
+        ]
+        yaml_content = """
+            concurrency: [1, 2, 3]
+            batch_sizes: [2, 3, 4]
+            profile_models:
+            - vgg_16_graphdef
+            """
+        # No exception: path is not validated intentionally in remote mode
+        self._evaluate_config(args, yaml_content)
```
tgerdesnv, line 65: Looks like these tests don't really do anything anymore since the validation was removed from the config evaluation and now lives up in entrypoint.  If you wanted to add testing it would likely need to be an L0 test.  It looks like there is a test called L0_server_launch_modes which may be the correct place to put this.  If it is already launching in different modes, you could update the non-local tests with a corrupt triton_server_path and make sure it still runs correctly.

model_analyzer/entrypoint.py
```diff
@@ -24,7 +24,8 @@
 from .config.input.config_command_profile import ConfigCommandProfile
 from .config.input.config_command_analyze import ConfigCommandAnalyze
 from .config.input.config_command_report import ConfigCommandReport
-
```
tgerdesnv, line 27: remove endline (so all imports are together). Add the constants import to line 19 instead of a new line.

tgerdesnv (changes requested):

model_analyzer/config/input/config_value.py
cvinson830: Fixed

tests/test_run_triton_server_path.py
cvinson830: Removed this file.  Added `set_triton_server_path()` to `test_config_generator.py` (awaiting to see what happens in CI)

model_analyzer/entrypoint.py
cvinson830: Done

model_analyzer/config/input/config_command.py
cvinson830: Done

model_analyzer/entrypoint.py
cvinson830: Added `validate_triton_server_path()`

model_analyzer/entrypoint.py
cvinson830: Done

qa/L0_server_launch_modes/test_config_generator.py
```diff
@@ -90,5 +91,17 @@ def generate_configs(self):
                         yaml.dump(self.config, f)
 
 
+    def set_triton_server_path(self, launch_mode, config):
+        """
+        Helper function to set the 'triton_server_path' to something bogus
+            for all configuration except 'local'
+        This ensures only the 'local' launch mode uses 'triton_server_path'.
+        """
+        config.pop('triton_server_path', None)
```
tgerdesnv, line 100: Doesn't this remove the triton server path for local?

tgerdesnv (approved):

PR #249, opened 2021-11-19
Update colors for detail report legend

cvinson830:
1) View the labels inside six of the colors below.
- Notice three colors of the 'server compute' co-related items are 1) from the same shade, and 2) in the cool palette
- The other three unrelated items are 1) not of the same shade, and 2) from the warm palette

image

2) Resulting plot (for presentation purposes, some values were spoofed to show their contribution to the bar from a color perspective)
image

3) Fixed spacing
image

- Old (starting point before this PR)
image

### DIFF
--- a/model_analyzer/plots/detailed_plot.py
+++ b/model_analyzer/plots/detailed_plot.py
@@ -74,7 +74,7 @@ class DetailedPlot:
             'perf_server_queue': '#825D16',
             'perf_server_compute_input': '#0071C5',
             'perf_server_compute_infer': '#C5B600',
-            'perf_server_compute_output': '#F0D4BF',
+            'perf_server_compute_output': '#C55400',
             'perf_throughput': '#5E5E5E'
         }
 
--- a/model_analyzer/plots/detailed_plot.py
+++ b/model_analyzer/plots/detailed_plot.py
@@ -69,12 +69,12 @@ class DetailedPlot:
         ]
 
         self._bar_colors = {
-            'perf_client_send_recv': '#4285f4',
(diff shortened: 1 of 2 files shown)


tgerdesnv:
I know the proposal was to mix warm and cool, but after seeing it I don't like it. The blues pop, but the 'warm' colors just look muted and dull.  A few possibilities:
1) Only use cool colors. Shades of blue for compute, and non-blue colors for the rest
2) Still use warm and cool, but eliminate red60, neutral 60, and dark neutral (which to me are the dull ones)
3) Swap warm and cool since a majority of the plot will be non-compute. Basically have compute be orange/red, and the others be cool colors (but not all blue)

cvinson830:
> I know the proposal was to mix warm and cool, but after seeing it I don't like it. The blues pop, but the 'warm' colors just look muted and dull. A few possibilities:
> 
> 1. Only use cool colors. Shades of blue for compute, and non-blue colors for the rest
> 2. Still use warm and cool, but eliminate red60, neutral 60, and dark neutral (which to me are the dull ones)
> 3. Swap warm and cool since a majority of the plot will be non-compute. Basically have compute be orange/red, and the others be cool colors (but not all blue)

Option #2 selected

tgerdesnv:
Looks good to me, although not all colors are showing. Is it hard to fudge the numbers in the chart and regenerate?

cvinson830:
> Looks good to me, although not all colors are showing. Is it hard to fudge the numbers in the chart and regenerate?

I updated the comment up above.  Here it is for convenience
image

tgerdesnv (approved):

PR #250, opened 2021-11-19
Add next commands to cli output

matthewkotila:
This PR adds suggestions to the user on what command to run next after both the `profile` and `analyze` commands

i.e.
* after `profile` command is run, the `analyze` command is pre-created and output to the user in the shell for them to copy and paste/run
* after `analyze` command is run, the `report` command is given

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -119,9 +119,17 @@ class Analyzer:
         profiled_model_list = list(
             self._state_manager.get_state_variable(
                 'ResultManager.results').keys())
+
+        num_profiled_configs = sum([
+            len(x) for x in self._state_manager.get_state_variable(
+                'ResultManager.results').values()
+        ])
         logger.info(
-            f"Finished profiling. Obtained measurements for models: {profiled_model_list}."
-        )
+            f"Profile complete. Profiled {num_profiled_configs} configurations "
+            f"for models: {profiled_model_list}. To analyze the profile "
+            f"results and find the best configurations, please run "
+            f"`model-analyzer analyze --analysis-models "
+            f"{','.join(profiled_model_list)}`.")
 
     def analyze(self, mode, quiet):
         """
@@ -164,6 +172,16 @@ class Analyzer:
         self._result_manager.export_results()
         if not quiet:
             self._result_manager.write_results()
+
+        top_3_model_config_names = [
+            x.model_config().get_config()['name']
+            for x in self._result_manager.top_n_results(n=3)
+        ]
+        logger.info(f"Run `model-analyzer report --report-model-configs "
+                    f"{','.join(top_3_model_config_names)} -e "
+                    f"{self._config.export_path}` to generate detailed reports "
+                    f"for the {len(top_3_model_config_names)} best "
+                    f"configurations.")
 
     def report(self, mode):
         """
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -119,17 +119,13 @@ class Analyzer:
         profiled_model_list = list(
             self._state_manager.get_state_variable(
                 'ResultManager.results').keys())
+        num_profiled_configs = self._get_num_profiled_configs()
 
-        num_profiled_configs = sum([
-            len(x) for x in self._state_manager.get_state_variable(
-                'ResultManager.results').values()
-        ])
-        logger.info(
-            f"Profile complete. Profiled {num_profiled_configs} configurations "
-            f"for models: {profiled_model_list}. To analyze the profile "
-            f"results and find the best configurations, please run "
-            f"`model-analyzer analyze --analysis-models "
-            f"{','.join(profiled_model_list)}`.")
+        logger.info(f"Profile complete. Profiled {num_profiled_configs} "
+                    f"configurations for models: {profiled_model_list}.")
+        logger.info(f"To analyze the profile results and find the best "
+                    f"configurations, please run `model-analyzer analyze "
+                    f"--analysis-models {','.join(profiled_model_list)}`.")
 
     def analyze(self, mode, quiet):
         """
@@ -173,15 +169,12 @@ class Analyzer:
         if not quiet:
             self._result_manager.write_results()
 
-        top_3_model_config_names = [
-            x.model_config().get_config()['name']
-            for x in self._result_manager.top_n_results(n=3)
-        ]
-        logger.info(f"Run `model-analyzer report --report-model-configs "
+        top_3_model_config_names = self._get_top_3_model_config_names()
+        logger.info(f"To generate detailed reports for the "
+                    f"{len(top_3_model_config_names)} best configurations, run "
+                    f"`model-analyzer report --report-model-configs "
                     f"{','.join(top_3_model_config_names)} -e "
-                    f"{self._config.export_path}` to generate detailed reports "
-                    f"for the {len(top_3_model_config_names)} best "
-                    f"configurations.")
+                    f"{self._config.export_path}`")
 
     def report(self, mode):
         """
@@ -212,3 +205,15 @@ class Analyzer:
 
         self._report_manager.create_detailed_reports()
         self._report_manager.export_detailed_reports()
+
+    def _get_num_profiled_configs(self):
+        return sum([
+            len(x) for x in self._state_manager.get_state_variable(
+                'ResultManager.results').values()
+        ])
+
+    def _get_top_3_model_config_names(self):
+        return [
+            x.model_config().get_config()['name']
+            for x in self._result_manager.top_n_results(n=3)
+        ]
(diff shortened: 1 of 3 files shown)


matthewkotila:
will add testing soon

model_analyzer/analyzer.py
```diff
@@ -119,9 +119,17 @@ def profile(self, client, gpus):
...
                 'ResultManager.results').keys())
+
+        num_profiled_configs = sum([
+            len(x) for x in self._state_manager.get_state_variable(
+                'ResultManager.results').values()
+        ])
         logger.info(
-            f"Finished profiling. Obtained measurements for models: {profiled_model_list}."
-        )
+            f"Profile complete. Profiled {num_profiled_configs} configurations "
```
matthewkotila, line 128: example:

```
...
2021-11-19 01:28:58.440 INFO[analyzer.py:127] Profile complete. Profiled 2 configurations for models: ['add_sub']. To analyze the profile results and find the best configurations, please run `model-analyzer analyze --analysis-models add_sub`.
$ 
```

model_analyzer/analyzer.py
```diff
@@ -165,6 +173,16 @@ def analyze(self, mode, quiet):
         if not quiet:
             self._result_manager.write_results()
 
+        top_3_model_config_names = [
+            x.model_config().get_config()['name']
+            for x in self._result_manager.top_n_results(n=3)
+        ]
+        logger.info(f"Run `model-analyzer report --report-model-configs "
```
matthewkotila, line 180: example:

```
...
2021-11-19 01:31:40.95 INFO[analyzer.py:180] Run `model-analyzer report --report-model-configs add_sub_i0,add_sub_i1 -e /opt/triton-model-analyzer` to generate detailed reports for the 2 best configurations.
$ 
```

model_analyzer/analyzer.py
tgerdesnv: That is one really long line. Can we add your changes to a new logger line?

model_analyzer/analyzer.py
tgerdesnv: Format here is different. For analyze you had  "<explaining words>, <command>".  Here you have "<command>, <explaining words>".  Lets make them consistent.  I think it should be "<explaining words>: <command>"

model_analyzer/analyzer.py
```diff
@@ -119,9 +119,17 @@ def profile(self, client, gpus):
         profiled_model_list = list(
             self._state_manager.get_state_variable(
                 'ResultManager.results').keys())
+
+        num_profiled_configs = sum([
```
tgerdesnv, line 123: I know the existing code isn't clean, but lets make sure our new code is clean.  I asked below to use a new logger line for this, so you should be able to extract it all out into a function that is called after the normal "finished profiling" print.  Same issue down below for the other case.  Once you have standalone functions you can add unit tests.

tgerdesnv (changes requested):

model_analyzer/analyzer.py
matthewkotila: broke up into multiple calls to `logger.info()` 👍

model_analyzer/analyzer.py
matthewkotila: reworded 👍

model_analyzer/analyzer.py
matthewkotila: extracted both into separate functions and added unit tests 👍

model_analyzer/analyzer.py
```diff
@@ -194,3 +205,15 @@ def report(self, mode):
...
         self._report_manager.create_detailed_reports()
         self._report_manager.export_detailed_reports()
+
+    def _get_num_profiled_configs(self):
+        return sum([
+            len(x) for x in self._state_manager.get_state_variable(
+                'ResultManager.results').values()
+        ])
+
+    def _get_top_3_model_config_names(self):
```
tgerdesnv, line 215: lets not hardcode this at 3. Pass it in

tests/test_analyzer.py
```diff
@@ -0,0 +1,116 @@
...
+
+from .mocks.mock_config import MockConfig
+
+
+class TestAnalyzer(trc.TestResultCollector):
+    """
+    Tests the methods of the Analyzer class
+    """
+
+    def _evaluate_config(self, args, yaml_content):
```
tgerdesnv, line 36: reorder the file so the tests are at the top and helper functions at the bottom

model_analyzer/analyzer.py
```diff
@@ -165,6 +169,13 @@ def analyze(self, mode, quiet):
         if not quiet:
             self._result_manager.write_results()
 
+        top_3_model_config_names = self._get_top_3_model_config_names()
+        logger.info(f"To generate detailed reports for the "
```
tgerdesnv, line 173: My expectation was that this would be logger.info(some_function_to_get_string()), and that the testing would just test the two get_string() functions.  That is really what we want to test and protect, not the lower level functions that function would use (although those get tested for free through this). You should be able to leverage most of the testing code you wrote.

model_analyzer/analyzer.py
```diff
@@ -119,9 +119,13 @@ def profile(self, client, gpus):
...
             self._state_manager.get_state_variable(
                 'ResultManager.results').keys())
-        logger.info(
-            f"Finished profiling. Obtained measurements for models: {profiled_model_list}."
-        )
+        num_profiled_configs = self._get_num_profiled_configs()
+
+        logger.info(f"Profile complete. Profiled {num_profiled_configs} "
+                    f"configurations for models: {profiled_model_list}.")
+        logger.info(f"To analyze the profile results and find the best "
```
tgerdesnv, line 126: Are there no other options that need to be passed forward from one command to the next?  What about something like --checkpoint-directory or --output-model-repository-path ? Do those matter if they are set in the first command?  And possibly --export-path for analyze->report?

tgerdesnv (changes requested):

model_analyzer/analyzer.py
matthewkotila: @Tabrizian can you comment on this? i don't think i have enough understanding of model analyzer to know

Tabrizian: Only checkpoint directory and export path are needed for the analyze phase. Maybe we can suggest a link to the Model Analyzer docs for the analyze phase, so that they can build the command by themselves?

model_analyzer/analyzer.py
matthewkotila: redesigned 👍

tests/test_analyzer.py
matthewkotila: reordered 👍

model_analyzer/analyzer.py
matthewkotila: redesigned 👍

model_analyzer/analyzer.py
matthewkotila: i added in `--config-file` and `--checkpoint-directory` for both analyze and report, if they're used 👍

report already included `--export-path` 👍

model_analyzer/analyzer.py
```diff
@@ -165,6 +163,8 @@ def analyze(self, mode, quiet):
         if not quiet:
             self._result_manager.write_results()
 
+        logger.info(self._get_report_command_help_string())
```
tgerdesnv, line 166: curious if `quiet` should be honored here?

tests/test_analyzer.py
```diff
@@ -0,0 +1,151 @@
...
+
+    @patch(
+        'model_analyzer.config.input.config_command_analyze.file_path_validator',
+        lambda _: ConfigStatus(status=CONFIG_PARSER_SUCCESS))
+    @patch(
+        'model_analyzer.config.input.config_command_analyze.ConfigCommandAnalyze._preprocess_and_verify_arguments',
+        lambda _: None)
+    @patch('model_analyzer.result.result_manager.ResultManager.top_n_results',
+           mock_top_n_results)
+    def test_get_report_command_help_string(self):
```
tgerdesnv, line 103: can this be moved to the top of the class? Or does it require that some other functions are defined first?

tgerdesnv (changes requested):

tests/test_analyzer.py
matthewkotila: https://github.com/triton-inference-server/model_analyzer/pull/250#discussion_r758081557

stuff used in the patch decorator needs to be defined before the python interpreter reaches the patch decorator statement

the stuff above this needs to be above it

model_analyzer/analyzer.py
matthewkotila: `logger.info()` already respects the quiet option:

https://github.com/triton-inference-server/model_analyzer/blob/matthewkotila-add-next-steps-to-cli-output/model_analyzer/entrypoint.py#L246-L247

https://github.com/triton-inference-server/model_analyzer/blob/matthewkotila-add-next-steps-to-cli-output/model_analyzer/entrypoint.py#L246-L247

tgerdesnv (approved):

PR #253, opened 2021-11-23
Always run default config

tgerdesnv:
Run the default config (the unmodified model_config passed to MA) on all MA profile runs.

Also updated test_model_manager tests to use a dict of expected values ranges instead of lists, and added many new tests to lock up the behavior.

### DIFF
--- a/model_analyzer/config/run/run_config_generator.py
+++ b/model_analyzer/config/run/run_config_generator.py
@@ -147,6 +147,7 @@ class RunConfigGenerator:
         specified list of configs
         """
 
+        return False  # FIXME
         found = False
         for config in configs:
             if config is None:
--- a/model_analyzer/config/run/run_config_generator.py
+++ b/model_analyzer/config/run/run_config_generator.py
@@ -137,36 +137,8 @@ class RunConfigGenerator:
 
     def generate_model_config_combinations(self, value):
         configs = self._generate_model_config_combinations_helper(value)
-        if not self._default_config_in_configs(configs):
-            self._add_default_config(configs)
+        self._add_default_config(configs)
         return configs
-
-    def _default_config_in_configs(self, configs):
-        """ 
-        Returns true if a default configuration exists within the
-        specified list of configs
-        """
-
-        return False  # FIXME
-        found = False
-        for config in configs:
-            if config is None:
-                found = True
-                break
-
-            default_dynamic_batching_found = config.get(
-                "dynamic_batching") is None
-
-            instances = None
-            instance_group = config.get("instance_group")
-            if instance_group is not None:
-                instances = instance_group[0]["count"]
-            default_instances_found = instances is None or instances == 1
-
-            if default_instances_found and default_dynamic_batching_found:
-                found = True
-                break
-        return found
 
     def _add_default_config(self, configs):
         # Add in an empty configuration, which will apply the default values
--- a/model_analyzer/config/run/run_config_generator.py
+++ b/model_analyzer/config/run/run_config_generator.py
@@ -113,6 +113,8 @@ class RunConfigGenerator:
 
             model_name_index = self._model_name_index
             model_config_dict = model_config.get_config()
+            self._apply_model_config_defaults(model_config_dict,
+                                              model_config.cpu_only())
 
             try:
                 model_name_index = self._model_configs.index(model_config_dict)
@@ -135,14 +137,33 @@ class RunConfigGenerator:
                     RunConfig(model.model_name(), model_config, perf_config,
                               model.triton_server_environment()))
 
+    def _apply_model_config_defaults(self, config_dict, cpu_only):
+        if "instance_group" not in config_dict:
+            config_dict["instance_group"] = [{}]
+        for instance_group in config_dict["instance_group"]:
+            self._apply_instance_group_defaults(instance_group, cpu_only)
+
+    def _apply_instance_group_defaults(self, instance_group, cpu_only):
+        if "count" not in instance_group:
+            instance_group["count"] = 1
+        if "kind" not in instance_group:
+            if (cpu_only):
+                instance_group["kind"] = "KIND_CPU"
+            else:
+                instance_group["kind"] = "KIND_GPU"
+
     def generate_model_config_combinations(self, value):
         configs = self._generate_model_config_combinations_helper(value)
-        self._add_default_config(configs)
+        if not self._default_config_in_configs(configs):
+            self._add_default_config(configs)
         return configs
+
+    def _default_config_in_configs(self, configs):
+        return None in configs
 
     def _add_default_config(self, configs):
         # Add in an empty configuration, which will apply the default values
-        configs.append({})
+        configs.append(None)
 
     def _generate_model_config_combinations_helper(self, value):
         """
--- a/model_analyzer/config/run/run_config_generator.py
+++ b/model_analyzer/config/run/run_config_generator.py
@@ -136,6 +136,42 @@ class RunConfigGenerator:
                               model.triton_server_environment()))
 
     def generate_model_config_combinations(self, value):
+        configs = self._generate_model_config_combinations_helper(value)
+        if not self._default_config_in_configs(configs):
+            self._add_default_config(configs)
+        return configs
+
+    def _default_config_in_configs(self, configs):
+        """ 
+        Returns true if a default configuration exists within the
+        specified list of configs
+        """
+
+        found = False
+        for config in configs:
+            if config is None:
+                found = True
+                break
+
+            default_dynamic_batching_found = config.get(
+                "dynamic_batching") is None
+
+            instances = None
+            instance_group = config.get("instance_group")
+            if instance_group is not None:
+                instances = instance_group[0]["count"]
+            default_instances_found = instances is None or instances == 1
+
+            if default_instances_found and default_dynamic_batching_found:
+                found = True
+                break
+        return found
+
+    def _add_default_config(self, configs):
+        # Add in an empty configuration, which will apply the default values
+        configs.append({})
+
+    def _generate_model_config_combinations_helper(self, value):
         """
         Generates all the alternative config fields for
         a given value.
@@ -163,7 +199,7 @@ class RunConfigGenerator:
                 # here.
                 for sweep_choice in sweep_choices:
                     sweep_parameter_list += \
-                        self.generate_model_config_combinations(
+                        self._generate_model_config_combinations_helper(
                                                     sweep_choice
                                                     )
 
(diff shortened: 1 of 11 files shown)


model_analyzer/config/run/run_config_generator.py
```diff
@@ -136,6 +136,42 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
                               model.triton_server_environment()))
 
     def generate_model_config_combinations(self, value):
+        configs = self._generate_model_config_combinations_helper(value)
+        if not self._default_config_in_configs(configs):
```
tgerdesnv, line 140: This is where I injected the new behavior.

model_analyzer/config/run/run_config_generator.py
```diff
@@ -136,6 +136,42 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
...
+                instances = instance_group[0]["count"]
+            default_instances_found = instances is None or instances == 1
+
+            if default_instances_found and default_dynamic_batching_found:
+                found = True
+                break
+        return found
+
+    def _add_default_config(self, configs):
+        # Add in an empty configuration, which will apply the default values
```
tgerdesnv, line 171: Is my understanding correct here? Empty config = default?

tests/test_model_manager.py
```diff
@@ -36,108 +35,155 @@
 from unittest.mock import patch
 
 
+class MockRunConfigs():
```
tgerdesnv, line 38: I need to comment this class

tests/test_model_manager.py
```diff
@@ -36,108 +35,155 @@
...
-        super().__init__(config, client, server, metrics_manager, result_manager, state_manager)
-        self._configs = []
-    
+    def __init__(self, config, client, server, metrics_manager, result_manager,
+                 state_manager):
+        super().__init__(config, client, server, metrics_manager,
+                         result_manager, state_manager)
+        self._configs = MockRunConfigs()
+
     def _execute_run_configs(self):
```
tgerdesnv, line 78: comment this function

model_analyzer/model_manager.py
```diff
@@ -126,27 +126,29 @@ def _run_model_with_search(self, model):
...
-                else:
-                    # Search through concurrency values only
-                    for user_model_config_sweep in user_model_config_sweeps:
-                        if self._state_manager.exiting():
-                            return
-                        self._run_model_config_sweep(
-                            model,
-                            search_model_config=False,
-                            user_model_config_sweep=user_model_config_sweep)
+            user_model_config_sweeps = \
```
tgerdesnv, line 129: This code block now always runs, regardless of if there are any model_config_parameters specified. For the case where no model parameters are specified, it will just contain a single empty entry, which will correspond to the default config.

tests/mocks/mock_config.py
```diff
@@ -29,3 +29,4 @@ def _fill_patchers(self):
         patchers.append(
             patch("builtins.open", mock_open(read_data=self.yaml_file_content)))
         patchers.append(patch("sys.argv", self.args))
+        patchers.append(patch('numba.cuda.is_available', MagicMock(True)))
```
tgerdesnv, line 32: without this, all of our test default configs would use KIND_CPU because cuda isn't available.

tests/test_model_manager.py
debermudez: should this be in its own file?

model_analyzer/config/run/run_config_generator.py
```diff
@@ -113,6 +113,8 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
 
             model_name_index = self._model_name_index
             model_config_dict = model_config.get_config()
+            self._apply_model_config_defaults(model_config_dict,
```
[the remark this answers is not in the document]
Tabrizian: >For example, an unspecified instance group was considered "different" than an instance group with count=1 kind=GPU, even though they are the same. This change fixes that problem. However, the onus shouldn't be on this class to update the default. I suppose this functionality should go into ModelConfig class?

I think depending on the backend it could be a different configuration. My suggestion is to not try to go very deep on detecting whether two configurations are really equal to each other or not. We should just try to make sure that we have always run the config provided by the user. Two configurations are only equal when they look exactly like each other and are specifying the same fields.

model_analyzer/config/run/run_config_generator.py
Tabrizian: I think it would be better to remove this one line method and have the logic in L156.

model_analyzer/config/run/run_config_generator.py
```diff
@@ -135,7 +137,35 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
...
+                instance_group["kind"] = "KIND_GPU"
+
     def generate_model_config_combinations(self, value):
+        configs = self._generate_model_config_combinations_helper(value)
+        if not self._default_config_in_configs(configs):
+            self._add_default_config(configs)
+        return configs
+
+    def _default_config_in_configs(self, configs):
+        return None in configs
```
Tabrizian, line 162: Can we move this to L156?

model_analyzer/config/run/run_config_generator.py
```diff
@@ -135,7 +137,35 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
                     RunConfig(model.model_name(), model_config, perf_config,
                               model.triton_server_environment()))
 
+    def _apply_model_config_defaults(self, config_dict, cpu_only):
+        if "instance_group" not in config_dict:
+            config_dict["instance_group"] = [{}]
+        for instance_group in config_dict["instance_group"]:
+            self._apply_instance_group_defaults(instance_group, cpu_only)
+
+    def _apply_instance_group_defaults(self, instance_group, cpu_only):
```
Tabrizian, line 146: We should run the default config as is without any modifications as it might change behavior.

tests/test_model_manager.py
```diff
@@ -36,156 +35,278 @@
...
+        return set(self._configs)
+
+    def add_config(self, config):
+        """ 
+        Add a single config from a dict
+        """
+        list_of_tuples = [
+            (y, config[y]) for y in sorted(config.keys()) if config[y] != None
+        ]
+        out_tuple = tuple(item for x in list_of_tuples for item in x)
```
debermudez, line 53: (nit) This feels like one of those things you can "technically" do but its unclear that you should. It is hard to read list comprehensions as they get more complex. I would prefer we wrote this a different way. Would this provide the functionality you need: https://docs.python.org/3/library/functions.html#zip

Tabrizian (changes requested):

tests/test_model_manager.py
```diff
@@ -233,40 +367,304 @@ def test_manual_parameters(self):
...
 
-        config = self._evaluate_config(args, yaml_content)
         state_manager = AnalyzerStateManager(config, MagicMock())
-        model_manager = ModelManagerSubclass(config, MagicMock(), MagicMock(), MagicMock(), MagicMock(), state_manager)
+        model_manager = ModelManagerSubclass(config, MagicMock(), MagicMock(),
+                                             MagicMock(), MagicMock(),
+                                             state_manager)
 
         model_manager.run_model(config.profile_models[0])
+        self.mock_model_config.stop()
```
debermudez, line 655: Why did this get take out of the tearDown function?

model_analyzer/config/run/run_config_generator.py
tgerdesnv: Done

model_analyzer/config/run/run_config_generator.py
tgerdesnv: This is intentional. It is self-documenting code and part of clean code.

tests/test_model_manager.py
tgerdesnv: Done

tests/test_model_manager.py
tgerdesnv: Done

model_analyzer/config/run/run_config_generator.py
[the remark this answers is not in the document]
tgerdesnv: Reverted

model_analyzer/config/run/run_config_generator.py
```diff
@@ -135,7 +137,35 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
                     RunConfig(model.model_name(), model_config, perf_config,
                               model.triton_server_environment()))
 
+    def _apply_model_config_defaults(self, config_dict, cpu_only):
```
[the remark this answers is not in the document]
tgerdesnv: Reverted

model_analyzer/config/run/run_config_generator.py
tgerdesnv: This is intentional and self-documenting

model_analyzer/config/run/run_config_generator.py
tgerdesnv: reverted

model_analyzer/config/run/run_config_generator.py
```diff
@@ -135,7 +137,35 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
...
+        if "instance_group" not in config_dict:
+            config_dict["instance_group"] = [{}]
+        for instance_group in config_dict["instance_group"]:
+            self._apply_instance_group_defaults(instance_group, cpu_only)
+
+    def _apply_instance_group_defaults(self, instance_group, cpu_only):
+        if "count" not in instance_group:
+            instance_group["count"] = 1
+        if "kind" not in instance_group:
+            if (cpu_only):
```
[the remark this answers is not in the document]
tgerdesnv: removed

model_analyzer/config/run/run_config_generator.py
```diff
@@ -135,7 +137,35 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
...
+        if not self._default_config_in_configs(configs):
+            self._add_default_config(configs)
+        return configs
+
+    def _default_config_in_configs(self, configs):
+        return None in configs
+
+    def _add_default_config(self, configs):
+        # Add in an empty configuration, which will apply the default values
+        configs.append(None)
```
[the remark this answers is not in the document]
tgerdesnv: These aren't true configurations. They are each a set of changes to apply to the supplied default configuration before running the model.  None just means "don't apply any changes, run the supplied configuration as-is".

tests/test_model_manager.py
tgerdesnv: Zip won't work. I kind of need the opposite of zip. I also need it to be explicitly sorted and a tuple so it can be hashed as a set and easily compared.  I have improved the documentation.

tests/test_model_manager.py
tgerdesnv: Before it was a fixed protobuf, so start and stop could be in setup and teardown. I added the ability to supply a different/custom protobuf, so I had to move the start/stop into _test_model_manager()

Tabrizian (approved):

PR #254, opened 2021-11-29
Reloading of model can now be disabled

cvinson830:
New cmdline argument added: `reload_model_disable`.  When specified, the loading and unloading of model is disabled.

### DIFF
--- a/model_analyzer/config/input/config_command_profile.py
+++ b/model_analyzer/config/input/config_command_profile.py
@@ -418,6 +418,17 @@ class ConfigCommandProfile(ConfigCommand):
                 description=
                 "Comma-delimited list of concurrency values or ranges <start:end:step>"
                 " to be used during profiling"))
+        self._add_config(
+            ConfigField(
+                'model_loading_unloading_disable',
+                field_type=ConfigPrimitive(bool),
+                parser_args={'action': 'store_true'},
+                default_value=False,
+                flags=['--model-loading-unloading-disable'],
+                description=
+                'Flag to indicate whether or not to disable model '
+                'loading and unloading in remote mode.'
+            ))
 
     def _add_client_configs(self):
         """
--- a/model_analyzer/config/input/config_command_profile.py
+++ b/model_analyzer/config/input/config_command_profile.py
@@ -420,11 +420,11 @@ class ConfigCommandProfile(ConfigCommand):
                 " to be used during profiling"))
         self._add_config(
             ConfigField(
-                'model_loading_unloading_disable',
+                'reload_model_disable',
                 field_type=ConfigPrimitive(bool),
                 parser_args={'action': 'store_true'},
                 default_value=False,
-                flags=['--model-loading-unloading-disable'],
+                flags=['--reload-model-disable'],
                 description=
                 'Flag to indicate whether or not to disable model '
                 'loading and unloading in remote mode.'
--- a/model_analyzer/config/run/run_config_generator.py
+++ b/model_analyzer/config/run/run_config_generator.py
@@ -87,7 +87,7 @@ class RunConfigGenerator:
         num_retries = analyzer_config['client_max_retries']
 
         if analyzer_config['triton_launch_mode'] == 'remote':
-            reload_model  = not analyzer_config['reload_model_disable']
+            reload_model = not analyzer_config['reload_model_disable']
             if reload_model:
                 self._client.load_model(model.model_name())
             model_config = ModelConfig.create_from_triton_api(
--- a/model_analyzer/config/run/run_config_generator.py
+++ b/model_analyzer/config/run/run_config_generator.py
@@ -87,12 +87,12 @@ class RunConfigGenerator:
         num_retries = analyzer_config['client_max_retries']
 
         if analyzer_config['triton_launch_mode'] == 'remote':
-            load_and_unload_model  = not analyzer_config['model_loading_unloading_disable']
-            if load_and_unload_model:
+            reload_model  = not analyzer_config['reload_model_disable']
+            if reload_model:
                 self._client.load_model(model.model_name())
             model_config = ModelConfig.create_from_triton_api(
                 self._client, model.model_name(), num_retries)
-            if load_and_unload_model:
+            if reload_model:
                 self._client.unload_model(model.model_name())
             model_config.set_cpu_only(model.cpu_only())
             perf_configs = self._generate_perf_config_for_model(
--- a/model_analyzer/config/run/run_config_generator.py
+++ b/model_analyzer/config/run/run_config_generator.py
@@ -87,8 +87,13 @@ class RunConfigGenerator:
         num_retries = analyzer_config['client_max_retries']
 
         if analyzer_config['triton_launch_mode'] == 'remote':
+            load_and_unload_model  = not analyzer_config['model_loading_unloading_disable']
+            if load_and_unload_model:
+                self._client.load_model(model.model_name())
             model_config = ModelConfig.create_from_triton_api(
                 self._client, model.model_name(), num_retries)
+            if load_and_unload_model:
+                self._client.unload_model(model.model_name())
             model_config.set_cpu_only(model.cpu_only())
             perf_configs = self._generate_perf_config_for_model(
                 model.model_name(), model,
--- a/model_analyzer/triton/client/grpc_client.py
+++ b/model_analyzer/triton/client/grpc_client.py
@@ -50,9 +50,9 @@ class TritonGRPCClient(TritonClient):
             A dictionary containg the model config.
         """
 
-        self.load_model(model_name)
+        # self.load_model(model_name)
         self.wait_for_model_ready(model_name, num_retries)
         model_config_dict = self._client.get_model_config(model_name,
                                                           as_json=True)
-        self.unload_model(model_name)
+        # self.unload_model(model_name)
         return model_config_dict['config']
--- a/model_analyzer/triton/client/grpc_client.py
+++ b/model_analyzer/triton/client/grpc_client.py
@@ -50,9 +50,8 @@ class TritonGRPCClient(TritonClient):
             A dictionary containg the model config.
         """
 
-        # self.load_model(model_name)
+        self.load_model(model_name)
         self.wait_for_model_ready(model_name, num_retries)
         model_config_dict = self._client.get_model_config(model_name,
                                                           as_json=True)
-        # self.unload_model(model_name)
         return model_config_dict['config']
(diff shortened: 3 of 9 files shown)


model_analyzer/triton/client/grpc_client.py
```diff
@@ -54,5 +54,4 @@ def get_model_config(self, model_name, num_retries):
         self.wait_for_model_ready(model_name, num_retries)
         model_config_dict = self._client.get_model_config(model_name,
                                                           as_json=True)
-        self.unload_model(model_name)
```
Tabrizian, line 57: Can we add an option to only disable it when the option is set to True? (e.g. the option could be named `allow_model_unloading`) Also, I think as we discussed this requires removing `unload_model` and `load_model` from `get_model_config`.

Tabrizian (changes requested):

model_analyzer/triton/client/grpc_client.py
tgerdesnv: `allow_model_unloading` would require the user to set it to false?  I think we want the default to be false for this new option, and if the user wants different behavior they set an option to true.  I think it should be called `model-unloading-disable` (I think disable-model-unloading is better, but for consistency with other options we should put the disable at the end of the name)

tgerdesnv (changes requested):

model_analyzer/triton/client/grpc_client.py
Tabrizian: `model-unloading-disable` sounds good to me. We need to also document that this option is only used during `remote` mode.

cvinson830: @Tabrizian @tgerdesnv To handle both `unloading` and `loading`, how about renaming this to `model-loading-unloading-disable`?

cvinson830: Red flag: Using this negative/not/disable flag name will yield code like this.
@tgerdesnv See the `not not` below?  I think we should reconsider the name of this flag.  It's a tradeoff between consistent (albeit bad) argument names, and clean code.  

```
            not_loading_and_unloading = analyzer_config['model_loading_unloading_disable']

            if not not_loading_and_unloading:
                self._client.load_model(model.model_name())
```

tgerdesnv: I guess I need more context for this story/change. Surely we still have to initially load a model, right?  Is this story going to remove the **reloading** of the model?

model_analyzer/config/input/config_command_profile.py
```diff
@@ -418,6 +418,17 @@ def _add_profile_models_configs(self):
                 description=
                 "Comma-delimited list of concurrency values or ranges <start:end:step>"
                 " to be used during profiling"))
+        self._add_config(
+            ConfigField(
+                'model_loading_unloading_disable',
```
Tabrizian, line 423: Can we rename this to `model_reloading_disable` as @tgerdesnv suggested?

Tabrizian (changes requested):
Can you add a test for this in L0_server_launch_modes? We should be able to launch Triton without `--explicit-model-control` and this flag set to true and observe similar results.

model_analyzer/config/input/config_command_profile.py
cvinson830: At this reading, the suggestion was actually `model-reload-disable`.

In light of existing arguments being verb_noun_disable, how about `reload_model_disable`?

tgerdesnv: Sounds good to me! 👍

Tabrizian: Sounds good to me too.

qa/L0_server_launch_modes/test.sh
```diff
@@ -109,41 +102,103 @@ function run_server_launch_modes() {
...
+
+function _run_server() {
+    if [ "$LAUNCH_MODE" != "remote" ]; then 
+        echo -e "\n***\n*** Test Failed. Function '_run_server' can only be run in remote mode (not '${LAUNCH_MODE}' mode). \n***"
+        exit 1
+    fi
+
+    # For remote launch, set server args and start server
+    SERVER=`which tritonserver`
+    SERVER_ARGS="--model-repository=$MODEL_REPOSITORY $MODEL_CONTROL_MODE --http-port $http_port --grpc-port $grpc_port --metrics-port $metrics_port"
```
cvinson830, line 148: Notice how `--model-control-mode=explicit` is replaced with `$MODEL_CONTROL_MODE`

qa/L0_server_launch_modes/test.sh
```diff
@@ -109,41 +102,103 @@ function run_server_launch_modes() {
...
+    done
+}
+
+function run_model_modes() {
+    if [ "$LAUNCH_MODE" != "remote" ]; then 
+        echo -e "\n***\n*** Test Failed. Function 'run_model_modes' can only be run in remote mode (not '${LAUNCH_MODE}' mode). \n***"
+        exit 1
+    fi
+
+    model_mode_combos=('--model-control-mode=explicit,' ',--reload-model-disable')
```
cvinson830, line 119: The server argument is the first element, the MA argument is the second.
```
[ server-arg-1, ma-arg1 ],
[ server-arg-2, ma-arg2 ],
```

qa/L0_server_launch_modes/test.sh
```diff
@@ -85,20 +85,13 @@ function run_server_launch_modes() {
 
         MODEL_ANALYZER_GLOBAL_OPTIONS="-v"
         MODEL_ANALYZER_ARGS="$MODEL_ANALYZER_BASE_ARGS `convert_gpu_array_to_flag ${gpus[@]}` -f $CONFIG_FILE"
+        MODEL_CONTROL_MODE='--explicit-model-control=explicit'
+        RELOAD_MODEL_DISABLE=''
 
         # Set arguments for various launch modes
-        if [ "$LAUNCH_MODE" == "remote" ]; then
```
cvinson830, line 90: Abstract this section into it's own function `run_model_modes()`

qa/L0_server_launch_modes/test.sh
```diff
@@ -109,41 +102,103 @@ function run_server_launch_modes() {
...
+    if [ "$SERVER_PID" == "0" ]; then
+        echo -e "\n***\n*** Failed to start $SERVER\n***"
+        cat $SERVER_LOG
+        exit 1
+    fi
+}
+
+function _run_analyzer() {
+    MODEL_ANALYZER_SUBCOMMAND="profile"
+    MODEL_ANALYZER_ARGS="$MODEL_ANALYZER_BASE_ARGS $RELOAD_MODEL_DISABLE"
```
cvinson830, line 161: Append the new `RELOAD_MODEL_DISABLE` argument here

tgerdesnv (dismissed):

model_analyzer/config/run/run_config_generator.py
```diff
@@ -87,8 +87,13 @@ def generate_run_config_for_model_sweep(self, model, model_sweep):
         num_retries = analyzer_config['client_max_retries']
 
         if analyzer_config['triton_launch_mode'] == 'remote':
+            reload_model  = not analyzer_config['reload_model_disable']
```
Tabrizian, line 90: Looks like there is an extra space after `reload_model`.  Did you run formatter?

qa/L0_server_launch_modes/test.sh
```diff
@@ -85,65 +85,147 @@ function run_server_launch_modes() {
...
+        MA_EXPECTED_RESULT='EP'     # Expected Pass
+
+        _do_config
+        if [ $? -ne 0 ]; then
+            break
+        fi
+    done
+}
+
+function _do_config() {
```
Tabrizian, line 99: Could use a more descriptive name? Maybe `_run_single_config` instead?

qa/L0_server_launch_modes/test.sh
```diff
@@ -85,65 +85,147 @@ function run_server_launch_modes() {
...
+        MODEL_ANALYZER_ARGS="$MODEL_ANALYZER_ARGS --triton-output-path=${SERVER_LOG}"
+    fi
+
+    _do_analyzer
+    if [ $? -ne 0 ]; then
+        return 1
+    fi        
+}
+
+function _do_analyzer() {
```
Tabrizian, line 169: nit: How about `_run_analyzer_and_check_results` instead?

qa/L0_server_launch_modes/test.sh
```diff
@@ -85,65 +85,147 @@ function run_server_launch_modes() {
...
+function _do_config() {
+    # Set arguments for various launch modes
+    if [ "$LAUNCH_MODE" == "remote" ]; then    
+
+        # EP = "expected pass"
+        # EF = "expected fail"
+        model_mode_combos=(
+            '--model-control-mode=explicit;;EP'
+            ';--reload-model-disable;EP'
+            '--model-control-mode=explicit;--reload-model-disable;EF'
```
Tabrizian, line 108: Is the MA status code non-zero in this case? It is curious that the status code for the case below this is zero but for this case the status code is non-zero.

Tabrizian (commented):
Some minor comments. The rest looks good to me.

model_analyzer/config/run/run_config_generator.py
tgerdesnv: I'm pretty sure it is set up to run the formatter when you commit, but maybe that is something I configured locally?

tgerdesnv: Looks like the pre-commit hook it has to manually be enabled per person. Chris, see "Code Formatting" in the onboarding doc.

Tabrizian: Yes, running the formatter automatically before commit needs local configuration which I think you did.

cvinson830: Done

qa/L0_server_launch_modes/test.sh
cvinson830: Done

qa/L0_server_launch_modes/test.sh
cvinson830: Done

qa/L0_server_launch_modes/test.sh
cvinson830: yes https://gitlab-master.nvidia.com/dl/dgx/tritonmodelanalyzer/-/jobs/30549955#L6446

Tabrizian (approved):

PR #257, opened 2021-11-30
Update perf_analyzer_args to match perf_analyzer's actual set of args

tgerdesnv:
Update PA flags in MA to match actual PA flags

### DIFF
--- a/model_analyzer/perf_analyzer/perf_config.py
+++ b/model_analyzer/perf_analyzer/perf_config.py
@@ -26,12 +26,13 @@ class PerfAnalyzerConfig:
         'service-kind', 'model-signature-name', 'async', 'sync',
         'measurement-interval', 'concurrency-range', 'request-rate-range',
         'request-distribution', 'request-intervals', 'binary-search',
-        'num-of-sequence', 'latency-threshold', 'max-threads',
+        'num-of-sequences', 'latency-threshold', 'max-threads',
         'stability-percentage', 'max-trials', 'percentile', 'input-data',
         'shared-memory', 'output-shared-memory-size', 'shape',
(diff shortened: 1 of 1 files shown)


model_analyzer/perf_analyzer/perf_config.py
```diff
@@ -26,12 +26,13 @@ class PerfAnalyzerConfig:
...
         'measurement-interval', 'concurrency-range', 'request-rate-range',
         'request-distribution', 'request-intervals', 'binary-search',
-        'num-of-sequence', 'latency-threshold', 'max-threads',
+        'num-of-sequences', 'latency-threshold', 'max-threads',
         'stability-percentage', 'max-trials', 'percentile', 'input-data',
         'shared-memory', 'output-shared-memory-size', 'shape',
         'sequence-length', 'string-length', 'string-data', 'measurement-mode',
         'measurement-request-count', 'streaming', 'grpc-compression-algorithm',
-        'triton-server-directory', 'model-repository'
+        'triton-server-directory', 'model-repository', 'data-directory',
```
jbkyang-nvi, line 34: nit: maybe add `sequence_id_range` after `sequence-length` so the order stays the same?

jbkyang-nvi (commented):
I was reading the code more and I noticed there seems to be many PA configuration options that MA is not using right now. I think it's fine to add more options but they might not be used.

model_analyzer/perf_analyzer/perf_config.py
jbkyang-nvi: Not nit: `--data-directory` seems like a deprecated option. We probably don't want to add it

jbkyang-nvi (approved):

PR #258, opened 2021-12-01
Add unit test for relative model path symlink creation

matthewkotila:
Adds unit test for work previously done in #256

### DIFF
--- a/tests/test_model_config.py
+++ b/tests/test_model_config.py
@@ -132,12 +132,7 @@ instance_group [
         return ['1', 'config.pbtxt', 'output0_labels.txt']
 
     def mock_os_symlink(src, dst):
-        if src == '../model_i0/1':
-            assert dst == './output_model_repository/model_i1/1'
-        elif src == '../model_i0/output0_labels.txt':
-            assert dst == './output_model_repository/model_i1/output0_labels.txt'
-        else:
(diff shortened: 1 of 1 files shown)


tests/test_model_config.py
```diff
@@ -127,3 +127,33 @@ def test_write_config_file(self):
                    MagicMock(return_value=True)):
             with self.assertRaises(TritonModelAnalyzerException):
                 ModelConfig.create_from_file(model_output_path)
+
+    def mock_os_listdir(path='.'):
+        return ['1', 'config.pbtxt', 'output0_labels.txt']
+
+    def mock_os_symlink(src, dst):
+        assert src in ['../model_i0/1', '../model_i0/output0_labels.txt']
```
Tabrizian, line 135: Could you please replace this assert with `assert_called_with` similar to what is done? [here ](https://github.com/triton-inference-server/model_analyzer/blob/19d7c138c879808a951295c1d76fe96538a24f02/tests/mocks/mock_client.py#L76)

matthewkotila: switched to `assert_any_call()` since it gets called two times in my test 👍

Tabrizian (dismissed):

Tabrizian (approved):

PR #267, opened 2021-12-07
Add concurrent profile option to CLI

nv-braf:
Adds a new CLI option: **--run-config-profile-models-concurrently-enable**

- By default this is false to maintain backwards compatibility

Fixed **--run-config-search-preferred-batch-size-disable** to no longer require a value (as this was being incorrectly used by the CLI)

Updated CI testing.


docs/config.md
```diff
@@ -173,6 +173,9 @@ profile_models: <comma-delimited-string-list>
 # Disables automatic config search
 [ run_config_search_disable: <bool> | default: false ]
 
+# Enables profiling models concurrently
+[ run_config_profile_models_concurrently_enable: <bool> | default: false ]
```
tgerdesnv, line 177: Do we need 'enable' for a positive option? Seems like `run_config_profile_models_concurrently` is enough

docs/quick_start.md
```diff
@@ -59,7 +59,7 @@ configuration:
 ```
 --run-config-search-max-concurrency 2 \
 --run-config-search-max-instance-count 2 \
---run-config-search-preferred-batch-size-disable true
```
tgerdesnv, line 62: Was this the conclusion of talks with Iman? That the value should be unspecified?  What was the behavior if 'True' is also specified here? I would expect an error

tests/test_run_config_generator.py
```diff
@@ -50,7 +50,9 @@ def setUp(self):
     def test_generate_model_config_combinations(self):
         args = [
             'model-analyzer', 'profile', '--model-repository', 'cli_repository',
-            '-f', 'path-to-config-file', '--profile-models', 'vgg11'
+            '-f', 'path-to-config-file', '--profile-models', 'vgg11',
+            '--run-config-profile-models-concurrently-enable',
```
tgerdesnv, line 54: Why is this test changed? It doesn't look like any behavior is changed?

tgerdesnv (changes requested):

docs/config.md
nv-braf: My personal preference is that we be explicit with enable/disable. He had the same debate in HW years back (around CREGs) and this helped to prevent issues.

docs/quick_start.md
nv-braf: Yes. Now, if a value is specified it will result in a CLI error. Before, you could put whatever you wanted: false, cat, rutabaga... and the end result would always be having the variable set to true.

tests/test_run_config_generator.py
nv-braf: I added this here to check that the CLI is being processed correctly. There's no behavior change to the test, but in the future if someone changes this CLI option it will cause the test to fail w/ a CLI parsing error.

tgerdesnv: I understand the desire to protect it, but I really don't like it here. It is very confusing to anyone looking at this test, trying to understand what effect these options have on it.  The testing should really go into a CLI testing area, which it appears we don't currently have. I have created a ticket for that. For now, we should revert this file.  Later, once the feature is hooked up and does something it will be protected by new tests.

docs/quick_start.md
tgerdesnv: ok, good fix. I created a ticket to improve our CLI protection

docs/config.md
tgerdesnv: It should be consistent with the rest of the options. It looks like the other run_config* options are disable/enable, so this is fine to keep.

tests/test_run_config_generator.py
nv-braf: Understood and reverted.

PR #273, opened 2021-12-17
Fix missing config directory and incorrect checkpoint duplicate confi…

matthewkotila:
…g detection bugs

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -152,7 +152,8 @@ profile_models:
 ```
 
 **Note**: The checkpoint directory should be removed between consecutive runs of
-the `model-analyzer profile` command.
+the `model-analyzer profile` command. If you use the
+`--override-output-model-repository` flag, this will be done automatically.
 
 ## Subcommand: `analyze`
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -152,8 +152,7 @@ profile_models:
 ```
 
 **Note**: The checkpoint directory should be removed between consecutive runs of
-the `model-analyzer profile` command. If you use the
-`--override-output-model-repository` flag, this will be done automatically.
+the `model-analyzer profile` command.
 
 ## Subcommand: `analyze`
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -151,6 +151,9 @@ profile_models:
 
 ```
 
+**Note**: The checkpoint directory should be removed between consecutive runs of
+the `model-analyzer profile` command.
+
 ## Subcommand: `analyze`
 
 The `analyze` subcommand allows the user to create summaries and data tables
--- a/docs/quick_start.md
+++ b/docs/quick_start.md
@@ -91,8 +91,7 @@ Here is some sample output:
 ```
 
 **Note**: The checkpoint directory should be removed between consecutive runs of
-the `model-analyzer profile` command. If you use the
-`--override-output-model-repository` flag, this will be done automatically.
+the `model-analyzer profile` command.
 
 ## Generate tables and summary reports
 In order to generate tables and summary reports, use the `analyze` subcommand as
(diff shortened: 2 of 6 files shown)


model_analyzer/model_manager.py
```diff
@@ -200,6 +200,10 @@ def _execute_run_configs(self):
             # Remove one run config from the list
             run_config = self._run_config_generator.next_config()
 
+            # Create model variant
+            self._create_model_variant(original_name=run_config.model_name(),
+                                       variant_config=run_config.model_config())
+
             # If this run config was already run, do not run again, just get the measurement
             measurement = self._get_measurement_if_config_duplicate(run_config)
```
Tabrizian, line 208: I think this requires some changes here too. This is the main place where the model config variant name is being used for duplicate detection. Could you please also add a test where Model Analyzer runs the profile command twice with completely different parameters used for sweep, then the analyze step is run and the number of rows in the table should be equal to the total of measurements run in each step?

model_analyzer/model_manager.py
```diff
@@ -283,7 +292,8 @@ def _get_measurement_if_config_duplicate(self, run_config):
         # check whether perf config string is a key in result dict
         if model_name not in results:
             return False
-        if model_config_name not in results[model_name]:
+        if not self._is_config_in_results(
+                run_config.model_config()._model_config, results[model_name]):
```
Tabrizian, line 296: I think this will not be equal to True even if the model configurations are equal. The reason is that the model configuration stored in the checkpoint contains the model config variant name of the old run. For example if the current model configuration is `model_i1` and the profiled model is `model_i2` and all the other model configuration fields are equal, then these model configs will not be considered equal.

Tabrizian: Could you please add a test where the profile command is run twice and there are some intersections between the `model_config_parameters` being profiled in each stage. Then, the total number of results during the analyze step should be `measurements_in_stage1 + measurements_in_stage2 - (measurements_in_both_stages)`.

matthewkotila: we discussed offline, and we're just going to remove the checkpoint directory on a re-profile with override 👍

model_analyzer/model_manager.py
```diff
@@ -269,3 +275,44 @@ def _load_model_variant(self, variant_config):
...
+        else:
+            return None
+
+    def _is_config_in_results(self, config, model_results):
+        """
+        Returns true if `config` exists in the checkpoint `model_results`
+        """
+
+        for result in model_results.values():
+            if config == result[0]._model_config:
```
Tabrizian, line 316: Could you add a check here to make sure that result has more than one element?

docs/cli.md
```diff
@@ -151,6 +151,9 @@ profile_models:
 
 ```
 
+**Note**: The checkpoint directory should be removed between consecutive runs of
+the `model-analyzer profile` command.
```
Tabrizian, line 155: Please add this in the end. "If you use --override-output-model-repository flag, this will be done automatically".

docs/quick_start.md
```diff
@@ -90,6 +90,9 @@ Here is some sample output:
 ...
 ```
 
+**Note**: The checkpoint directory should be removed between consecutive runs of
+the `model-analyzer profile` command.
```
Tabrizian, line 94: Similar to the comment above.

model_analyzer/model_manager.py
matthewkotila: discussed offline--how it currently is is fine 👍

docs/cli.md
matthewkotila: added 👍

docs/quick_start.md
matthewkotila: added 👍

Tabrizian (approved):

PR #274, opened 2021-12-20
move MM decisions to RCG

tgerdesnv:


model_analyzer/config/run/run_config_generator.py
```diff
@@ -36,21 +41,114 @@ def __init__(self, config, client):
...
+            logger.info(
+                f"Running auto config search for model: {model.model_name()}")
+            self._run_model_with_search(model)
+
+    def is_done(self):
+        """ Return true if all RunConfigs for the model have been returned """
+        return len(self._run_configs) == 0
+
+    def add_measurement(self, measurement):
+        """ Gives the results from the last RunConfig """
```
debermudez, line 68: This comment is incorrect now. This method isnt returning anything.

debermudez (changes requested):

model_analyzer/config/run/run_config_generator.py
tgerdesnv: The comment isn't incorrect, I just think I poorly worded it.  It is given the results of the last RunConfig.  I'll clean it up. Thanks for reviewing!

PR #278, opened 2021-12-21
move running code to metrics manager

tgerdesnv:


tests/test_model_manager.py
```diff
@@ -34,19 +35,19 @@
 from unittest.mock import patch
 
 
-class ModelManagerSubclass(ModelManager):
+class MetricsManagerSubclass(MetricsManager):
```
debermudez, line 38: Do we need to rename this file too? Seems like this might be hard to pick up later and follow if you didnt review this commit or do the change.

debermudez (approved):

tests/test_model_manager.py
tgerdesnv: This file (for now) still tests ModelManager. It is just that the function I need to overload to grab the RunConfigs that are 'run' is now in MetricsManager.

PR #280, opened 2022-01-04
Fixing errors in unit tests when running from vscode

nv-braf:
This is the issue:
File "/usr/lib/python3.8/gettext.py", line 411, in _parse
    magic = unpack('<I', buf[:4])[0]

It expects the buffer to be of type byte and a minimum length of 4. Still not sure why this only complains from vscode, but it's simple enough to fix:

- create_yaml() to convert strings and append blanks (if too short)

I have observed all unit tests passing and interactive debug is now working in vscode.

### DIFF
--- a/tests/common/test_utils.py
+++ b/tests/common/test_utils.py
@@ -18,7 +18,7 @@ from model_analyzer.record.metrics_manager import MetricsManager
 from model_analyzer.perf_analyzer.perf_config import PerfAnalyzerConfig
 
 
-def create_yaml(string):
+def convert_to_bytes(string):
     """
     Converts string into bytes and ensures minimum length requirement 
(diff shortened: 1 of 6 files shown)


tests/common/test_utils.py
```diff
@@ -18,6 +18,17 @@
 from model_analyzer.perf_analyzer.perf_config import PerfAnalyzerConfig
 
 
+def create_yaml(string):
```
debermudez, line 21: This method isnt creating anything related to yaml. It is converting a string to bytes. 
Can we rename it to something like convert_to_bytes()?

nv-braf: Done!

tgerdesnv:
sweeeeet! good find

tgerdesnv (approved):

PR #281, opened 2022-01-05
Rename output directories to use config instead of i

debermudez:
Changing the output directory naming convention to remove the ambiguous i.


nv-braf (commented):
My preference is to have an underscore after the 'config': <model_name>__config_<index>. Thoughts?

tgerdesnv:
> My preference is to have an underscore after the 'config': <model_name>__config_. Thoughts?

I think it is fine how it is.

tgerdesnv (approved):

Issue #279, opened 2021-12-23
handle server_config unexceptedly when cli parameter contains "="


AmazingSkyLine:
Hi~guys
I have run model_analyzer in local mode, and I want to run tritonserver using tensorflow backend version 2, so I set server_config as below:
```yaml
    triton_server_flags:
      backend_config: tensorflow,version=2
      strict_model_config: false
```
But when I execute model-analyzer, tritonserver has error in log:
`--backend-config option format is '<backend name>,<setting>=<value>'. Got tensorflow,version`

And I see source code in server_local.py:

```python
        if self._server_path:
            # Create command list and run subprocess
            cmd = [self._server_path]
            cmd += self._server_config.to_cli_string().replace('=', ' ').split()
```
There just replace '=' with ''

So how can I set a server_config parameter with '='?

Tabrizian:
@AmazingSkyLine Thanks for reporting this issue. I have filed a ticket to fix this.

cvinson830:
Thank you for the report.  Problem confirmed.

cvinson830:
@AmazingSkyLine  Fixed and merged.

AmazingSkyLine:
thanks

PR #283, opened 2022-01-05
Fix argument parsing

cvinson830:
Fixes issue: https://github.com/triton-inference-server/model_analyzer/issues/279
Feature: https://github.com/triton-inference-server/tensorflow_backend#--backend-configtensorflowversionint
Exercises this server code: https://github.com/triton-inference-server/server/blob/main/src/servers/main.cc#L1132-L1159

Problem: The MA code was not handling properly the case where TWO equals signs exist
(eg, `--backend-config=tensorflow,version=2`)
Before, the code accidentally split on both equals signs:
```
['tritonserver',  '--backend-config', 'tensorflow,version', '2']
```
The fix is to 
1) split on the argument list by spaces
2) Then for each of those arguments, split on the `first` equals sign
After the fix, the code splits on only the first equals sign:
```
['tritonserver', '--backend-config', 'tensorflow,version=2']
```
Notice how `version=2` is now correctly part of the `--backend_config` value

### DIFF
--- a/model_analyzer/triton/server/server_config.py
+++ b/model_analyzer/triton/server/server_config.py
@@ -142,11 +142,11 @@ class TritonServerConfig:
             output: ['--model-control-mode', 'explicit', 
                 '--backend-config', 'tensorflow,version=2']
         """
-        list = []
+        args_list = []
         args = self.to_cli_string().split()
         for arg in args:
-            list += arg.split('=', 1)
-        return list
+            args_list += arg.split('=', 1)
+        return args_list
 
     def copy(self):
         """
--- a/model_analyzer/triton/server/server_config.py
+++ b/model_analyzer/triton/server/server_config.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020-2022, NVIDIA CORPORATION. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


Tabrizian (changes requested):
Thanks for root causing this. Is it possible to add a method to https://github.com/triton-inference-server/model_analyzer/blob/main/model_analyzer/triton/server/server_config.py named `to_args_list` and then it returns a list that can be added to `cmd`?

I think in the current fix we may still have issues if the argument/value has spaces in it. For example:
```
triton_server_flags:
  backend_config: "my parameter"
```

The cmd in this case will look like: ['tritonserver', '--backend-config', '"my', 'parameter"'] instead of ['tritonserver', '--backend-config', '"my parameter"'].

cvinson830:
Regarding your comment above: there are two issues there.  Let's break them down.

1) Abstract functionality into function: Good idea. Done.
2) Handle embedded spaces within the value:  I agree that this is needed, but 
-- outside of the scope of this PR
-- It's not a "quick fix" since there are many ways to implement this (regex, csv, pyparsing, re to name four) 
-- it's always existed (it's not a result of the fix to this PR)
-- We need to make triton server can handle it.  

Let's decouple the second issue from this PR.  OK with you?

model_analyzer/triton/server/server_config.py
```diff
@@ -123,6 +123,31 @@ def to_cli_string(self):
...
+            the tritonserver.
+
+            Example:
+            input cli_string: "--model-control-mode=explicit 
+                --backend-config=tensorflow,version=2"
+
+            output: ['--model-control-mode', 'explicit', 
+                '--backend-config', 'tensorflow,version=2']
+        """
+        list = []
```
Tabrizian, line 145: Could you please only rename this `list` variable to `args_list`? `list` is already defined and might create weird issues. No need to rerun the CI after the variable change. Thank you.

cvinson830: Done.

Tabrizian (approved):

PR #285, opened 2022-01-07
Model analyzer skips configuration when perf_output is True and perf …

matthewkotila:
…analyzer has no output instead of stopping

### DIFF
--- a/model_analyzer/record/metrics_manager.py
+++ b/model_analyzer/record/metrics_manager.py
@@ -307,12 +307,14 @@ class MetricsManager:
 
         status = perf_analyzer.run(self._perf_metrics, env=perf_analyzer_env)
 
-        if perf_output_writer and perf_analyzer.output():
+        if perf_output_writer:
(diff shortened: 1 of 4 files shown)


model_analyzer/record/metrics_manager.py
```diff
@@ -307,7 +307,7 @@ def _get_perf_analyzer_metrics(self,
 
         status = perf_analyzer.run(self._perf_metrics, env=perf_analyzer_env)
 
-        if perf_output_writer:
+        if perf_output_writer and perf_analyzer.output():
```
Tabrizian, line 310: I think it would still good to write the perf analyzer command to the output so that users can use it later to debug the issue.

matthewkotila: moved this condition lower 👍

Tabrizian (approved):

jbkyang-nvi (approved):

PR #287, opened 2022-01-10
new refactor class ModelConfigGenerator

tgerdesnv:


debermudez (approved):

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -0,0 +1,79 @@
...
+        else:
+            configs = self._automatic_search_configs()
+
+        return configs
+
+    def _automatic_search_configs(self):
+        """ Search through automatic search variables to generate configs """
+
+        configs = []
+        if not self._search_disabled:
```
tgerdesnv, line 67: disabled likely still needs dynamic batching on

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -0,0 +1,79 @@
...
+            configs = self._get_local_mode_model_configs()
+
+        if not self._is_default_config_in_configs(configs):
+            self._add_default_config(configs)
+
+        return configs
+
+    def _get_remote_mode_model_configs(self):
+        """ Generate model configs for remote mode """
+        return []
```
tgerdesnv, line 50: brian says this may be incorrect

nv-braf, line 50: In the original code we are loading a config from the Triton API.

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _get_local_mode_model_configs(self):
+        """ Generate model configs for local mode """
+
+        model_config_params = self._model.model_config_parameters()
+        if model_config_params:
```
nv-braf, line 56: I don't understand how this is equivalent to what exists at ToT.

model_analyzer/config/generate/model_config_generator.py
nv-braf, line 67: It seems weird to me that we're checking to see if search is disabled from inside the auto_search method. It would be more clear if this was moved up a level and that _auto_search is only called if search is enabled

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -0,0 +1,79 @@
...
+            configs = self._automatic_search_configs()
+
+        return configs
+
+    def _automatic_search_configs(self):
+        """ Search through automatic search variables to generate configs """
+
+        configs = []
+        if not self._search_disabled:
+            for instances in range(1, self._max_instance_count + 1):
```
nv-braf, line 68: Is this new functionality? I don't see this in the original code

model_analyzer/config/generate/model_config_generator.py
tgerdesnv: updated

model_analyzer/config/generate/model_config_generator.py
tgerdesnv: I think we agreed this is fine in our chat

model_analyzer/config/generate/model_config_generator.py
tgerdesnv: I believe this is the onus of one level above (the run_config_generator) in our new scheme.  It is in the function "generate_run_config_for_model_sweep" which combines 3 things into a run_config:  The existing/default config, the model_config changes to apply on top of the existing/default config, and the perf_config.  That is what the new RCG will handle.

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -0,0 +1,79 @@
...
+        return self._configs.pop(0)
+
+    def _generate_model_configs(self):
+        """ Generate all model config combinations """
+        if self._remote_mode:
+            configs = self._get_remote_mode_model_configs()
+        else:
+            configs = self._get_local_mode_model_configs()
+
+        if not self._is_default_config_in_configs(configs):
```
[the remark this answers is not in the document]
tgerdesnv: I think we are on the same page now with this. Here is the experiment I said I would run to verify the behavior:

`
>>> x = []
>>> None in x
False
>>> x = [None]
>>> None in x
True
`

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -0,0 +1,79 @@
...
+    def next_config(self):
+        """ Returns the next generated config """
+        return self._configs.pop(0)
+
+    def _generate_model_configs(self):
+        """ Generate all model config combinations """
+        if self._remote_mode:
+            configs = self._get_remote_mode_model_configs()
+        else:
+            configs = self._get_local_mode_model_configs()
```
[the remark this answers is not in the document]
tgerdesnv: updated variables and comments

PR #289, opened 2022-01-10
Improve launch mode docs

matthewkotila:
The ask was to add a table to the launch mode docs showing tradeoffs between various model analyzer launch modes to help users understand which launch mode they may want to use.

### DIFF
--- a/docs/launch_modes.md
+++ b/docs/launch_modes.md
@@ -19,79 +19,83 @@ Triton Model Analyzer's `profile` subcommand supports four different launch
 modes along with Triton Inference Server. In the `local` and `docker` modes,
 Triton Inference Server will be launched by the Model Analyzer. In the `c_api`
 mode, the Triton Inference Server is launched locally via a C API. In the
-`remote` mode, it is assumed there is an already running instance of
-Triton Inference Server.
+`remote` mode, it is assumed there is an already running instance of Triton
+Inference Server.
 
 ### Local
 
-CLI Option: **`--triton-launch-mode=local`**
+| CLI Option | **`--triton-launch-mode=local`** |
+| - | - |
 
-In this mode, Model Analyzer will launch
-   Triton Server using the local binary supplied using `--triton-server-path`,
-   or if none is supplied, the `tritonserver` binary in `$PATH`.
+In this mode, Model Analyzer will launch Triton Server using the local binary
+supplied using `--triton-server-path`, or if none is supplied, the
+`tritonserver` binary in `$PATH`.
+
+This mode is useful if you don't want to install/use Docker. You will need
+Triton Server and all of its dependencies installed locally, though.
 
 ### Docker
 
-CLI Option: **`--triton-launch-mode=docker`**
+| CLI Option | **`--triton-launch-mode=docker`** |
+| - | - |
 
-In this mode, Model Analyzer uses the
-   Python Docker API to launch the Triton Inference Server container. If you are
-   running Model Analyzer inside a Docker container, make sure that the
-   container is launched with appropriate flags. The following flags are
-   mandatory for correct behavior:
-   ```
-   --gpus <gpus> -v /var/run/docker.sock:/var/run/docker.sock --net host
-   ```
+In this mode, Model Analyzer uses the Python Docker API to launch the Triton
+Inference Server container. If you are running Model Analyzer inside a Docker
+container, make sure that the container is launched with appropriate flags. The
+following flags are mandatory for correct behavior:
 
-   Additionally, Model Analyzer uses the `output_model_repository_path` to
-   manipulate and store model config variants. When Model Analyzer launches the
-   Triton container, it does so as a *sibling container*. The launched Triton
-   container will only have access to the host filesystem. **As a result, in the
-   docker launch mode, the output model directory will need to be mounted to the
-   Model Analyzer docker container at the same absolute path it has outside the
-   container.** So you must add the following when you launch the model analyzer
-   container as well.
+```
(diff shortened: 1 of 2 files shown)


matthewkotila:
@Tabrizian let me know what you think of the wording, particularly in [this section](https://github.com/triton-inference-server/model_analyzer/blob/matthewkotila-improve-docs/docs/launch_modes.md#tradeoffs)

docs/launch_modes.md
```diff
@@ -70,3 +86,12 @@ Triton Inference Server.
...
    loading/unloading of the models.
+
+### Tradeoffs
+
+| Mode | Pro | Con |
+| - | - | - |
+| *docker* | Don't need Triton Server or its dependencies installed locally | Need Docker installed |
+| *local* | Don't need Docker installed | Need Triton Server and all of its dependencies installed locally |
+| *c_api* | Same as local, but uses faster C API | Need Triton Server and all of its dependencies installed locally |
+| *remote* | Can use Model Analzer on already-running Triton Server | Need Triton Server running on same machine as Model Analyzer |
```
Tabrizian, line 97: Remote mode con is that you cannot use all the Model Analyzer features such as changing model config parameters. All the modes require both the Triton server and MA to be running on the same machine.

docs/launch_modes.md
```diff
@@ -70,3 +86,12 @@ Triton Inference Server.
    Analyzer does not currently support profiling remote GPUs. Triton Server in this
    mode needs to be launched with `--model-control-mode explicit` flag to support
    loading/unloading of the models.
+
+### Tradeoffs
```
Tabrizian, line 90: Instead of using "tradeoffs" which can imply lack of features, how about renaming this section to "Deciding what mode to use?" or similar? Or we could remove this section and put a sentence in the end of the description for each mode on when this mode is useful? For example:

Remote mode is useful when you have an already running instance of Triton that you cannot restart. The model config parameters cannot be changed in remote mode.

docs/launch_modes.md
```diff
@@ -22,11 +22,19 @@ mode, the Triton Inference Server is launched locally via a C API. In the
 `remote` mode, it is assumed there is an already running instance of
 Triton Inference Server.
 
-1. **`--triton-launch-mode=local`** In this mode, Model Analyzer will launch
+### Local
+
+CLI Option: **`--triton-launch-mode=local`**
```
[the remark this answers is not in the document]
matthewkotila: changed to table 👍

docs/launch_modes.md
matthewkotila: updated wording 👍

docs/launch_modes.md
matthewkotila: removed this section and put a sentence at the end of the description for each mode on when that mode is useful 👍

docs/launch_modes.md
```diff
@@ -19,54 +19,83 @@ Triton Model Analyzer's `profile` subcommand supports four different launch
...
+### Local
+
+| CLI Option | **`--triton-launch-mode=local`** |
+| - | - |
+
+In this mode, Model Analyzer will launch Triton Server using the local binary
+supplied using `--triton-server-path`, or if none is supplied, the
+`tritonserver` binary in `$PATH`.
+
+This mode is useful if you don't want to install/use Docker. You will need
```
Tabrizian, line 34: For `local` mode in the quick start don't we use Docker to create a container that has MA and Triton? How about changing this sentence to:

Local mode is the recommended method of getting started with Model Analyzer. There are detailed instructions about this mode in the [Quick Start Guide](docs/quick_start.md).

docs/launch_modes.md
```diff
@@ -19,54 +19,83 @@ Triton Model Analyzer's `profile` subcommand supports four different launch
...
+
+Finally, when launching model analyzer, the argument `--output-model-repository`
+must be provided as a directory inside `<path-to-output-model-repository>`. This
+directory need not exist. 
+
+```
+--output-model-repository=<path-to-output-model-repository>/output
+```
+
+This mode is useful if you don't want to install Triton Server and all of its
```
Tabrizian, line 72: This mode is useful if you want to use the Model Analyzer installed in the [Triton SDK Container]().

docs/launch_modes.md
```diff
@@ -19,54 +19,83 @@ Triton Model Analyzer's `profile` subcommand supports four different launch
...
+
+| CLI Option | **`--triton-launch-mode=c_api`** |
+| - | - |
+
+In this mode, Triton server is launched locally via the
+[C_API](https://github.com/triton-inference-server/server/blob/main/docs/inference_protocols.md#c-api)
+by the perf_analyzer instances launched by Model Analyzer.
+
+This mode is useful if you want to run with the Triton Server installed locally
+and want the increased performance from the C API. You will need Triton Server
```
Tabrizian, line 85: "You will need Triton Server and all of its dependencies installed locally, though." -> "Similar to the [local mode](#local-mode), Triton Server must be installed in the environment that the Model Analyzer is being used."

Tabrizian (changes requested):

docs/launch_modes.md
matthewkotila: reworded 👍

You're right. And you just made me realize that the quick start does not need to mount the docker socket. I've added a change to remove that mount from that doc.

docs/launch_modes.md
matthewkotila: reworded 👍

docs/launch_modes.md
matthewkotila: reworded 👍

Tabrizian (approved):

PR #290, opened 2022-01-10
Fix L0_stability_metrics

cvinson830:
Convert test from comparing values to simply test for non-zero values.
Due to the change in requirements for this L0 test, the `tolerance` value was removed.

Since we do not currently have a way to test our tests, this is a list of values I tested manually.

```
[0.0]       # fail
[0.000000]  # fail
[0.000001]  # pass
[1.0]       # pass
['0']       # fail
['zero']    # fail
['1']       # fail
['one']     # fail
['']        # fail
```

### DIFF
--- a/qa/L0_stability_metrics/check_results.py
+++ b/qa/L0_stability_metrics/check_results.py
@@ -72,21 +72,24 @@ class TestOutputValidator:
         for model in metric_values:
             for metric, values in metric_values[model].items():
                 start_value = values[0]
+                deviations = []
                 for value in values[1:]:
                     deviation_percent = abs(
                         (value - start_value) / start_value) * 100
+                    str_deviation = '{:.2%}'.format(deviation_percent / 100)
+                    deviations.append(str_deviation)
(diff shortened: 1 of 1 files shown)


qa/L0_stability_metrics/check_results.py
```diff
@@ -49,6 +48,7 @@ def check_metrics_stability(self):
         pathname = os.path.join(self._result_path, 'result_*.csv')
         csv_contents = []
         for filename in glob.glob(pathname):
+            print('Opening file <' + str(filename) + '>')
```
debermudez, line 51: Did you want to leave these print statements in?

cvinson830: Yes - when I ran it, and all tests passed, I did not see any confirmation that anything ran at all.  There was a part of me that thought maybe no test was performed since nothing appeared.  So, this print() and the other print() are there to confirm that files were opened and tests were performed.

debermudez (approved):

PR #292, opened 2022-01-11
Fix L0_perf_analyzer test bug

matthewkotila:
This test was previously failing because the wrong checker function from `check_results.py` was being used. Now the correct checker function is being used 👍


Tabrizian (commented):
`count_window_adjust` seems to be missing here.

tgerdesnv (approved):

matthewkotila:
> `count_window_adjust` seems to be missing here.

@Tabrizian 

No config ever used that. It would get used if a config didn't have `no-adjust` in the name and also didn't have `time` in the name, which there were no configs with a name like that.

PR #294, opened 2022-01-12
Conditionally load model variant

cvinson830:
Disable loading of a model variant if 

- in `remote launch mode` and 
- not `reload_model_disable`

### DIFF
--- a/model_analyzer/model_manager.py
+++ b/model_analyzer/model_manager.py
@@ -264,15 +264,15 @@ class ModelManager:
 
     def _load_model_variant(self, variant_config):
         """
-        Decides on whether or not to load a model variant in the client
+        Conditionally loads a model variant in the client
         """
+        remote = self._config.triton_launch_mode == 'remote'
+        c_api = self._config.triton_launch_mode == 'c_api'
+        disabled = self._config.reload_model_disable
+
+        do_load = (remote and not disabled) or (not remote and not c_api)
         retval = True
-        if self._config.triton_launch_mode == 'remote':
-            # Don't load model if in remote and reload disabled
-            reload_model = not self._config.reload_model_disable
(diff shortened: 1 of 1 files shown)


model_analyzer/model_manager.py
```diff
@@ -263,21 +263,33 @@ def _create_model_variant(self, original_name, variant_config):
                 pass
 
     def _load_model_variant(self, variant_config):
+        """
+        Decides on whether or not to load a model variant in the client
```
tgerdesnv, line 267: This implies that it only decides IF it should be loaded. It does however still do loading in some (most) cases.  Maybe `conditionally loads a model variant in the client`?

model_analyzer/model_manager.py
```diff
@@ -263,21 +263,33 @@ def _create_model_variant(self, original_name, variant_config):
...
+        """
+        retval = True
+        if self._config.triton_launch_mode == 'remote':
+            # Don't load model if in remote and reload disabled
+            reload_model = not self._config.reload_model_disable
+            if reload_model:
+                retval = self._do_load_model_variant(variant_config)
```
tgerdesnv, line 274: I think this would read cleaner if there was only a single call to `_do_load_model_variant`.  Make a local boolean `do_load` based on launch_mode and disable, and then guard the single call with that bool.

tgerdesnv (changes requested):

model_analyzer/model_manager.py
cvinson830: Nice.  Done.

model_analyzer/model_manager.py
cvinson830: Done.

tgerdesnv (approved):

PR #297, opened 2022-01-13
New run config generator

tgerdesnv:
Apply interface to all 3 classes
Fixes and cleanup
Apply cpu_only and variant_name inside of model_config_generator
No new testing, but hooked up new RCG to model_manager, and thus the full sweep of MM tests use RCG and are passing (except for the 2 expected to fail)


nv-braf (commented):
It's so much easier to understand now! This is great.

PR #302, opened 2022-01-19
Update PA CLI to support multiple models

tgerdesnv:


tests/common/test_utils.py
```diff
@@ -73,7 +73,7 @@ def construct_measurement(model_name, gpu_metric_values, non_gpu_metric_values,
 
     # Perf Config needs a protocol
     perf_config = PerfAnalyzerConfig()
-    perf_config['model-name'] = model_name
+    perf_config['model-names'] = model_name
```
matthewkotila, line 76: can `construct_measurement()` also take a comma-separated model name list string, now? instead of just a single model name string?

tests/test_perf_analyzer_config_generator.py
```diff
@@ -314,6 +314,31 @@ def test_perf_analyzer_flags(self):
...
+        pacg = PerfAnalyzerConfigGenerator(
+            config, [model.model_name() for model in config.profile_models],
+            config.profile_models[0].perf_analyzer_flags(),
+            config.profile_models[0].parameters())
+
+        perf_analyzer_configs = []
+        while not pacg.is_done():
+            perf_analyzer_configs.append(pacg.next_config())
+
+        self.assertGreater(len(perf_analyzer_configs), 0)
```
matthewkotila, line 337: while some of these `self.assert<SOMETHING>()` are necessary and/or convenient, i generally think it's more readable to just use `self.assertTrue()` and put the condition inside as regular python

nv-braf (commented):
Changes look good to me (other than Matt's comments)

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -111,7 +117,7 @@ def _generate_perf_configs(self):
 
     def _create_perf_config_params(self):
         perf_config_params = {
-            'model-name': [self._model_name],
+            'model-names': [self._model_names],
```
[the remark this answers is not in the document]
tgerdesnv: looks like the downstream code will permutate through all combinations. If that is a string, then it treats it as an array of characters and creates a config for each character.  I see other places of code (maybe only test code) that doesn't use an array though. I'm going to dig deeper

tests/test_perf_analyzer_config_generator.py
tgerdesnv: fixed

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -170,3 +176,6 @@ def _calculate_throughput_gain(self, reverse_index):
 
     def _get_throughput(self, measurement):
         return measurement.get_metric_value('perf_throughput')
+
+    def _make_model_names_str(self, model_names):
+        return ",".join(model_names)
```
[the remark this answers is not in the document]
tgerdesnv: fixed

tests/common/test_utils.py
tgerdesnv: This is just a single string (that may have multiple model names separated by strings). This is just a test helper function

Part of the confusion was perf_config_parameters[model-names] vs PerfConfig[model-names]. The first needs an array of possible values. The second is a string.

matthewkotila (approved):

PR #307, opened 2022-01-20
add cli testing

debermudez:
Adds testing to the cli options

This test does 2 things:
First it will  exercise every cli option from the command line with:
- long flag
- short-flag
- default option
- a value that should cause a failure

Second, it compares all of the options tested against all available options in the config objects. If there is a discrepancy, the test will fail, forcing users to add a test for any new cli options.


tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+    Struct class to hold the common variables shared between analyze tests
+    """
+
+    def __init__(self):
+        #yapf: disable
+        self.args = [
+            '/usr/local/bin/model-analyzer',
+            'report',
+            '--report-model-configs',
+            'a, b, c'
```
nv-braf, line 113: Is there any reason this has spaces after the commas and analyze doesn't?

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+                 default_value=None,
+                 expected_failing_value=None,
+                 extra_commands=None):
+
+        self.long_flag = long_flag
+        self.short_flag = short_flag
+        self.expected_value = expected_value
+        self.default_value = default_value
+        self.expected_failing_value = expected_failing_value
+        self.type = type_str
```
nv-braf, line 141: why is this called type_str (versus just matching the internal variable name)

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+    """
+    Tests the methods of the CLI class
+    """
+
+    def test_basic_cli_config_profile_options(self):
+        """
+        Test the minimal set of cli commands necessary to run Model Analyzer profile
+        """
+        cli = CLIConfigProfileStruct()
+        _, config = cli.parse()
```
nv-braf, line 171: Not familiar with this syntax: what does _, do?

tests/test_cli_options.py
debermudez: Nope. They should both work

tests/test_cli_options.py
debermudez: No reason. I will update it.

tests/test_cli_options.py
debermudez: _ is a assignment placeholder that is immediately thrown away.  The code here `cli.parse()` returns a tuple. I am immediately unpacking the tuple to get the config arguments but i dont care about the args that are returned as part of the tuple. I cannot unpack the tuple and not assign the args so I use the `_` here to assign them but immediately throw them away, thereby avoiding errors and warnings around unused variables.

nv-braf: Thanks!

docs/config.md
```diff
@@ -78,16 +78,16 @@ profile_models: <comma-delimited-string-list>
 [ override_output_model_repository: <boolean> | default: false ]
 
 # Concurrency values to be used for the analysis
-[ concurrency: <comma-delimited-string|list|range> | default: 1 ]
```
tgerdesnv, line 81: It wouldn't be handled in the CLI code, but technically the true default if unspecified is 1,2,4,8,...,1024  (right @nv-braf ? Think we should mention the true default?)

model_analyzer/analyzer.py
```diff
@@ -153,9 +153,8 @@ def analyze(self, mode, quiet):
         # Create result tables, put top results and get stats
         self._result_manager.create_tables()
         self._result_manager.compile_and_sort_results()
-        if self._config.summarize:
```
tgerdesnv, line 156: this was removed because it was always true?

model_analyzer/cli/cli.py
```diff
@@ -88,13 +88,13 @@ def _add_config_arguments(self, subparser, config):
         config : Config
             Model Analyzer config object.
         """
-
+        #configs is dictionary of config_fields objects from config_command_*
```
tgerdesnv, line 91: ty for clarifying/cleaning this

model_analyzer/config/input/config_command_analyze.py
```diff
@@ -249,15 +249,6 @@ def _add_report_configs(self):
         Adds report related configs
         """
 
-        self._add_config(
```
tgerdesnv, line 252: this is removed because now we have explicit profile/analyze steps?

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+from model_analyzer.config.input.config_command_report import ConfigCommandReport
+from model_analyzer.config.input.config_status import ConfigStatus
+from model_analyzer.constants import CONFIG_PARSER_SUCCESS
+
+import psutil
+
+from unittest.mock import patch
+
+
+class CLISubclass(CLI):
```
tgerdesnv, line 36: I feel like this could have been done with a single patch or some other way of faking out the ARGV.  
This maybe? @matthewkotila 
https://github.com/triton-inference-server/model_analyzer/blob/main/tests/test_cli.py#L35

Also, should that file be combined into this one?

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+        #yapf: enable
+        config_report = ConfigCommandReport()
+        self.cli = CLISubclass()
+        self.cli.add_subcommand(cmd='report', help='', config=config_report)
+
+    def parse(self):
+        return self.cli.parse(self.args)
+
+
+class OptionStruct():
```
tgerdesnv, line 124: comment

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+    @patch(
+        'model_analyzer.config.input.config_command_analyze.ConfigCommandAnalyze._load_config_file'
+    )
+    @patch(
+        'model_analyzer.config.input.config_command_analyze.ConfigCommandAnalyze._preprocess_and_verify_arguments'
+    )
+    @patch(
+        'model_analyzer.config.input.config_command_profile.ConfigCommandProfile._load_config_file'
+    )
+    def test_all_options(self, mocked_load_config_file_profile,
```
tgerdesnv, line 192: We need a descriptive comment of all the magic, either here or at the top of the class.  People need to be able to come to this file and add new options without needing to ask you how it all works.

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+            OptionStruct("stringlist", "analyze", "--inference-output-fields", None, "a, b, c",
+                "model_name,batch_size,concurrency,model_config_path,instance_group,satisfies_constraints,perf_throughput,perf_latency_p99"),
+            OptionStruct("stringlist", "analyze", "--gpu-output-fields", None, "a, b, c",
+                "model_name,gpu_uuid,batch_size,concurrency,model_config_path,instance_group,satisfies_constraints,gpu_used_memory,gpu_utilization,gpu_power_usage"),
+            OptionStruct("stringlist", "analyze", "--server-output-fields", None, "a, b, c",
+                "model_name,gpu_uuid,gpu_used_memory,gpu_utilization,gpu_power_usage"),
+
+            # No OP Options:
+            # Option format:
+            # (noop, any MA step, long_flag,)
```
tgerdesnv, line 280: extra comma

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+                 short_flag=None,
+                 expected_value=None,
+                 default_value=None,
+                 expected_failing_value=None,
+                 extra_commands=None):
+
+        self.long_flag = long_flag
+        self.short_flag = short_flag
+        self.expected_value = expected_value
+        self.default_value = default_value
```
tgerdesnv, line 139: is this the expected default value?

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+        # Verify the default value for the option
+        if default_value is not None:
+            cli = option_struct.cli_subcommand()
+            _, config = cli.parse()
+            option_value = config.get_config().get(
+                long_option_with_underscores).default_value()
+            self.assertEqual(option_value, default_value_converted)
+
+    # Helper methods
+    def _convert_flag(self, option):
```
tgerdesnv, line 491: convert_flag_to_?

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+            OptionStruct("noop", "yaml_analyze", "plots"),
+        ]
+        #yapf: enable
+
+        all_tested_options_set = set()
+
+        for option in options:
+            all_tested_options_set.add(self._convert_flag(option.long_flag))
+
+            if option.type in ["bool"]:
```
tgerdesnv, line 307: using an enum here would make sure that you are covering all the cases via an 'else: assert'

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+            CLIConfigProfileStruct, CLIConfigAnalyzeStruct,
+            CLIConfigReportStruct
+        ]
+        for struct in structs:
+            cli = struct()
+            _, config = cli.parse()
+            for key in config.get_config().keys():
+                cli_option_set.add(key)
+
+        self.assertEqual(cli_option_set, all_tested_options_set)
```
tgerdesnv, line 332: people are going to hit this assert when adding a new option and not understand why or how to fix it.  It looks like assertEqual takes an optional 3rd arg for the message to print when failing. We should really be utilizing that more in our tests.

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+
+        # Test long_flag
+        cli = option_struct.cli_subcommand()
+        cli.args.extend([long_option, expected_value_string])
+        _, config = cli.parse()
+        option_value = config.get_config().get(
+            long_option_with_underscores).value()
+        self.assertEqual(option_value, expected_value)
+
+        # Test short_flag
```
tgerdesnv, line 374: Is there any difference between all the test_short_flag in the various options? If not, seems like all of these should go into helper functions and be shared

tgerdesnv (changes requested):

docs/config.md
Tabrizian: I just wanted to mention that the default value (before the refactor) should've been 1. The exponential increase that I mentioned during the meeting was only for the automatic sweep.

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+        self.cli.add_subcommand(cmd='report', help='', config=config_report)
+
+    def parse(self):
+        return self.cli.parse(self.args)
+
+
+class OptionStruct():
+
+    def __init__(self,
+                 type,
```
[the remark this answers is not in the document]
debermudez: It does: https://docs.python.org/3/library/enum.html
I don't have a lot of experience with enums so I wasnt sure if this class would provide similar functionality. I came up with this for now but I am happy to change it if you think there is a better way.

tests/test_cli_options.py
debermudez: This will be the expected default value that is set in config_defaults.py. I set most of those in the test.

tests/test_cli_options.py
debermudez: I tried to add comments about each of the forms of the OptionStruct class that I used and I grouped the different types together. 
If you do not think those comments are not sufficient, lets chat really fast offline about it. I am probably too close to this to know what to describe at this point.

tests/test_cli_options.py
debermudez: I can always just add an assert here as an else anyway. If we switch to using an enum, I will update this part too.

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+    # Helper methods
+    def _convert_flag(self, option):
+        return option.lstrip("-").replace("-", "_")
+
+    def _convert_string_to_numeric(self, number):
+        return float(number) if "." in number else int(number)
+
+    def _convert_string_to_int_list(self, list_values):
+        ret_val = [int(x) for x in list_values.split(",")]
+        if len(ret_val) == 1:
```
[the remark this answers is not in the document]
debermudez: All of the args need to come in as strings.
Some of the args are ultimately stored as lists of ints. However, if the list of ints is length(1), then the single element list is stored as an int primitive. 
This function handles all of that to mimic our configuration field behavior.

tests/test_cli_options.py
```diff
@@ -0,0 +1,507 @@
...
+            _, config = cli.parse()
+            option_value = config.get_config().get(
+                long_option_with_underscores).default_value()
+            self.assertEqual(option_value, default_value_converted)
+
+    # Helper methods
+    def _convert_flag(self, option):
+        return option.lstrip("-").replace("-", "_")
+
+    def _convert_string_to_numeric(self, number):
```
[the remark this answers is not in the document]
debermudez: a type cast will do string->int or string->float.
Since I handle them with the same functions, this one handles both.

tests/test_cli_options.py
debermudez: the internal representation that the config fields use.

tests/test_cli_options.py
debermudez: and the other ones too. Should be a bit cleaner now.

docs/config.md
tgerdesnv: ok, I'm fine to keep the doc how it is then.

model_analyzer/analyzer.py
tgerdesnv: but.. should we fix the functionality instead of ripping it out?

tests/test_cli_options.py
tgerdesnv: ok, I would recommend renaming to expected_default_value then

tests/test_cli_options.py
```diff
@@ -0,0 +1,517 @@
...
+        ]
+        for struct in structs:
+            cli = struct()
+            _, config = cli.parse()
+            for key in config.get_config().keys():
+                cli_option_set.add(key)
+
+        self.assertEqual(
+            cli_option_set, all_tested_options_set,
+            "The available options on the CLI does not match the available options tested."
```
tgerdesnv, line 339: add "If you recently added or removed a CLI option, please update the OptionsStruct list above"

tgerdesnv (changes requested):

tgerdesnv (approved):

PR #308, opened 2022-01-21
Changes/fixes to get model manager unit tests passing

nv-braf:
The big change here is that I had the PACG/MCG iteration loops reversed, so early exit wasn't working.
Everything else is minor.


model_analyzer/model_manager.py
```diff
@@ -74,9 +74,11 @@ def run_models(self, models):
         rcg = RunConfigGenerator(config=self._config,
                                  models=models,
                                  client=self._client)
+
+        run_config = rcg.next_config()
         while not rcg.is_done() and not self._state_manager.exiting():
-            run_config = rcg.next_config()
-            measurement = self._metrics_manager.execute_run_config(run_config)
+            measurement = self._metrics_manager.execute_run_config(
+                next(run_config))
```
matthewkotila, line 81: how is `next()` `StopIteration` exception ([link](https://docs.python.org/3/library/functions.html#next)) avoided here? is it tied to `rcg.is_done()`?

tests/test_model_manager.py
```diff
@@ -674,7 +674,7 @@ def _test_model_manager(self, yaml_content, expected_ranges):
                                      MagicMock(), MagicMock(), metrics_manager,
                                      MagicMock(), state_manager)
 
-        model_manager.run_models(config.profile_models[0])
+        model_manager.run_models(config.profile_models)
```
matthewkotila, line 677: why this change?

model_analyzer/config/generate/run_config_generator.py
```diff
@@ -60,22 +66,22 @@ def next_config(self):
...
-                                           self._model_pa_flags,
-                                           self._model_parameters)
-
         model_configs_list = self._generate_all_model_config_permuations(
             self._models)
 
-        while not pacg.is_done():
-            perf_analyzer_config = pacg.next_config()
+        for model_configs in model_configs_list:
+            self._pacg = PerfAnalyzerConfigGenerator(self._config,
```
matthewkotila, line 73: why set this here if it's set in `__init__()`?

matthewkotila (changes requested):

model_analyzer/analyzer.py
```diff
@@ -110,7 +110,8 @@ def profile(self, client, gpus):
             self._server.stop()
 
         # Profile all models concurrently
-        if self._config.run_config_profile_models_concurrently_enable:
+        if self._config.run_config_profile_models_concurrently_enable and len(
```
tgerdesnv, line 113: turn this into a function

model_analyzer/config/generate/run_config_generator.py
```diff
@@ -46,11 +46,17 @@ def __init__(self, config, models, client):
...
+                                                 self._model_pa_flags,
+                                                 self._model_parameters)
+
+        self._model_configs_are_on_final_iteration = False
 
     def is_done(self):
-        """ Returns true if this generator is done generating configs """
-        return self._is_done
+        return self._pacg.is_done(
+        ) and self._model_configs_are_on_final_iteration
```
tgerdesnv, line 59: did yapf do this?

model_analyzer/config/generate/run_config_generator.py
```diff
@@ -60,22 +66,22 @@ def next_config(self):
...
-        pacg = PerfAnalyzerConfigGenerator(self._config, self._model_names,
-                                           self._model_pa_flags,
-                                           self._model_parameters)
-
         model_configs_list = self._generate_all_model_config_permuations(
             self._models)
 
-        while not pacg.is_done():
-            perf_analyzer_config = pacg.next_config()
+        for model_configs in model_configs_list:
```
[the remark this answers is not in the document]
tgerdesnv: I think model_configs will have multiple model configs, and model_configs_list is a list of those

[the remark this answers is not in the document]
nv-braf: I don't think so, because each entry in `model_configs_list` is a list of `model_configs`

model_analyzer/config/generate/run_config_generator.py
```diff
@@ -88,7 +94,16 @@ def set_last_results(self, measurements):
         ----------
         measurements: List of Measurements from the last run(s)
         """
-        self._curr_pacg.set_last_results(measurements)
+        self._pacg.set_last_results(measurements)
+
+    def _create_run_config(self, model_configs, perf_analyzer_config):
+        if isinstance(model_configs, list):
```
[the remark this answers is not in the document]
tgerdesnv: yeah it seems like it should be the same in all cases.  Even in the single model case we should move to it being a list with 1 element

tgerdesnv (changes requested):

model_analyzer/config/generate/run_config_generator.py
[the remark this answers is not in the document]
nv-braf: I discovered that when a model has only one config the recursive function doesn't return it as a list. Rather than try to fix the function, I thought it was easier just to check for this when creating the run config. I'll add a comment about this in the code

matthewkotila (approved):

model_analyzer/analyzer.py
nv-braf: Created `_should_profile_models_concurrently()`

model_analyzer/config/generate/run_config_generator.py
nv-braf: Yes. Wrapping it in `()`'s fixes the weird formatting

tgerdesnv (approved):

PR #309, opened 2022-01-21
Add support for loading multiple models onto the Triton Server

matthewkotila:


tests/test_metrics_manager.py
```diff
@@ -0,0 +1,95 @@
...
+from model_analyzer.triton.model.model_config import ModelConfig
+
+
+class TestMetricsManager(trc.TestResultCollector):
+
+    @patch('model_analyzer.record.metrics_manager.MetricsManager.__init__',
+           return_value=None)
+    def test_execute_run_config(self, mock_metrics_manager_init):
+        """
+        Tests that something something.
```
nv-braf, line 40: Update comment

tests/test_metrics_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        metrics_manager._get_measurement_if_config_duplicate = MagicMock(
+            return_value=None)
+        metrics_manager._server = MagicMock()
+        metrics_manager._config = config
+        metrics_manager._client = MagicMock()
+        model_configs = [
+            ModelConfig(
+                json_format.ParseDict({'name': 'model_config_1'},
+                                      model_config_pb2.ModelConfig())),
+            ModelConfig(
```
nv-braf, line 60: Should these have different names?

model_analyzer/record/metrics_manager.py
```diff
@@ -166,22 +166,23 @@ def execute_run_config(self, run_config):
...
-        if not self._load_model_variant(
-                variant_config=run_config.model_config()):
-            self._server.stop()
-            return
-
+        for model_config in run_config.model_configs():
+            if not self._load_model_variant(variant_config=model_config):
+                self._server.stop()
+                return
+        return
```
nv-braf, line 185: Is the second return needed?

tests/test_metrics_manager.py
matthewkotila: updated 👍

tests/test_metrics_manager.py
matthewkotila: nice catch--yes, for illustrative purposes. updated to have different names 👍

model_analyzer/record/metrics_manager.py
matthewkotila: whoops, left this in by mistake. removed it and added a few more mocks to test to skip subsequent irrelevant parts below this return 👍

nv-braf (approved):

model_analyzer/record/metrics_manager.py
```diff
@@ -166,21 +166,22 @@ def execute_run_config(self, run_config):
         measurement to the result manager
         """
 
-        # Create model variant
-        self._create_model_variant(original_name=run_config.model_name(),
-                                   variant_config=run_config.model_config())
+        # Create model variants
+        for model_config in run_config.model_configs():
+            self._create_model_variant(original_name=run_config.model_name(),
```
tgerdesnv, line 171: is this function smart enough to do nothing if it is already done? Unlike before, much of the time the model variant will already exist from a previous run.

model_analyzer/record/metrics_manager.py
```diff
@@ -166,21 +166,22 @@ def execute_run_config(self, run_config):
         measurement to the result manager
```
tgerdesnv, line 166: This function does a lot. Can we refactor into some smaller functions so it is easier to read? (At least do the code you touched)

tgerdesnv (changes requested):

model_analyzer/record/metrics_manager.py
matthewkotila: yes, if it fails to create the directory, it just continues with the rest of the program:

https://github.com/triton-inference-server/model_analyzer/blob/main/model_analyzer/record/metrics_manager.py#L223-L224

tgerdesnv: What about the 'first config variant' code? That seems like it wouldn't work for multiple models?

model_analyzer/record/metrics_manager.py
matthewkotila: refactored into smaller functions 👍

model_analyzer/record/metrics_manager.py
matthewkotila: updated logic to have a first config variant per model 👍

tgerdesnv: looks like it should work. Is there any testing or at least a manual run to confirm?

matthewkotila: This can't really be tested yet with multiple models since other end-to-end pieces are missing, and it's not well set up for a unit test (an L0 is better since the feature we want to test ranges across several layers of the code execution, meaning the mocking would be a lot).

I did test, however, with one model, and the old behavior still works. The variant directories get created correctly and all subsequent variant directories still link to the first variant.

tgerdesnv (approved):

PR #311, opened 2022-01-26
Rename the default configuration to {model_name}_config_default

tgerdesnv:
Before, the default config was always run but it was still named <model>_config<#>.  Now it is named <model>_config_default.

### DIFF
--- a/model_analyzer/config/generate/model_config_generator.py
+++ b/model_analyzer/config/generate/model_config_generator.py
@@ -39,6 +39,7 @@ class ModelConfigGenerator(ConfigGeneratorInterface):
         self._base_model = model
         self._base_model_name = model.model_name()
         self._model_name_index = 0
+        self._default_param_combo = {}
         self._configs = self._generate_model_configs()
 
     def is_done(self):
@@ -103,6 +104,9 @@ class ModelConfigGenerator(ConfigGeneratorInterface):
             for key, value in param_combo.items():
                 if value is not None:
                     model_config_dict[key] = value
+
+        model_config_dict['name'] = self._get_model_variant_name(param_combo)
+
         model_config = ModelConfig.create_from_dictionary(model_config_dict)
         return model_config
 
@@ -121,14 +125,14 @@ class ModelConfigGenerator(ConfigGeneratorInterface):
             else:
                 param_combos = self._automatic_search_configs()
 
-        if not self._is_default_config_in_configs(param_combos):
-            self._add_default_config(param_combos)
+        if not self._is_default_combo_in_param_combos(param_combos):
+            self._add_default_combo(param_combos)
 
         return param_combos
 
     def _automatic_search_disabled_configs(self):
         """ Return the configs when we want to search but searching is disabled """
-        return [{}]
+        return [self._default_param_combo]
 
     def _automatic_search_configs(self):
         """ Search through automatic search variables to generate configs """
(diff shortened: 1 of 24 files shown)


model_analyzer/config/generate/model_config_generator.py
```diff
@@ -160,23 +164,22 @@ def _make_remote_model_config(self):
...
-        # Add in an empty configuration, which will apply the default values
-        configs.append({})
+    def _add_default_combo(self, param_combos):
+        # Add in an empty combo at the start of the list, which will just apply the default values
+        #
+        param_combos.insert(0, self._default_param_combo)
 
     def _finalize_configs(self, configs):
         for config in configs:
-            model_tmp_name = self._base_model_name
```
tgerdesnv, line 172: naming is now done earlier

model_analyzer/triton/model/model_config.py
```diff
@@ -183,22 +183,16 @@ def write_config_to_file(self, model_path, src_model_path,
         model_config_bytes = text_format.MessageToBytes(self._model_config)
         # Create current variant model as symlinks to first variant model
         if first_variant_model_path is not None:
-            src_model_name = Path(src_model_path).name
-            first_variant_model_name = Path(first_variant_model_path).name
-
-            # If the model files have been copied once, do not copy it again
-            if re.search(f'^{src_model_name}_config\d+$',
```
tgerdesnv, line 190: @matthewkotila I know you touched code around here recently. Do I need this code that I ripped out? In what case would it kick in and skip?

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -160,23 +164,22 @@ def _make_remote_model_config(self):
...
             config.set_cpu_only(self._base_model.cpu_only())
 
-    def _get_model_variant_name(self):
-        variant_name = f'{self._base_model_name}_config{self._model_name_index}'
-        self._model_name_index += 1
+    def _get_model_variant_name(self, param_combo):
+        if param_combo is self._default_param_combo:
+            variant_name = f'{self._base_model_name}_config_default'
+        else:
+            variant_name = f'{self._base_model_name}_config_{self._model_name_index}'
```
tgerdesnv, line 183: @debermudez I am adding the underscore between "config" and the # (that @nv-braf requested in your review) now. That way the format is consistent with the new config_default. Is there anything else that needs to change (documentation?)

model_analyzer/triton/model/model_config.py
debermudez: Why did we remove this regex check?
It was checking to make sure we didn't copy files multiple times. It would create a symlink to the first variant for a subset of the output files that don't change between variants.

model_analyzer/config/generate/model_config_generator.py
debermudez: - qa/L0_profile/check_results.py
- qa/L0_quick_start/test.sh
- tests/test_model_config.py
- tests/test_run_config_generator.py
- model_analyzer/config/run/run_config_generator.py
- model_analyzer/triton/model/model_config.py

Those would be all the files to touch.

model_analyzer/config/generate/model_config_generator.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, config, model, client):
         self._base_model = model
         self._base_model_name = model.model_name()
         self._model_name_index = 0
+        self._default_param_combo = {}
```
tgerdesnv, line 42: this feels like it should be a constant? What is the correct way to do that in python?

model_analyzer/triton/model/model_config.py
tgerdesnv: Yes, I've manually run and all tests are passing

matthewkotila: then we should be good to remove it 👍

debermudez: works for me.

model_analyzer/config/generate/model_config_generator.py
debermudez: If the constant is only used in this file, you can pull it out to a class variable like here: https://github.com/triton-inference-server/model_analyzer/blob/7a6792e94969fdddbec8ee5a62b2b80a2b9e91ad/model_analyzer/reports/report_factory.py#L29

If you are using this in multiple files, you can add it to the constants file and import it.

model_analyzer/config/generate/model_config_generator.py
debermudez: run_config_generator change: https://github.com/triton-inference-server/model_analyzer/pull/281/files#diff-9f379e5c26489133c3fdf98fabfc766b3845fc0f9164f920b525750b57c17995R133 
I think you are okay on this one.

test_run_config_generator: instances of the name vgg_16_graphdef_config0 should have to change to vgg_16_graphdef_config_0 right?

tgerdesnv: I don't see that anywhere in test_run_config_generator or anywhere in our repo

debermudez: must have changes since i made the first upgrade. Lets resolve this one then.

debermudez (approved):

PR #315, opened 2022-01-28
fix issue where early PACG exit skipped over other batch sizes

tgerdesnv:


model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -95,31 +104,35 @@ def set_last_results(self, measurements):
 
     def _create_concurrency_list(self, cli_config, model_parameters):
         if model_parameters['concurrency']:
-            return model_parameters['concurrency']
+            return sorted(model_parameters['concurrency'])
```
tgerdesnv, line 107: sorted to guarantee we are always strictly increasing concurrency

However... are we even supposed to early exit if the user specified a fixed list? Shouldn't we only be exiting on the auto-search?

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -69,16 +77,17 @@ def __init__(self, cli_config, model_names, model_perf_analyzer_flags,
...
-              or     self._last_results_erroneous() \
-              or not self._throughput_gain_valid()
-        #yapf: enable
+        return self._done_walking() or self._last_results_erroneous()
 
     def next_config(self):
         """ Returns the next generated config """
-        config = self._configs[self._curr_config_index]
-        self._curr_config_index += 1
+        if self._done_walking_concurrencies():
```
tgerdesnv, line 84: @nv-braf I could use some guidance on state machine. It seems bad or unclean to step one thing before, and another thing after. Is there a proper way to do this?

nv-braf (dismissed):
Looks good

nv-braf (approved):

PR #316, opened 2022-01-28
Include default config in summary reports

tgerdesnv:
Before and after reports and images are in the JIRA story

### DIFF
--- a/model_analyzer/result/result_manager.py
+++ b/model_analyzer/result/result_manager.py
@@ -750,17 +750,15 @@ class ResultManager:
         if not model_name:
             return
 
-        found = False
         for result in results:
             if result.model_config().get_field(
                     'name') == f"{model_name}_config_default":
-                found = True
-        if not found:
-            for result in result_heap.results():
-                if result.model_config().get_field(
-                        'name') == f"{model_name}_config_default":
-                    results.append(result)
-                    break
+                return
+        for result in result_heap.results():
+            if result.model_config().get_field(
+                    'name') == f"{model_name}_config_default":
+                results.append(result)
+                return
 
     def get_result_statistics(self):
         """
--- a/model_analyzer/result/result_manager.py
+++ b/model_analyzer/result/result_manager.py
@@ -358,7 +358,7 @@ class ResultManager:
             model_config_data = results[model_name][model_config_name]
             return model_config_data[0], list(model_config_data[1].values())
 
-    def top_n_results(self, model_name=None, n=-1):
+    def top_n_results(self, model_name=None, n=-1, include_default=False):
         """
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


cvinson830:
Comparing 'old_summary' with 'new_summary' PNG files:
- The legend for the old summary is tucked away nicely on the top left corner of the plot
- The legend for the new summary overwrites the line series
- Can we move the legend back into the top left?

image

cvinson830:
The file 'old result summary online.pdf' has 'default' in the legend.  Are you sure the new string 'default' was in the old result summary pdf?

model_analyzer/result/result_manager.py
```diff
@@ -733,6 +742,26 @@ def _write_result(self,
...
+        add it from the result heap
+        '''
+        if not model_name:
+            return
+
+        found = False
+        for result in results:
+            if result.model_config().get_field(
+                    'name') == f"{model_name}_config_default":
+                found = True
```
cvinson830, line 757: Needs a break here (like line 763) so that the remaining iterations of the loop are not executed

cvinson830:
Highlighting potential engineer-speak: 

Instead of embedding `underscore default` as a suffix like this

image


...it might be more clear to put default in parentheses like this

add_sub_config_x (default)
add_sub_config_0
add_sub_config_1


Additionally, I think I heard you say in a meeting that the default config is available in all the reports.  What I'm seeing here is we're losing the name of the default config (`config_x`).  It's being replaced with the word `default`.  

In summary, 
- restore the name of the config in the line series
- append ` (default)`.

tgerdesnv:
> Comparing 'old_summary' with 'new_summary' PNG files:
> 
>     * The legend for the old summary is tucked away nicely on the top left corner of the plot
> 
>     * The legend for the new summary overwrites the line series
> 
>     * Can we move the legend back into the top left?
> 
> 
> image

I noticed that as well. We are using matplotlib under the hood to generate these, which puts the legend "at the location, among the nine locations defined so far, with the minimum overlap with other drawn artists".  Thus the location is data specific. There are more advanced controls in matplotlib for legend location, but I don't want to modify them, at least not in this story.

model_analyzer/result/result_manager.py
tgerdesnv: I removed the breaks and just return early now

tgerdesnv:
> Highlighting potential engineer-speak:
> 
> Instead of embedding `underscore default` as a suffix like this
> 
> image
> 
> ...it might be more clear to put default in parentheses like this
> 
> add_sub_config_x (default) add_sub_config_0 add_sub_config_1
> 
> Additionally, I think I heard you say in a meeting that the default config is available in all the reports. What I'm seeing here is we're losing the name of the default config (`config_x`). It's being replaced with the word `default`.
> 
> In summary,
> 
>     * restore the name of the config in the line series
> 
>     * append ` (default)`.

We aren't losing the name of the default config. The name of the default config is `add_sub_config_default`. That was changed recently [here](https://github.com/triton-inference-server/model_analyzer/pull/311)

cvinson830 (commented):
test comment

cvinson830 (approved):

PR #317, opened 2022-01-31
Support scientific notation PA throughput

tgerdesnv:
Example data:
Pass [1] throughput: 5.80911e+07 infer/sec. Avg latency: 18023 usec (std 699 usec)

Before this change that data would be observed as 5.8 inferences/second instead of 58091100 inferences/second. This change fixes it

### DIFF
--- a/model_analyzer/perf_analyzer/perf_analyzer.py
+++ b/model_analyzer/perf_analyzer/perf_analyzer.py
@@ -356,7 +356,8 @@ class PerfAnalyzer:
         Parses throughput from the perf analyzer output
         """
 
-        throughput = re.search('Throughput: (\d+\.\d+|\d+)', section)
+        throughput = re.search('Throughput: (\d+\.\d+e\+\d+|\d+\.\d+|\d+)',
+                               section)
         if throughput:
             throughput = float(throughput.group(1))
             return PerfThroughput(value=throughput)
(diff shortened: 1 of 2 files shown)


model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -356,7 +356,8 @@ def _parse_perf_throughput(self, section):
         Parses throughput from the perf analyzer output
         """
 
-        throughput = re.search('Throughput: (\d+\.\d+|\d+)', section)
+        throughput = re.search('Throughput: (\d+\.\d+e\+\d+|\d+\.\d+|\d+)',
```
cvinson830, line 359: This assumes positive numbers.  Any chance this could every be negative?

cvinson830, line 359: `Throughput:` is a literal string.  Can this be abstracted?  If this string ever changes, two code changes will be needed..

cvinson830 (changes requested):

model_analyzer/perf_analyzer/perf_analyzer.py
tgerdesnv: No, we can't have negative inferences per second ;)

tgerdesnv: @nv-braf is already working on a story that converts this parsing to pull from a CSV instead, which would make this moot (due to column headers).  The most important part of this story is the unit test that will protect the feature when he refactors.

cvinson830 (approved):

PR #321, opened 2022-02-05
add shm size support

debermudez:
Adds shm_size support when running triton_server in docker mode


model_analyzer/config/input/config_command_profile.py
```diff
@@ -581,6 +581,14 @@ def _add_triton_configs(self):
                 description=
                 'A dictionary of name-value labels to set metadata for the Triton '
                 'server docker container in docker launch mode'))
+        self._add_config(
```
tgerdesnv, line 584: Have you confirmed that if the user doesn't pass in anything, that the docker still has the default shm?

debermudez: yes. manually

PR #324, opened 2022-02-08
Remove dependency on python-apt

debermudez:


wphicks (approved):
LGTM! Should handle the various install cases I was concerned about.

tgerdesnv (approved):

PR #333, opened 2022-02-24
Max-batch-size sweep

tgerdesnv:
Prep work to maintain the same functionality:

- ModelConfigGenerator split into BaseModelConfigGenerator, ManualModelConfigGenerator (precomputed), and AutomaticModelConfigGenerator (live state machine)
- BaseModelConfigGenerator defines interface functions is_done(), next_config() and set_last_results(), but adds new virtual functions _done_walking(), _get_next_model_config(), and _step() that the children must define
- New factory ConfigGeneratorFactory to pick between manual vs automatic
- Change PACG, PAC, and RunConfig to only have a single model instead of array of models
- ModelManager run_model only takes a single model
- Remove some invalid tests

New changes:
- If in automatic sweep mode, sweep exponentially from min_model_batch_size to max_model_batch_size
- If an increase in throughput is not observed, stop walking max_batch_size
- If the supplied model config has max_batch_size of 0, do not sweep max_batch_size
- Skip run configs where PA batch size is greater than the model variant's max_batch_size


model_analyzer/analyzer.py
```diff
@@ -112,7 +112,8 @@ def profile(self, client, gpus):
         # Profile all models concurrently
         if self._should_profile_multiple_models_concurrently():
             try:
-                self._model_manager.run_models(
+                # FIXME TKG need new function for this
```
tgerdesnv, line 115: need to resolve this fixme. Just put an assert here for now? Will be added back in by 518

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        Returns
+        -------
+        ModelConfig
+            The next ModelConfig generated by this class
+        """
+        self._live = True
+        while True:
+            config = self._get_next_model_config()
+            yield (config)
```
nv-braf, line 60: I don't understand how this doesn't return an instance count of 0 because we don't step before returning the first model config

model_analyzer/config/generate/manual_model_config_generator.py
```diff
@@ -0,0 +1,124 @@
...
+        if self._remote_mode:
+            configs = self._generate_remote_mode_model_configs()
+        else:
+            configs = self._generate_direct_modes_model_configs()
+
+        self._finalize_configs(configs)
+
+        return configs
+
+    def _generate_remote_mode_model_configs(self):
```
nv-braf, line 64: Why isn't this also part of the base class?

model_analyzer/config/generate/manual_model_config_generator.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        model_config_params = self._base_model.model_config_parameters()
+        if model_config_params:
+            param_combos = GeneratorUtils.generate_combinations(
+                model_config_params)
+        else:
+            if self._search_disabled:
+                param_combos = self._generate_search_disabled_configs()
+            else:
+                raise TritonModelAnalyzerException(
```
nv-braf, line 91: Is it even possible to reach this? Doesn't the factory prevent this?

debermudez: Its a good idea to include our assumption in the code. If we ever update this code and this line is reached, its enshrined that a mistake was made somewhere.

tests/test_model_config_generator.py
```diff
@@ -41,11 +41,46 @@ def test_direct_no_params(self):
             """)
 
         expected_configs = [
-            {'dynamic_batching': {}, 'instance_group': [{'count': 1, 'kind': 'KIND_GPU'}]},
-            {'dynamic_batching': {}, 'instance_group': [{'count': 2, 'kind': 'KIND_GPU'}]},
-            {'dynamic_batching': {}, 'instance_group': [{'count': 3, 'kind': 'KIND_GPU'}]},
-            {'dynamic_batching': {}, 'instance_group': [{'count': 4, 'kind': 'KIND_GPU'}]},
-            {'dynamic_batching': {}, 'instance_group': [{'count': 5, 'kind': 'KIND_GPU'}]},
+            {'dynamic_batching': {}, 'max_batch_size': 1, 'instance_group': [{'count': 1, 'kind': 'KIND_GPU'}]},
```
debermudez, line 44: Should we add a utility function to generate this dynamically?  
These types of configs are in multiple files

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,67 @@
...
+        super().__init__(config, model, client)
+
+        self._max_instance_count = config.run_config_search_max_instance_count
+        self._max_max_batch_size = config.run_config_search_max_max_batch_size
+
+        self._instance_kind = "KIND_GPU"
+        if self._cpu_only:
+            self._instance_kind = "KIND_CPU"
+
+        self._curr_instance_count = 0
```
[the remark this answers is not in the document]
tgerdesnv: Bad state machine design (that worked, but was unclear). I have fixed it locally to be more clear

model_analyzer/config/generate/base_model_config_generator.py
tgerdesnv: It worked correctly, but now is more explicit

model_analyzer/config/generate/manual_model_config_generator.py
tgerdesnv: I moved _make_remote_model_config to base, as we discussed

model_analyzer/config/generate/manual_model_config_generator.py
tgerdesnv: Yes, just an assert to make sure we don't ever mess that up, or create from outside the factory

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _step_instance_count(self):
+        self._curr_instance_count += 1
+
+    def _done_walking_max_batch_size(self):
+        return self._max_batch_size_limit_reached() \
+            or not self._last_results_increased_throughput()
+
+    def _done_walking_instance_count(self):
+        return self._curr_instance_count >= self._max_instance_count
```
nv-braf, line 94: This should just be an == compare. We can't ever exceed max_instance_count

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _max_batch_size_limit_reached(self):
+        return self._curr_max_batch_size * 2 > self._max_model_batch_size
```
nv-braf, line 97: Add a ()'s around the multiplier for readability

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _reset_max_batch_size(self):
+        if self._sweep_max_batch_size_disabled:
+            self._curr_max_batch_size = self._max_model_batch_size
+        else:
+            self._curr_max_batch_size = self._min_model_batch_size
+        self._curr_max_batch_size_throughputs = []
+
+    def _get_last_results_max_throughput(self):
+
```
nv-braf, line 121: Remove whitespace

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            return max(throughputs)
+
+    def _get_next_model_config(self):
+        param_combo = self._get_curr_param_combo()
+        model_config = self._make_direct_mode_model_config(param_combo)
+        return model_config
+
+    def _get_curr_param_combo(self):
+
```
nv-braf, line 138: Remove whitespace

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+        if not self._sweep_max_batch_size_disabled:
+            config['max_batch_size'] = self._curr_max_batch_size
+
+        return config
+
+    def _determine_sweep_max_batch_size_disabled(self):
+        max_batch_size_disabled = False
+        config = self._get_base_model_config_dict()
+        if "max_batch_size" not in config or config['max_batch_size'] == 0:
```
nv-braf, line 158: Make this a one-liner (see my above comment) or at least swap the setting of False to be the line directly above

model_analyzer/config/input/config_command_profile.py
```diff
@@ -477,6 +478,24 @@ def _add_run_search_configs(self):
...
                 "Max instance count value that run config search should not go beyond that."
             ))
+        self._add_config(
+            ConfigField(
+                'run_config_search_max_model_batch_size',
+                flags=['--run-config-search-max-model-batch-size'],
+                field_type=ConfigPrimitive(int),
+                default_value=DEFAULT_RUN_CONFIG_MAX_MODEL_BATCH_SIZE,
+                description=
+                "Max value for the model's max_batch_size that run config search should not go beyond."
```
nv-braf, line 488: Get rid of Max at the start of the description: "Value for the ...."
And change "should" to "will"

model_analyzer/config/input/config_command_profile.py
```diff
@@ -477,6 +478,24 @@ def _add_run_search_configs(self):
...
+            ))
+        self._add_config(
+            ConfigField(
+                'run_config_search_min_model_batch_size',
+                flags=['--run-config-search-min-model-batch-size'],
+                field_type=ConfigPrimitive(int),
+                default_value=DEFAULT_RUN_CONFIG_MIN_MODEL_BATCH_SIZE,
+                description=
+                "Min value for the model's max_batch_size that run config search should start from."
```
nv-braf, line 497: Get rid of Min at the start of the description: "Value for the ...."
And change "should" to "will"

model_analyzer/config/run/run_config.py
```diff
@@ -75,6 +75,21 @@ def perf_config(self):
 
         return self._perf_config
 
+    def is_legal_combination(self):
+        """
+        Returns true if the run_config is valid and should be run. Else false
+        """
+        max_batch_size = 1
+        model_config = self._model_config.get_config()
+        if "max_batch_size" in model_config:
```
nv-braf, line 84: Change this to be a one-liner:
max_batch_size = model_config['max_batch_size'] if "max_batch_size" in model_config else 1

model_analyzer/config/run/run_config.py
```diff
@@ -75,6 +75,21 @@ def perf_config(self):
...
+            max_batch_size = model_config['max_batch_size']
+
+        perf_batch_size = 1
```
nv-braf, line 87: One-liner here as well

model_analyzer/model_manager.py
```diff
@@ -69,16 +70,19 @@ def run_models(self, models):
...
-                                 models=models,
+                                 model=model,
                                  client=self._client)
 
         run_config_generator = rcg.next_config()
         while not rcg.is_done() and not self._state_manager.exiting():
-            measurement = self._metrics_manager.execute_run_config(
-                next(run_config_generator))
+            run_config = next(run_config_generator)
+            measurement = None
```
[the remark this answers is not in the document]
tgerdesnv: The others are basic if/else for assigning a variable, but this one is gating actual execution of code and I don't think should be a 1-liner.

tests/test_model_manager.py
```diff
@@ -60,11 +60,15 @@ def execute_run_config(self, config):
     def _get_next_measurements(self):
         """ Return fake measurements as if the run_configs had been executed """
 
-        perf_throughput = PerfThroughput(self._get_next_perf_throughput_value())
-        non_gpu_data = [perf_throughput]
-        return Measurement(gpu_data=MagicMock(),
-                           non_gpu_data=non_gpu_data,
-                           perf_config=MagicMock())
+        throughput_value = self._get_next_perf_throughput_value()
+        if throughput_value is None:
```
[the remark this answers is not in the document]
tgerdesnv: See test_bad_result_early_PA_exit().  It overrides the value to be None. This is a valid case where PA can fail (out of memory or other reason) and we return None.

model_analyzer/config/run/run_config.py
tgerdesnv: done

model_analyzer/config/run/run_config.py
tgerdesnv: done

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+        model: The model to generate ModelConfigs for
+        client: TritonClient
+        """
+        super().__init__(config, model, client)
+
+        self._max_instance_count = config.run_config_search_max_instance_count
+        self._max_model_batch_size = config.run_config_search_max_model_batch_size
+        self._min_model_batch_size = config.run_config_search_min_model_batch_size
+
+        self._instance_kind = "KIND_GPU"
```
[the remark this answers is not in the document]
tgerdesnv: fixed

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+        if self._cpu_only:
+            self._instance_kind = "KIND_CPU"
+
+        # State machine counters
+        # (will be properly initialized in start_state_machine())
+        #
+        self._curr_instance_count = 0
```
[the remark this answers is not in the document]
tgerdesnv: They don't need to be initialized, but I think they should exist and be defined in init.

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -0,0 +1,123 @@
...
+        self._base_model = model
+        self._base_model_name = model.model_name()
+        self._cpu_only = model.cpu_only()
+        self._model_name_index = 0
+        self._live = False
+        self._last_results = []
+
+    def is_done(self):
+        """ Returns true if this generator is done generating configs """
+        return self._live and self._done_walking()
```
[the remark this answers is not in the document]
tgerdesnv: _last_results can be empty for a faulty run.  I've updated the name for this class (and for perf_analyzer_config_generator, which had the same variable)

model_analyzer/config/generate/manual_model_config_generator.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        Parameters
+        ----------
+        config: ModelAnalyzerConfig
+        model: The model to generate ModelConfigs for
+        client: TritonClient
+        """
+        super().__init__(config, model, client)
+
+        self._remote_mode = config.triton_launch_mode == 'remote'
```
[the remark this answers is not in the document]
tgerdesnv: fixed

model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # Contains the max throughput from each provided list of measurements
+        # since the last time we stepped max_batch_size
+        #
+        self._curr_max_batch_size_throughputs = []
+
+        # We return the default combo first before starting the state machine.
+        # This flag tracks if we have done that and thus are in the state machine
+        #
+        self._state_machine_started = False
```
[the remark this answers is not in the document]
tgerdesnv: I feel like the same comment would be needed, but in reverse? (to explain why I'm starting the state machine based on the default_combo_returned variable)

model_analyzer/model_manager.py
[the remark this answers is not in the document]
nv-braf: I guess I'm just not a fan of the style of setting and then conditionally overwriting. My preference would be to turn this into an if-else, even if it's not a one-liner:

Even though your executing code I still think this reads cleaner to me that what currently exists:

`measurement = self._metrics_manager.execute_run_config(run_config)  
                            if run_config.is_legal_combination() else None`

model_analyzer/config/generate/automatic_model_config_generator.py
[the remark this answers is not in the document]
nv-braf: But doesn't that expose you to a potential bug? If you remove these from __init__ then you guarantee a failure if they are used before _start_state_machine is called

I don't think these belong in init and should only exist once _start_state_machine is called, but if you want to leave them in __init__ then they should be assigned to None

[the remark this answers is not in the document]
tgerdesnv: I guess I'm just used to looking in init to see all the class member variables. It is scary to me for member variables to just appear inside of functions.

I tried removing the lines or setting them to None but tests failed. They are in fact referenced (in is_done()) before we call start_state_machine()

model_analyzer/model_manager.py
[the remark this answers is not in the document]
tgerdesnv: I put the None in an else case

nv-braf:
Everything looks good

nv-braf (approved):
Everything looks good

PR #336, opened 2022-03-02
Multi-model Measurement class creation

nv-braf:
Two new classes created to handle multi-model measurements:

- ModelConfigMeasurement
- RunConfigMeasurement

I've added logic for measurement comparison directly into this class (as opposed to putting it in ResultComparator)

Moved PA config generation into the test_utils (as it's used to create the unique key for RCMs)

Left the original Measurement class as is, no functionality has been changed, and these new classes are not instanced anywhere.


tests/test_model_config_measurement.py
```diff
@@ -0,0 +1,145 @@
...
+            "cpu_used_ram": 1000
+        }
+
+        self.mcmB = self._construct_model_config_measurement(
+            "modelB", self.model_specific_pa_params, mcmB_non_gpu_metric_values)
+
+    def tearDown(self):
+        NotImplemented
+
+    def test_init(self):
```
tgerdesnv, line 53: What do these all test? Add test intent comment to every test.

tests/test_model_config_measurement.py
```diff
@@ -0,0 +1,145 @@
...
+        mcmB_non_gpu_metric_values = {
+            "perf_throughput": 2000,
+            "perf_latency_p99": 40,
+            "cpu_used_ram": 1000
+        }
+
+        self.mcmB = self._construct_model_config_measurement(
+            "modelB", self.model_specific_pa_params, mcmB_non_gpu_metric_values)
+
+    def tearDown(self):
```
tgerdesnv, line 50: move to bottom of file (is this function necessary to implement? If not just remove)

model_analyzer/result/run_config_measurement.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def gpus_used(self):
+        """
+        Returns
+        -------
+        list of ints
+            list of device IDs used in this measurement
+        """
+
+        return list(self._gpu_data.keys())
```
tgerdesnv, line 254: will this ever return the same gpu multiple times?  Should this be a `set`?

model_analyzer/result/run_config_measurement.py
```diff
@@ -0,0 +1,373 @@
...
+            Metrics from the monitors that have a GPU UUID
+            associated with them            
+        """
+        self._key = key
+
+        self._gpu_data = gpu_data
+        self._avg_gpu_data = self._average_list(list(self._gpu_data.values()))
+
+        self._model_config_measurements = []
+        self._model_config_weights = []
```
tgerdesnv, line 52: Is there no default? Is it required to set this before doing any operations?  Should it be part of from_dict?

model_analyzer/result/run_config_measurement.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        Returns
+        -------
+        per model list of a list of Records
+            All the metric values in each model's measurement 
+            for both GPU and non-GPU
+        """
+
+        return [
+            self._avg_gpu_data + mcm.non_gpu_data()
```
tgerdesnv, line 133: remind me what the avg_gpu_data is for?

model_analyzer/result/model_config_measurement.py
```diff
@@ -0,0 +1,283 @@
...
+        Sets the metric weighting for this measurement based
+        on the objectives
+        
+        Parameters
+        ----------
+        metric_objectives : dict of RecordTypes
+            keys are the metric types, and values are the relative importance
+            of the keys with respect to each other
+        """
+        self._metric_weights = {
```
tgerdesnv, line 80: I do'nt see this defined in init. Also, are there default values? If not, are we requiring this function be called before calling other functions?  Also, should it be part of from_dict?

model_analyzer/result/model_config_measurement.py
```diff
@@ -0,0 +1,283 @@
...
+        self._model_specific_pa_params = model_specific_pa_params
+        self._non_gpu_data = non_gpu_data
+
+        self._non_gpu_data_from_tag = {
+            type(metric).tag: metric for metric in self._non_gpu_data
+        }
+
+    @classmethod
+    def from_dict(cls, model_config_measurement_dict):
+        model_config_measurement = ModelConfigMeasurement(None, {}, [])
```
tgerdesnv, line 56: Does it make sense to have a second constructor with no args that calls the constructor with these args? So the user doesn't have to know to pass these special empty variables?

model_analyzer/result/model_config_measurement.py
```diff
@@ -0,0 +1,283 @@
...
+            this ModelConfig measurement is better than the other
+        """
+
+        weighted_score = 0.0
+        for objective, weight in self._metric_weights.items():
+            self_metric = self.get_metric(tag=objective)
+            other_metric = other.get_metric(tag=objective)
+
+            # Handle the case where objective GPU metric is queried on CPU only
+            # TODO: I left this code here, but I'm not 100% sure that it is
```
tgerdesnv, line 270: We need to resolve this TODO

tgerdesnv (changes requested):

model_analyzer/result/model_config_measurement.py
```diff
@@ -0,0 +1,283 @@
...
+from functools import total_ordering
+import logging
+
+logger = logging.getLogger(LOGGER_NAME)
+
+
+@total_ordering
+class ModelConfigMeasurement:
+    """
+    Encapsulates the set of metrics obtained from a single model config's
```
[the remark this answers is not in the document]
nv-braf: Yes. I've updated the comment to indicate this.

tests/test_run_config_measurement.py
```diff
@@ -0,0 +1,264 @@
...
+class TestRunConfigMeasurement(trc.TestResultCollector):
+
+    def setUp(self):
+        self._construct_rcm0()
+        self._construct_rcm1()
+
+    def tearDown(self):
+        NotImplemented
+
+    def _construct_rcm0(self):
```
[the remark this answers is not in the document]
nv-braf: Done

tests/test_run_config_measurement.py
```diff
@@ -0,0 +1,264 @@
...
+                objective: value * self.weights[index] / sum(weights)
+                for (objective, value) in non_gpu_metric_values.items()
+            })
+
+        self.rcm1 = construct_run_config_measurement(
+            model_name, model_config_name, model_specific_pa_params,
+            gpu_metric_values, self.rcm1_non_gpu_metric_values,
+            metric_objectives, weights)
+
+    def test_key(self):
```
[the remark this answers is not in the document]
nv-braf: Done

tests/test_model_config_measurement.py
nv-braf: Done

tests/common/test_utils.py
```diff
@@ -29,6 +36,133 @@ def convert_to_bytes(string):
         return bytes(string + "    ", 'utf-8')
 
 
+def convert_non_gpu_metrics_to_data(non_gpu_metric_values):
```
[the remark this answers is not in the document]
nv-braf: Done

model_analyzer/result/run_config_measurement.py
nv-braf: The requirement is that this function is called before any comparisons are done. This is not a part of the dict as it is set during the analyze step (from_dict is only relevant for restoring values from the profile step).

The expectation is that model and metric weighting will be set before any comparisons are done.

model_analyzer/result/run_config_measurement.py
nv-braf: If running on multiple GPUs we average the GPU stats across the them and report this to the user.

model_analyzer/result/run_config_measurement.py
nv-braf: Dictionaries keys must be unique, so this is guaranteed to be unique

tests/test_model_config_measurement.py
nv-braf: It's not. I've left it here for documentation/completeness

model_analyzer/result/model_config_measurement.py
nv-braf: I'd like to leave this for a future investigation as I don't think it's blocking any work. If you're okay with that I'll file a story for this.

model_analyzer/result/run_config_measurement.py
tgerdesnv: "The expectation is that model and metric weighting will be set before any comparisons are done"

Is there any assert to protect this statement?

model_analyzer/result/model_config_measurement.py
tgerdesnv: sure, file a story. It basically just handles "None" case in comparisons -- if one is valid and the other is None then the valid one is "better". If both None then they are considered equal. It looks like we can get None results for some GPU metrics when CPU_ONLY is true. Seems like you could easily add a test with a None metric to verify this.

model_analyzer/result/run_config_measurement.py
nv-braf: No assert needed. We will just report that the run's are equivalent. See my change in MCM

model_analyzer/result/model_config_measurement.py
nv-braf: Makes sense. I've added testing to cover this and removed the TODO

tgerdesnv (approved):

PR #341, opened 2022-03-10
Change the default sort for online mode

tgerdesnv:
The old behavior was to strictly minimize latency in online mode. This led to almost random results being returned. Instead, we will now default to maximizing throughput.

Here are old results, where we were literally only looking at how far left (minimize latency) the first data point of each line was:
image

Here are new results, where we do something much more reasonable and maximize latency:
image

### DIFF
--- a/tests/test_report_manager.py
+++ b/tests/test_report_manager.py
@@ -275,7 +275,7 @@ class TestReportManagerMethods(trc.TestResultCollector):
         '''
         num_plots_in_summary_report = 2
         num_tables_in_summary_report = 1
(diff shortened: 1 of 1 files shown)


Tabrizian (approved):

Tabrizian:
We may need to update the example online report used in the repo to reflect this change.

tgerdesnv:
The format of the example is still correct. I do agree it would be nice to have a better representative of that report though. I plan on doing a deep dive and scrubbing our reports, at which point we would regenerate them

PR #354, opened 2022-03-31
Clean up logging

tgerdesnv:
Create format class to handle custom formatting, with extensive testing
- Error will print `[ModelAnalyzer] ERROR: <msg>`, and will show up in any mode
- Info will print  `[ModelAnalyzer] <msg>`, and will show up for normal or verbose mode
- Warning will print  `[ModelAnalyzer] WARNING: <msg>`, and will show up for normal or verbose mode
- Debug will print `[ModelAnalyzer] DEBUG: <msg>`, and will show up for verbose mode

Furthermore, in verbose mode, the time will be prepended to the message

Also:
- Added meaningful prints for details on model config creation as well as perf analyzer batch/concurrency values
- Downgraded some info prints to debug (like starting/stopping triton server over and over)
- Fixed a bug in results.py where measurement is missing
- Removed ellipses from messages

OLD:
552-old

NEW:
552-new1
552-new2

Example of new WARNING:
image

Example of verbose with time printed:
image


model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -109,14 +113,20 @@ def _make_direct_mode_model_config(self, param_combo):
...
+        logger.info(f"Creating model config: {model_config_dict['name']}")
+
         if param_combo is not None:
             for key, value in param_combo.items():
                 if value is not None:
                     model_config_dict[key] = value
-
-        model_config_dict['name'] = self._get_model_variant_name(param_combo)
+                    value_str = "ON" if value == {} else value
+                    logger.info(f"  Setting {key} to {value_str}")
```
nv-braf, line 125: A small nit, but instead of saying "Setting {key} to ON", you could do something like:
if value == {}:
    logger.info(f"  Enabling {key}")
else:
   logger.info(f"  Setting {key} to {value_str}")

model_analyzer/model_manager.py
```diff
@@ -83,6 +83,7 @@ def run_model(self, model):
                 measurement = self._metrics_manager.execute_run_config(
                     run_config)
             else:
+                logger.info("Skipping illegal run configuration")
```
nv-braf, line 86: Is it possible to print anything about why it is illegal? Would probably need to enhance is_legal_combination()

tgerdesnv: It is there already, but only shows up in verbose (it is a debug flag):
https://github.com/triton-inference-server/model_analyzer/blob/tgerdes-update-logging/model_analyzer/config/run/model_run_config.py#L100

nv-braf (dismissed):

model_analyzer/entrypoint.py
```diff
@@ -369,13 +350,13 @@ def main():
     Main entrypoint of model_analyzer
     """
 
-    # Configs and logging
-    logging.basicConfig(format="%(asctime)s.%(msecs)d %(levelname)-4s"
-                        "[%(filename)s:%(lineno)d] %(message)s",
-                        datefmt="%Y-%m-%d %H:%M:%S")
+    # Need to create a basic logging format for logs we print
+    # before we have enough information to configure the full logger
+    logging.basicConfig(format="[ModelAnalyzer] %(message)s")
```
Tabrizian, line 355: Would it be possible to use "Model Analyzer" instead of "ModelAnalyzer"?

Tabrizian (approved):
Thanks! The logs look much better now.

model_analyzer/record/metrics_manager.py
```diff
@@ -182,6 +182,8 @@ def execute_run_config(self, run_config):
         measurement = self._get_measurement_if_config_duplicate(
             model_run_config)
         if measurement:
+            logger.info(
+                "Existing measurement found for run config. Skipping profile")
```
debermudez, line 186: Thoughts on making this debug level?

debermudez (approved):

PR #362, opened 2022-04-06
Don't early exit concurrency search unless auto sweep mode

tgerdesnv:


Tabrizian (commented):
LGTM, do we need to handle the same thing for early stopping of max_batch_size field?

tgerdesnv:
> LGTM, do we need to handle the same thing for early stopping of max_batch_size field?

No, there are two explicit classes for sweeping model configs -- one for auto and one for manual. This was just an artifact of Perf Analyzer configuration generation being pre-computed for both auto and manual.

Tabrizian (approved):

PR #375, opened 2022-04-11
RunConfigResult unit testing

nv-braf:
Added single and multi-model unit testing to RunConfigResult class

- Single model checking for all methods including boundary condition constraint checking
- Multi-model (2) checking for relevant methods

### DIFF
--- a/tests/test_run_config_result.py
+++ b/tests/test_run_config_result.py
@@ -1,0 +1,68 @@
+# Copyright (c) 2022 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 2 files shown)


tests/test_run_config_result.py
```diff
@@ -0,0 +1,319 @@
...
+        self._construct_throughput_with_latency_constraint_rcr()
+
+    def tearDown(self):
+        NotImplemented
+
+    def test_model_name(self):
+        """
+        Test that model_name is correctly returned
+        """
+        self.assertEqual(self.rcr_empty.model_name(),
```
tgerdesnv, line 37: does this test anything? Those are just empty MagicMocks().  Seems like you would want to explicitly create an RCR with a name passed in to confirm the path from constructor -> 'get' function.  Same for run_config test below

tgerdesnv (changes requested):

tests/test_run_config_result.py
nv-braf: Yes, it still tests that the value generated by the MagicMock is returned correctly. You don't need to explicitly create a string to ensure the value is checked. For example, if I changed the model_name() return value to "Tim" I would get:

AssertionError: 'Tim' != <MagicMock id='140736174667856'>

tgerdesnv: Yeah I get that, and i agree it is testing that model_name() returns _model_name, but that is only half of the equation. I need to know that _model_name is the variable that actually stores the model name information.  I feel like normally you test the set() and get() functions together to prove it is connected all the way through. In this case, the 'set' is the constructor.

nv-braf: Fixed

tgerdesnv (approved):

PR #386, opened 2022-04-22
Set default model and metric weightings to RCM/MCM classes

nv-braf:
Add the following default values:

- All models equally weighted (in RunConfigMeasurement)
- ModelConfigs metrics are compared against using perf_throughput (in ModelConfigMeasurement)

Added an assertion to ensure that the correct number of model/metric weightings are set before any comparisons are done.
Updated failing unit tests.

### DIFF
--- a/model_analyzer/result/model_config_measurement.py
+++ b/model_analyzer/result/model_config_measurement.py
@@ -49,7 +49,8 @@ class ModelConfigMeasurement:
 
         self._non_gpu_data_from_tag = self._get_non_gpu_data_from_tag()
 
-        self._metric_weights = {}
+        # Set a default metric weighting
+        self._metric_weights = {"perf_throughput": 1}
 
     @classmethod
     def from_dict(cls, model_config_measurement_dict):
@@ -66,6 +67,8 @@ class ModelConfigMeasurement:
 
         model_config_measurement._non_gpu_data_from_tag = cls._get_non_gpu_data_from_tag(
             model_config_measurement)
+
+        model_config_measurement._metric_weights = {"perf_throughput": 1}
 
         return model_config_measurement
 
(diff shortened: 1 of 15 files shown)


model_analyzer/result/model_config_measurement.py
```diff
@@ -67,6 +68,8 @@ def from_dict(cls, model_config_measurement_dict):
         model_config_measurement._non_gpu_data_from_tag = cls._get_non_gpu_data_from_tag(
             model_config_measurement)
 
+        model_config_measurement._metric_weights = {"perf_throughput": 1}
```
tgerdesnv, line 71: Won't real values be in the dict?

model_analyzer/result/run_config_measurement.py
```diff
@@ -520,4 +530,6 @@ def _deserialize_model_config_measurements(
             model_config_measurements.append(
                 ModelConfigMeasurement.from_dict(mcm_dict))
 
+            self._model_config_weights.append(1)
```
tgerdesnv, line 533: won't this have real data?

tgerdesnv (changes requested):

model_analyzer/result/model_config_measurement.py
nv-braf: I've updated the code to restore these from the checkpoint and updated the L0's to use the new checkpoint that contains the default values.

model_analyzer/result/run_config_measurement.py
nv-braf: I've updated the code to restore these from the checkpoint and updated the L0's to use the new checkpoint that contains the default values.

tgerdesnv (approved):

PR #387, opened 2022-04-23
multi model fixes

tgerdesnv:


model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -111,7 +111,9 @@ def set_last_results(self, measurements):
 
         # Remove 'NONE' cases, and find single max measurement from the list
         measurements = [m for m in measurements if m]
-        measurement = [max(measurements)] if measurements else [None]
+
+        # TMA-634 - convert to max
```
nv-braf, line 115: In python, heapq, is implemented as min heap. We'd have to create our own max heap, which I don't think is a good use of our time. 
I think a better solution is invert the values going into the heap, and then we can invert the __lt__ method, which will allow the heap to still work and prevent use from needed to reverse min/max if your doing a measurement comparison outside of the RunConfigResultComparator class.

nv-braf (approved):

model_analyzer/config/generate/perf_analyzer_config_generator.py
tgerdesnv: I left this comment in TMA-634

PR #389, opened 2022-04-25
Fix assert when all results for a max_batch_size step were None

tgerdesnv:


model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -96,16 +96,10 @@ def _done_walking_instance_count(self):
     def _max_batch_size_limit_reached(self):
         return (self._curr_max_batch_size * 2) > self._max_model_batch_size
 
-    def _last_results_erroneous(self):
```
nv-braf, line 99: Why is this being removed? Is it just dead code?

tgerdesnv: Yes. Deadcode. I had started to use it in the solution but ran into a failure in a different unit test that showed it was the wrong way to go.

nv-braf (approved):

nv-braf (approved):

PR #390, opened 2022-04-25
Invert polarity of data stored in minheap

nv-braf:
The issue is that because we use a minheap, the `__lt__` methods in MCM/RCM have to return the opposite of the users expected value. This creates problems if the user does comparisons outside of the heap.

To solve this I inverted the data going "in to/out of" the heap. This allows me to reverse the __lt__ methods in MCM/RCM. I've fixed the usages in perf_config_generator and report_manager and updated comments/unit tests.

### DIFF
--- a/model_analyzer/reports/report_manager.py
+++ b/model_analyzer/reports/report_manager.py
@@ -421,7 +421,8 @@ class ReportManager:
                                         title="Report Table")
 
         sorted_measurements = sorted(self._summary_data[report_key],
-                                     key=lambda x: x[1])
+                                     key=lambda x: x[1],
+                                     reverse=True)
 
         # Construct summary sentence using best config
         best_config = sorted_measurements[0][0]
--- a/model_analyzer/result/run_config_result.py
+++ b/model_analyzer/result/run_config_result.py
@@ -119,7 +119,10 @@ class RunConfigResult:
             of RunConfigMeasurements in this RunConfigResult
         """
 
-        return self._measurements
+        return [
+            RunConfigMeasurement.invert_values(measurement)
+            for measurement in self._measurements
+        ]
 
     def passing_measurements(self):
         """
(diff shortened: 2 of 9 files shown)


model_analyzer/reports/report_manager.py
```diff
@@ -421,7 +421,8 @@ def _build_summary_table(self,
                                         title="Report Table")
 
         sorted_measurements = sorted(self._summary_data[report_key],
-                                     key=lambda x: x[1])
+                                     key=lambda x: x[1],
+                                     reverse=True)
```
tgerdesnv, line 425: Can you explain this? Why do we want reverse=True here? Because it sorts ascending by default and we want descending?

model_analyzer/result/run_config_result.py
```diff
@@ -100,12 +101,15 @@ def add_run_config_measurement(self, run_config_measurement):
             The profiled RunConfigMeasurement
         """
 
-        heapq.heappush(self._measurements, run_config_measurement)
+        inverted_rcm = RunConfigMeasurement.invert_values(
```
tgerdesnv, line 104: These changes seem fine and mostly self-contained, but I'm still wondering why we are using heapq and jumping through all these hoops. The amount of data we are sorting is trivial. Any reason we can't just use a sorted list?

tgerdesnv (changes requested):

model_analyzer/reports/report_manager.py
nv-braf: That's correct - previously, it was relying on the opposite polarity `__lt__`, to sort in the correct direction.

model_analyzer/result/run_config_result.py
nv-braf: I've switched to using bisect and remove the polarity inversion and the heap. @tgerdesnv please re-review

tgerdesnv (approved):

PR #392, opened 2022-04-27
Replace get metric value calls

nv-braf:
Replace all calls to get_metric/get_metric_value() with gpu/non-gpu specific methods.
Remove these methods from RunConfigMeasurement

### DIFF
--- a/model_analyzer/plots/detailed_plot.py
+++ b/model_analyzer/plots/detailed_plot.py
@@ -125,7 +125,7 @@ class DetailedPlot:
                 tag='perf_throughput'))
 
         for metric in self.detailed_metrics:
-            if MetricsManager.is_gpu_metric(tag):
+            if MetricsManager.is_gpu_metric(tag=metric):
                 self._data[metric].append(
                     run_config_measurement.get_gpu_metric_value(tag=metric))
             else:
--- a/model_analyzer/plots/detailed_plot.py
+++ b/model_analyzer/plots/detailed_plot.py
@@ -121,7 +121,8 @@ class DetailedPlot:
             ['concurrency-range'])
 
         self._data['perf_throughput'].append(
-            run_config_measurement.get_metric_value(tag='perf_throughput'))
+            run_config_measurement.get_non_gpu_metric_value(
+                tag='perf_throughput'))
 
         for metric in self.detailed_metrics:
             self._data[metric].append(
--- a/model_analyzer/plots/detailed_plot.py
+++ b/model_analyzer/plots/detailed_plot.py
@@ -125,8 +125,12 @@ class DetailedPlot:
                 tag='perf_throughput'))
 
         for metric in self.detailed_metrics:
-            self._data[metric].append(
-                run_config_measurement.get_metric_value(tag=metric))
+            if MetricsManager.is_gpu_metric(tag):
+                self._data[metric].append(
+                    run_config_measurement.get_gpu_metric_value(tag=metric))
+            else:
+                self._data[metric].append(
+                    run_config_measurement.get_non_gpu_metric_value(tag=metric))
 
     def plot_data(self):
         """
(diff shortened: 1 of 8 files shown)


model_analyzer/result/run_config_measurement.py
```diff
@@ -318,29 +294,7 @@ def get_gpu_metric_value(self, tag, default_value=0):
...
-            Average of the values of the metric Record corresponding 
-            to the tag, default_value if tag not found.
-        """
-        return mean([
-            default_value if m is None else m.value()
-            for m in self.get_metric(tag)
-        ])
-
-    def get_weighted_metric_value(self, tag, default_value=0):
+    def get_weighted_non_gpu_metric_value(self, tag, default_value=0):
```
tgerdesnv, line 297: please remove line 292 (`foo = self.get_gpu_metric(tag)`)

model_analyzer/plots/detailed_plot.py
```diff
@@ -121,11 +121,16 @@ def add_run_config_measurement(self, run_config_measurement):
...
 
         self._data['perf_throughput'].append(
-            run_config_measurement.get_metric_value(tag='perf_throughput'))
+            run_config_measurement.get_non_gpu_metric_value(
+                tag='perf_throughput'))
 
         for metric in self.detailed_metrics:
-            self._data[metric].append(
-                run_config_measurement.get_metric_value(tag=metric))
+            if MetricsManager.is_gpu_metric(tag=metric):
```
tgerdesnv, line 128: So the old function `get_metric_value` seemed to handle both gpu and non-gpu cases. Now we are splitting it out and making the user know which it needs to call. Is the reason because one returns a single value and the other returns a list?

model_analyzer/reports/report_manager.py
```diff
@@ -450,11 +450,12 @@ def _build_summary_table(self,
...
                     model_config.dynamic_batching_string(), instance_group_str,
-                    measurement.get_metric_value('perf_latency_p99'),
-                    measurement.get_metric_value('perf_throughput'),
-                    measurement.get_metric_value('cpu_used_ram'),
-                    measurement.get_metric_value('gpu_used_memory'),
-                    round(measurement.get_metric_value('gpu_utilization'), 1)
+                    measurement.get_non_gpu_metric_value('perf_latency_p99'),
+                    measurement.get_non_gpu_metric_value('perf_throughput'),
+                    measurement.get_non_gpu_metric_value('cpu_used_ram'),
+                    measurement.get_gpu_metric_value('gpu_used_memory'),
```
tgerdesnv, line 456: Some of these are lists, and some are scalar, right? Are we just fixing single model here and pushing the multi-model change to later story?

model_analyzer/result/model_config_measurement.py
```diff
@@ -133,7 +133,7 @@ def get_metric(self, tag):
             return self._non_gpu_data_from_tag[tag]
         else:
             logger.warning(
-                f"No metric corresponding to tag '{tag}' "
+                f"No non-GPU metric corresponding to tag '{tag}' "
```
tgerdesnv, line 136: Does this function only handle non-GPU because the GPU metrics are held at a higher level class (RunConfigMeasurement)?

tgerdesnv (changes requested):

model_analyzer/plots/detailed_plot.py
nv-braf: That's correct.

model_analyzer/reports/report_manager.py
nv-braf: Correct.

model_analyzer/result/model_config_measurement.py
nv-braf: Yes.

model_analyzer/plots/detailed_plot.py
tgerdesnv: I'm curious how the functions will be used later on. It would be ideal if all of the gpu vs non-gpu information was still hidden from the user so they didn't have to have that knowledge. We can cross that bridge when you make the changes to support multi-model, but maybe get_metric_value returns with an optional "model" arg that returns the single scalar result of the given tag from that model?

tgerdesnv (approved):

PR #393, opened 2022-04-27
consistent model variant naming

tgerdesnv:
Make sure that if the same configuration is reused, it will have the same name. This is especially useful for multi-model.

### DIFF
--- a/model_analyzer/config/generate/model_variant_name_manager.py
+++ b/model_analyzer/config/generate/model_variant_name_manager.py
@@ -14,15 +14,10 @@
 
 import collections.abc
 from copy import deepcopy
+from model_analyzer.constants import DEFAULT_CONFIG_PARAMS
 
 
 class ModelVariantNameManager:
-
-    # FIXME
-    # Dict of parameters to apply on top of the default config to result
-    # in the default config (none)
-    #
-    DEFAULT_PARAM_COMBO = {}
 
     def __init__(self):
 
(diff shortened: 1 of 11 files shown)


model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -0,0 +1,115 @@
...
+        # Stored as _model_name_index[base_model_name] = current_count_integer
+        self._model_name_index = {}
+
+    def get_model_variant_name(self, base_model_name, param_combo):
+        """
+        Given a base model name and a dict of parameters to be applied
+        to the base model config, return the name of the model variant
+
+        If the same input values are provided to this function multiple times, 
+        the same value should be returned
```
nv-braf, line 36: Change to: "value **will** be returned"
There are no cases where this isn't true.

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -0,0 +1,115 @@
...
+
+        return self._get_hashable_key_helper(obj_copy)
+
+    def _get_hashable_key_helper(self, obj):
+        """
+        Given a list or dict which may have nested dicts/lists, return a key that 
+        is hashable 
+        """
+
+        if isinstance(obj, collections.abc.Hashable):
```
nv-braf, line 105: This is "magic" to me. If I'm understanding this correctly, we keep recursively calling until we get to something that's hashable?

tgerdesnv: That is correct. In order to make a list or dict hashable, first we need every element inside of it to be hashable. We recursively go down to the leaves of all elements and make them hashable, and then work our way back up to the top until we have a single hashable result.

model_analyzer/config/generate/model_variant_name_manager.py
tgerdesnv: fixed

PR #399, opened 2022-05-02
Fixing local build issue for preinstalled GPG key in base image

mc-nv:
I removed that part with adding package manager for `/etc/apt/sources.list` because it's reside in base image: 
```
docker run -it nvcr.io/nvidia/tritonserver:22.04-py3 cat /etc/apt/sources.list | grep nvidia
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license
   https://docs.nvidia.com/datacenter/cloud-native/ .
deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /
# deb-src https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /
```
Particular change in L33 is also temporary in such case and can be removed next to release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ ENV DCGM_VERSION=2.2.9
 # Ensure apt-get won't prompt for selecting options
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


tgerdesnv (approved):
I assume this passed CI?

matthewkotila (changes requested):
Is there a CI pipeline that shows that these changes resolved the apt update error we previously saw?

mc-nv:
> 

CI passed

matthewkotila (approved):

PR #401, opened 2022-05-03
Stop analyze/report if no/bad checkpoint

nv-braf:
These changes will cause an exception to be raised if we detect a missing checkpoint or if the checkpoint does not contain the model(s) specified by the user on the CLI.

Here is the new output if no checkpoint exists:
Capture4

Here is the new output if the checkpoint exists, but the model is not found:
Capture3

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -41,7 +41,7 @@ class Analyzer:
     result writing methods.
     """
 
-    def __init__(self, config, server, state_manager, run_without_checkpoint):
+    def __init__(self, config, server, state_manager, checkpoint_required):
         """
         Parameters
         ----------
@@ -51,12 +51,14 @@ class Analyzer:
             Server handle
         state_manager: AnalyzerStateManager
             The object that maintains Model Analyzer State
+        checkpoint_required : bool
+            If true, an existing checkpoint is required to run MA
         """
 
         self._config = config
         self._server = server
         self._state_manager = state_manager
-        state_manager.load_checkpoint(run_without_checkpoint)
+        state_manager.load_checkpoint(checkpoint_required)
 
         self._result_manager = ResultManager(config=config,
                                              state_manager=self._state_manager)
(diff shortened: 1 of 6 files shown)


model_analyzer/analyzer.py
```diff
@@ -41,7 +41,7 @@ class Analyzer:
     result writing methods.
     """
 
-    def __init__(self, config, server, state_manager):
+    def __init__(self, config, server, state_manager, run_without_checkpoint):
```
tgerdesnv, line 44: run_without_checkpoint implies that we are running without a checkpoint, not that we are allowed to run without a checkpoint.

Please change it to `checkpoint_required`

tgerdesnv (changes requested):

model_analyzer/analyzer.py
tgerdesnv: and add the new parameter to the parameters documentation on line ~54

model_analyzer/state/analyzer_state_manager.py
```diff
@@ -107,7 +107,7 @@ def set_state_variable(self, name, value):
         self._state_changed = True
         self._current_state.set(name, value)
 
-    def load_checkpoint(self):
+    def load_checkpoint(self, run_without_checkpoint):
```
tgerdesnv, line 110: add the parameter to the doc

tgerdesnv (changes requested):

model_analyzer/state/analyzer_state_manager.py
```diff
@@ -129,7 +134,7 @@ def load_checkpoint(self, run_without_checkpoint):
                         ' directory.')
             self._starting_fresh_run = False
         else:
-            if run_without_checkpoint:
+            if checkpoint_required:
```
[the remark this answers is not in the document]
nv-braf: Good point. I've reversed the polarity

tgerdesnv (approved):

PR #408, opened 2022-05-09
Fix inconsistent results from top N analysis

nv-braf:
The issue is that the MCM's weighted score for multi-model uses self's value as the denominator to determine % gain/loss. While I believe this is technically correct. The issue arises when doing comparisons between multiple models. See this example:

A: config_0,config_1 = {336, 223}
B: config_2,config_3 = {270, 272}

If you evaluate for A.is_better_than(B) you get {66/336, -49/223} = {19.6% gain, 22% loss} = B is better (by 2.4%)
If you evaluate for B.is_better_than(A) you get {-66/270, 49/272} = {24.4% loss, 18% gain} = A is better (by 6.4%)

And, since we can't control which way to comparison occurs in the minheap sort this leads to inconsistent sorting results for model configs that have close results.

To solve this I've replaced the denominator with the mean of the two models, now if you evaluate you get:

If you evaluate for A.is_better_than(B) you get {66/303, -49/248} = {21.8% gain, 19.8% loss} = A is better (by 2%)
If you evaluate for B.is_better_than(A) you get {-66/303, 49/248} = {21.8% loss, 19.8% gain} = B is worse (by 2%)

### DIFF
--- a/model_analyzer/result/model_config_measurement.py
+++ b/model_analyzer/result/model_config_measurement.py
@@ -17,6 +17,7 @@ from model_analyzer.constants import LOGGER_NAME
 
 from model_analyzer.record.record import RecordType
 
+from statistics import mean
 from functools import total_ordering
 import logging
 
@@ -285,8 +286,8 @@ class ModelConfigMeasurement:
                 return 0
 
             metric_diff = self_metric - other_metric
-            weighted_score += (weight *
-                               (metric_diff.value() / self_metric.value()))
+            weighted_score += (weight * (metric_diff.value() / mean(
+                [self_metric.value(), other_metric.value()])))
 
         return weighted_score
 
--- a/tests/test_run_config_measurement.py
+++ b/tests/test_run_config_measurement.py
@@ -34,6 +34,8 @@ class TestRunConfigMeasurement(trc.TestResultCollector):
     def setUp(self):
         self._construct_rcm0()
         self._construct_rcm1()
+        self._construct_rcm2()
+        self._construct_rcm3()
 
     def tearDown(self):
         NotImplemented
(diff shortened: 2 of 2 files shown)


model_analyzer/result/model_config_measurement.py
```diff
@@ -285,8 +286,8 @@ def _calculate_weighted_score(self, other):
                 return 0
 
             metric_diff = self_metric - other_metric
-            weighted_score += (weight *
-                               (metric_diff.value() / self_metric.value()))
+            weighted_score += (weight * (metric_diff.value() / mean(
+                [self_metric.value(), other_metric.value()])))
```
tgerdesnv, line 290: this line is so hard to parse. I know what it is supposed to be doing and I still can't manage to line up all the brackets and parenthesis in my head.  Can you refactor it a bit? Maybe pull the mean code out just like `metric_diff` was?

tgerdesnv (changes requested):

model_analyzer/result/model_config_measurement.py
```diff
@@ -285,8 +286,8 @@ def _calculate_weighted_score(self, other):
                 return 0
 
             metric_diff = self_metric - other_metric
-            weighted_score += (weight *
-                               (metric_diff.value() / self_metric.value()))
+            avgerage = mean([self_metric.value(), other_metric.value()])
```
tgerdesnv, line 289: avgerage? is that a typo?

tgerdesnv (changes requested):

model_analyzer/result/model_config_measurement.py
nv-braf: Typo + auto-complete!

tgerdesnv (approved):

PR #412, opened 2022-05-12
Remove cpu ram from reports when not collected

tgerdesnv:
- If no CPU metrics, do not print "Max CPU Memory Usage" to table for both detailed and summary reports
- If no CPU metrics, do not print "RAM usage vs latency" plot for both detailed and summary reports
- Fixed a bug where "cpu_only" was not put into checkpoint properly, and thus the GPU plots were printed for CPU cases
- Reduce the number of times the CPU warning is printed

I manually compared the resulting reports of all 16 combinations of the following {detailed vs summary}{online vs offline}{model run on cpu vs gpu}{cpu metrics being collected vs not} and confirmed all changes are as expected.

Showing the table and chart no longer being printed when CPU info not gathered:

412_compare

Showing the cpu_only bug fix that now correctly shows CPU information and charts instead of GPU:

412_bug_fix

Before and after of warning changes:

412-warnings
412-less-warnings

### DIFF
--- a/model_analyzer/reports/report_manager.py
+++ b/model_analyzer/reports/report_manager.py
@@ -881,8 +881,9 @@ class ReportManager:
             used_ram = None
             if self._detailed_report_data:
                 key = list(self._detailed_report_data.keys())[0]
-                _, measurement = self._detailed_report_data[key][0]
-                used_ram = measurement.get_non_gpu_metric_value('cpu_used_ram')
+                _, measurements = self._detailed_report_data[key]
+                used_ram = measurements[0].get_non_gpu_metric_value(
+                    'cpu_used_ram')
             else:
                 key = list(self._summary_data.keys())[0]
                 _, measurement = self._summary_data[key][0]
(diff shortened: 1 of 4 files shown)


model_analyzer/reports/report_manager.py
```diff
@@ -845,3 +886,22 @@ def _build_constraint_strings(self):
...
+                key = list(self._detailed_report_data.keys())[0]
+                _, measurements = self._detailed_report_data[key]
+                used_ram = measurements[0].get_non_gpu_metric_value(
+                    'cpu_used_ram')
+            else:
+                key = list(self._summary_data.keys())[0]
+                _, measurement = self._summary_data[key][0]
+                used_ram = measurement.get_non_gpu_metric_value('cpu_used_ram')
+
+            # FIXME I would like to use used_ram is not None, and pass None as
```
tgerdesnv, line 903: @nv-braf

debermudez (dismissed):

model_analyzer/reports/report_manager.py
tgerdesnv: It works as-is. I'll remove the FIXME

debermudez (approved):

PR #414, opened 2022-05-17
Multi model result manager updates and unit testing enchancements

nv-braf:
Added two golden models to `tests/common` , for single and multi-model with READMEs explaining how to regenerate them.
Added unit testing to result manager to check that the 3 metric CSVs (server, gpu, inference) created during analysis match the golden CSVs.

Refactored and updated code (tablulate_measurements() and its children) to work with multi-model. This is what the new format looks like (for inference) (for the others look at the golden CSVs that were created for this story):
image

Also added `patch.stopall()` to all unit test teardowns to ensure that any mock methods are properly destroyed when the unit test completes (was causing issues when trying to read-in the golden CSVs)


model_analyzer/result/result_manager.py
```diff
@@ -24,6 +24,10 @@
 from .run_config_result import RunConfigResult
 from .results import Results
 
+from statistics import mean
```
tgerdesnv, line 27: not needed?

model_analyzer/result/result_manager.py
```diff
@@ -480,133 +487,195 @@ def _tabulate_measurements(self, run_config_result):
...
+        concurrencies = [
+            pa_params['concurrency-range']
+            for pa_params in model_specific_pa_params
+        ]
+
+        return model_specific_pa_params, batch_sizes, concurrencies
+
+    def _populate_inference_rows(self, run_config_measurement, inference_fields,
+                                 inference_row):
+        # Just using this to get a list of tags, so only need to reference the first model
```
tgerdesnv, line 594: is it possible that models have different tags? If one model is cpu_only and the other isn't?

model_analyzer/result/run_config_measurement.py
```diff
@@ -293,7 +293,7 @@ def get_gpu_metric_value(self, tag, default_value=0):
         return default_value if self.get_gpu_metric(
             tag) is None else self.get_gpu_metric(tag).value()
 
-    def get_weighted_non_gpu_metric_value(self, tag, default_value=0):
+    def get_weighted_metric_value(self, tag, default_value=0):
```
tgerdesnv, line 296: is there no longer a difference between non-gpu metrics and gpu-metrics?

tests/test_result_manager.py
```diff
@@ -25,9 +27,85 @@
...
+    def test_single_model_csv_against_golden(self):
+        """
+        Match the csvs against the golden versions in
+        tests/common/single-model-ckpt
+        """
+        self._single_model_result_manager.tabulate_results()
+        self._single_model_result_manager.export_results()
+
+        self.assertTrue(
+            cmp(f"{ROOT_DIR}/single-model-ckpt/results/metrics-model-gpu.csv",
```
tgerdesnv, line 53: have you locally tried to edit the csv to a different value to confirm that this will fail?

tgerdesnv (changes requested):

model_analyzer/result/result_manager.py
nv-braf: Removed

tests/test_result_manager.py
nv-braf: No, but I have tested it the other way (modifying the new csv being outputted) and have observed the test case failing, so I have proof this is working as intended/

model_analyzer/result/run_config_measurement.py
nv-braf: Good catch. This shouldn't have changed names (I originally thought I needed to, but turned out not to be the case). I've reverted this name.

tgerdesnv:
does it make sense to show a 'combined' (averaged) p99 latency?  Mathematically I'm not sure that makes sense.

model_analyzer/result/result_manager.py
nv-braf: Yes, probably. I'll file a story to figure out how we want to deal with this.

nv-braf: TMA-686 filed

tgerdesnv:
> does it make sense to show a 'combined' (averaged) p99 latency? Mathematically I'm not sure that makes sense.

We can wait until we get feedback for what this should look like

PR #416, opened 2022-05-18
Add experiments folder

tgerdesnv:
... with ability to evaluate config generator algorithms offline. The runtime for the existing RunConfigGenerator algorithm on the resnet database is 3 seconds.

Here we can see it compare our existing algorithm with itself:
image

Here are results if I artificially limit the search space to max_batch_size=2, max_instance_count=2, max_concurrency=2:
image


experiments/profile_data.py
```diff
@@ -0,0 +1,129 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+
+
+class ProfileData:
```
nv-braf, line 18: This comment applies to the whole class:
My preference is that we do not use the generic terms config/measurement. For clarity, I think it pays to be a little more verbose and call these what they are run_config/run_config_measurement. 
This would apply to method names, arguments, and internal variables.

experiments/profile_data.py
```diff
@@ -0,0 +1,129 @@
...
+        self._add_measurement_from_keys(keys, measurement)
+
+    def get_measurement(self, run_config):
+        """ 
+        Get the measurement belonging to a given run config
+        """
+        keys = self._extract_run_config_keys(run_config)
+        return self._get_measurement_from_keys(keys)
+
+    def get_config_count(self):
```
nv-braf, line 43: change this to get_model_config_count

experiments/profile_data.py
```diff
@@ -0,0 +1,129 @@
...
+        """ 
+        Get the total number of model configs in the data
+        """
+        count = 0
```
nv-braf, line 47: Isn't this just `return len(self._data.keys())`?

experiments/profile_data.py
```diff
@@ -0,0 +1,129 @@
...
+            count += 1
+        return count
+
+    def get_measurement_count(self):
+        """ 
+        Get the total number of measurements in the data
+        """
+        count = 0
+        for ma_key in self._data.keys():
+            for pa_key in self._data[ma_key].keys():
```
nv-braf, line 58: Remove the inner loop and replace with `count += len(self._data[ma_key].keys())`

experiments/profile_data.py
```diff
@@ -0,0 +1,129 @@
...
+        Get the config corresponding to the best overall measurement in the data
+        """
+        return self._best_config
+
+    def _add_measurement_from_keys(self, keys, measurement):
+        self._update_best_trackers(keys, measurement)
+
+        curr_dict = self._data
+
+        for i in range(len(keys) - 1):
```
nv-braf, line 79: change this to `for i, key in enumerate(keys)` and replace `keys[i]` with `key`

experiments/profile_data.py
```diff
@@ -0,0 +1,129 @@
...
+            if keys[i] not in curr_dict:
+                curr_dict[keys[i]] = {}
+            curr_dict = curr_dict[keys[i]]
+        curr_dict[keys[-1]] = measurement
+
+    def _update_best_trackers(self, keys, measurement):
+        if not self._best_measurement or measurement > self._best_measurement:
+            self._best_measurement = measurement
+            self._best_config = keys[0]
```
nv-braf, line 88: keys[0] = model_config - would be better to make to define this in the class and get rid of the magic number
also this will need to change to work for multi-model

See results.py for an example of how I handled this

experiments/profile_data.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def _get_measurement_from_keys(self, keys):
+        curr_dict = self._data
+
+        for i in range(len(keys) - 1):
```
nv-braf, line 93: change this to `for key in keys:` and replace `keys[i]` with `key`

nv-braf (changes requested):

experiments/profile_data.py
nv-braf: This doesn't have anything to do with file names - it has to do with accurately representing the data/class that is being used/passed in. Due to multi-model, we have different types of measurements: run_config and model_config, and I just think we need to be explicit with our naming (even though I know that measurement will refer to run_config_measurement 99% of the time).

experiments/profile_data.py
tgerdesnv: yes, I used to have a nested n-dimention dictionary that I later cleaned up to always just be 2 layers deep. I was lazy and kept the code because I knew it would still work ;)

experiments/profile_data.py
tgerdesnv: I updated it to be explicit for the fixed 2-layer dict that it now is

experiments/profile_data.py
tgerdesnv: no longer using keys[0] or keys[1] anywhere.

experiments/profile_data.py
tgerdesnv: cleaned up to be explicitly 2 layer

experiments/evaluate_config_generator.py
```diff
@@ -0,0 +1,62 @@
...
+
+        generator, patches = GeneratorExperimentFactory.get_generator_and_patches(
+            generator_name, self._profile_config)
+
+        for patch in patches:
+            patch.start()
+
+        self._run_generator(generator)
+
+        for patch in patches:
```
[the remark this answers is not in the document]
tgerdesnv: hmm, if that works then I don't need to explicitly pass the patches back to the caller. I'll rework it so the patching happens in the factory

experiments/checkpoint_data.py
```diff
@@ -0,0 +1,45 @@
...
+    def _load_checkpoint(self, config):
+        state_manager = AnalyzerStateManager(config, MagicMock())
+        state_manager.load_checkpoint(checkpoint_required=True)
+
+        results = state_manager.get_state_variable('ResultManager.results')
+        model_name = config.profile_models[0].model_name()
+        model_measurements = results.get_model_measurements_dict(model_name)
+        for (run_config,
+             run_config_measurements) in model_measurements.values():
+            (ma_key, pa_key) = self._extract_run_config_keys(run_config)
```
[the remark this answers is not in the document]
tgerdesnv: good catch. This is intentional, but I will add a comment.  It is a function of how we store results in MA

experiments/evaluate_config_generator.py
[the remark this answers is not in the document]
tgerdesnv: actually the only things patched will be generator-specific and are safe to leave patched

experiments/profile_config.py
```diff
@@ -0,0 +1,66 @@
...
+            '--model-repository', data_path,
+            '--checkpoint-directory', checkpoint_dir,
+            '-f', 'path-to-config-file'
+        ]
+
+        yaml_content = ""
+
+        # Could overload yaml data here. May want to expose this
+        #yaml_content = convert_to_bytes("""
+        #    run_config_search_max_concurrency: 2
```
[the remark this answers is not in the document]
tgerdesnv: I'll add a TODO. I think we are going to need the ability to override this or pass some kind of other configuration to the generator

experiments/profile_data.py
tgerdesnv: I've cleaned up the variable names and have also renamed the classes

experiments/profile_data.py
nv-braf: Yes, I see that, but I'm not sure what you've done is better/cleaner. You still need to have this implicit understanding that keys = {ma_key, pa_key} which is not documented anywhere outside of the method that actually creates the keys. 

It would be cleaner to just have ma_key and pa_key be separate variables and pass them in separately to the methods that need them.

nv-braf (approved):
Look good (in reference to the multi-model changes). I did not mean to approve the review, but can't figure out how to undo this action.

nv-braf (approved):
Approved

PR #417, opened 2022-05-19
Plot manager unit testing

nv-braf:
I've added the plot_manager output to the single/multi-model checkpoint directories.

We then load this in during the unit test and compare the data and fields against this golden version. This should ensure that the plot outputs will match.

Note that I'm using pickle to write/read the class (as this saves me from having to create to/from dict methods for this class and the subclasses). I believe this is safe this is only done during unit testing, but we should confirm this won't get flagged by security during a release before committing.


tests/test_plot_manager.py
```diff
@@ -0,0 +1,177 @@
...
+
+from model_analyzer.plots.plot_manager import PlotManager
+from model_analyzer.result.result_manager import ResultManager
+from model_analyzer.state.analyzer_state_manager import AnalyzerStateManager
+
+from filecmp import cmp
+from shutil import rmtree
+from unittest.mock import MagicMock, patch
+
+import pickle
```
debermudez, line 33: https://medium.com/ochrona/python-pickle-is-notoriously-insecure-d6651f1974c9#:~:text=Pickle%20is%20dangerous%20because%20during,__%20method%20to%20be%20executed.  

bah! we are likely getting hit with this vulnerability even if its just test code.
Is there a different serialization/deserialization method we can use?

tests/test_plot_manager.py
```diff
@@ -0,0 +1,177 @@
...
+    def _compare_plot_data(self, golden, plot):
+        self.assertEqual(golden._name, plot._name)
+        self.assertEqual(golden._title, plot._title)
+        self.assertEqual(golden._x_axis, plot._x_axis)
+        self.assertEqual(golden._y_axis, plot._y_axis)
+        self.assertEqual(golden._monotonic, plot._monotonic)
+        self.assertEqual(golden._data, plot._data)
+
+        # These are generated by pyplot and will vary from run to run
+        #self.assertEqual(golden._fig, plot._fig)
```
debermudez, line 116: if these aren't comparable due to randomness, remove the commented out code from testing.

debermudez (changes requested):

tests/test_plot_manager.py
nv-braf: I think we're actually safe here. Taken from the article you linked:
`Utilize pickle, but only with strict controls in place such as only losing pickles that you created yourself, or only loading pickles from trusted parties where cryptographic hashes can be used to verify that the source is authentic and that the pickle hasn’t been modified.`

In this case I've hardcoded the file we are loading into the unit test (unlike say a checkpoint file that a user would provide) and saved it in the repo.

tests/test_plot_manager.py
nv-braf: This is documentation so that it doesn't look like I omitted comparing part of the classes data.

debermudez: oic. cool cool cool

tests/test_plot_manager.py
rmccorm4: It looks like _compare_plot_data is only comparing simple primitives, so I don't think it would be too difficult to compare to a JSON serialization or similar to avoid pickle usage altogether if concerned.

tests/test_plot_manager.py
```diff
@@ -0,0 +1,177 @@
...
+            result_manager=self._multi_model_result_manager)
+
+        plot_manager.create_summary_plots()
+
+        # Uncomment these lines to create a new output dump to compare against
+        # file = open(f"{ROOT_DIR}/multi-model-ckpt/plot_manager.pkl", "wb")
+        # pickle.dump(plot_manager._simple_plots, file)
+        # file.close()
+
+        file = open(f"{ROOT_DIR}/multi-model-ckpt/plot_manager.pkl", "rb")
```
rmccorm4, line 91: It's more pythonic to use [context managers](https://book.pythontips.com/en/latest/context_managers.html#context-managers) for this:
```
with open(filename, "rb"):
  golden_plot_manager_simple_plots = ...
```
# file automatically closed outside of context manager scope

tests/test_plot_manager.py
nv-braf: The issue is that the SimplePlot class is not JSON serializable. So, this doesn't work:
`json.dumps(plot_manager._simple_plots['add_sub']
                         ['throughput_v_latency'].__dict__)`
                        
The issue is the matlab.pyplot being stored in the SimplePlot.

rmccorm4: Right, but if you're hard-coding a golden file, you don't need to dump the object, you can manually construct/print a dict once.

Or if you implement the __dict__, you can exclude the non-serializable objects like the plots if not required (plots aren't being compared in your comparison)

nv-braf: I've replaced pickle with json

tests/test_plot_manager.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def _compare_plot_data(self, golden, plot):
+        self.assertEqual(golden._name, plot._name)
+        self.assertEqual(golden._title, plot._title)
+        self.assertEqual(golden._x_axis, plot._x_axis)
+        self.assertEqual(golden._y_axis, plot._y_axis)
+        self.assertEqual(golden._monotonic, plot._monotonic)
+        self.assertEqual(golden._data, plot._data)
+
+        # These are generated by pyplot and will vary from run to run
```
tgerdesnv, line 123: These comments can go away? They aren't part of the structure anymore?

model_analyzer/plots/plot_manager.py
```diff
@@ -53,6 +53,38 @@ def __init__(self, config, result_manager):
         self._simple_plots = defaultdict(list)
         self._detailed_plots = {}
 
+    def to_dict(self):
```
tgerdesnv, line 56: I prefer that these aren't in the production code. This code is only for testing and should be able to be hid in the test code

tests/test_plot_manager.py
```diff
@@ -0,0 +1,185 @@
...
+        Match the summary plots against the golden versions in
+        tests/common/multi-model-ckpt
+        """
+        plot_manager = PlotManager(
+            config=self._multi_model_config,
+            result_manager=self._multi_model_result_manager)
+
+        plot_manager.create_summary_plots()
+
+        # Uncomment these lines to create a new output dump to compare against
```
tgerdesnv, line 90: Should the code after this be commented out?

tgerdesnv (changes requested):

tests/test_plot_manager.py
nv-braf: Yes, good catch.

tests/test_plot_manager.py
nv-braf: Yes. I've removed this.

model_analyzer/plots/plot_manager.py
nv-braf: I've moved this into the test class.

debermudez (approved):

PR #419, opened 2022-05-20
Add max batch size to inference csv

tgerdesnv:
Max Batch Size is now one of the default columns printed in the output inference csv file in the analysis step


nv-braf (approved):
Looks good!

PR #424, opened 2022-05-24
Neighborhood class

tgerdesnv:


model_analyzer/config/generate/neighborhood.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def _convert_to_unit_vector(self, vector):
+        magnitude = 0
+        for v in vector:
+            magnitude += math.pow(v, 2)
+        magnitude = math.sqrt(magnitude)
+
+        # Convert the vector to unit vector
+        if magnitude == 0:
+            # FIXME
```
tgerdesnv, line 187: not quite sure what to do here. Original code returned None, and then handled it outside

tests/test_neighborhood.py
```diff
@@ -0,0 +1,185 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
nv-braf, line 1: Fix copyright date

model_analyzer/config/generate/neighborhood.py
```diff
@@ -0,0 +1,192 @@
...
+            distance += math.pow(diff, 2)
+        distance = math.sqrt(distance)
+        return distance
+
+    def get_num_initialized_points(self):
+        """ 
+        Returns the number of coordinates in the neighborhood that have a throughput
+        associated with it
+        """
+        num_initialized = 0
```
nv-braf, line 66: This can be done easily with list comprehension: 
 return sum([self._coordinate_data.get_throughput(coordinate) for coordinate in self.neighborhood])

nv-braf (commented):
Overall, things look good. But, I didn't do a deep dive and trust that the math is being done correctly.

model_analyzer/config/generate/neighborhood.py
```diff
@@ -0,0 +1,192 @@
...
+        self._neighborhood = self._create_neighborhood()
+
+    @classmethod
+    def calc_distance(cls, coordinate1, coordinate2):
+        """ 
+        Return the euclidean distance between two coordinates
+        """
+
+        distance = 0
+        for i, v in enumerate(coordinate1):
```
debermudez, line 55: if the value isn't being used, can we make this underscore?

model_analyzer/config/generate/neighborhood.py
debermudez: does sum handle the 'None ' case

model_analyzer/config/generate/neighborhood.py
```diff
@@ -0,0 +1,192 @@
...
+            coordinates, throughputs)
+
+        vector = throughput_center - coordinate_center
+
+        unit_vector = self._convert_to_unit_vector(vector)
+        return unit_vector
+
+    def _compile_neighborhood_results(self):
+        coordinates = []
+        results = []
```
debermudez, line 146: why not call this throughputs?

tgerdesnv: Fixed.

model_analyzer/config/generate/neighborhood.py
tgerdesnv: if magnitude is 0, then vector was all 0's.  I'll just return that vector.

nv-braf (approved):

debermudez (approved):

PR #438, opened 2022-06-15
Expose experiment knobs

tgerdesnv:
Added ability to set algorithm-specific knobs on the experiments command line.

For example, standard profile option of max model batch size can be specified:
`python3 main.py --model-name resnet50_libtorch --generator RunConfigGenerator --run-config-search-max-model-batch-size 2`

Also, new options for Undirected algorithm can be set:
`python3 main.py --model-name resnet50_libtorch --generator UndirectedRunConfigGenerator --magnitude 5`

### DIFF
--- a/experiments/main.py
+++ b/experiments/main.py
@@ -16,8 +16,8 @@
 #
 # Example usage:
 #
-# python3 main.py --model-name resnet50_libtorch --generator RunConfigGenerator
-#
+# python3 main.py --model-name resnet50_libtorch --generator RunConfigGenerator --run-config-search-max-model-batch-size 2
+# python3 main.py --model-name resnet50_libtorch --generator UndirectedRunConfigGenerator --magnitude 5
 #####################
 
 from evaluate_config_generator import EvaluateConfigGenerator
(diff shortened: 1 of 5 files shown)


experiments/main.py
```diff
@@ -43,14 +43,14 @@
                     type=str,
                     required=True,
                     help="The name of the config generator to evaluate")
-args = parser.parse_args()
+args, other_args = parser.parse_known_args()
```
debermudez, line 46: Why use `parse_known_args` here? From this main.py file's perspective, you will be using the configCommand subclass that understands all of the additional options.   
Is this future proofing if you create new ConfigCommand subclasses that dont all overlap?

tgerdesnv: I'm not using configCommand subclass in main.py.  Main is using the ArgumentParser defined right above this line which only knows about a couple options that are specific to this script.

I do agree that using configCommand or something similar in main.py would be the cleaner more robust option. If this was production code I would have done that. I was just trying to cobble together something that allowed me to expose both old and new configuration options via the command-line.

debermudez: cool. thanks for explaining.

debermudez (approved):

PR #445, opened 2022-06-27
New L0 test for quick algo

tgerdesnv:


model_analyzer/config/generate/undirected_run_config_generator.py
```diff
@@ -246,12 +246,11 @@ def _get_next_perf_analyzer_config(self, model_variant_name, model_num):
 
         perf_analyzer_config = PerfAnalyzerConfig()
 
+        perf_analyzer_config.update_params_from_profile_config(
```
nv-hwoo, line 249: nit: maybe `update_config_...` is more suitable prefix since we are updating config?

tgerdesnv: I agree. Fixed.

nv-hwoo (approved):

PR #446, opened 2022-06-30
Fix bug and add testing for setting metrics-interval-ms

tgerdesnv:
There were two bugs in setting of metrics-interval-ms:
1: the triton config class in MA didn't recognize metrics-interval-ms as a valid option
2: the value was being set when we were using local_gpu_monitor, when it should have been set when we were NOT using local_gpu_monitor

Fixed those bugs and added unit testing around the TritonServerFactory

### DIFF
--- a/tests/test_triton_server_factory.py
+++ b/tests/test_triton_server_factory.py
@@ -55,14 +55,15 @@ class TestTritonServerFactory(trc.TestResultCollector):
 
         config = ConfigCommandProfile()
         config.triton_launch_mode = launch_mode
-        config.triton_http_endpoint = "1:2"
-        config.triton_grpc_endpoint = "3:4"
+        config.triton_http_endpoint = "fake_address:2345"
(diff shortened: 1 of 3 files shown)


tests/test_triton_server_factory.py
```diff
@@ -0,0 +1,147 @@
...
+        config = ConfigCommandProfile()
+        config.triton_launch_mode = launch_mode
+        config.triton_http_endpoint = "1:2"
+        config.triton_grpc_endpoint = "3:4"
+        config.monitoring_interval = 0.5
+        config.use_local_gpu_monitor = use_dcgm
+
+        expected_http_port = '2'
+        expected_grpc_port = '4'
+        expected_metrics_interval_ms = 500
```
nv-hwoo, line 65: Is there a reason why the expected metrics interval is 500 (maybe a default)? I feel like we need to document this in a comment.

PR #447, opened 2022-06-30
Fix issue where PA would hang until timeout

tgerdesnv:
When PA aborts with GPU OOM, each of its threads dumps a huge error string.  Piping the result of STDERR was overflowing the buffer, resulting in the subprocess to hang.

The red warning below in the Popen documentation tells us that this can happen. It says that the solution is to use communicate(), but communicate blocks until finished and the code needs to continue gathering CPU utilization

image

The solution is to pipe both stderr and stdout to a (temporary) file

### DIFF
--- a/model_analyzer/perf_analyzer/perf_analyzer.py
+++ b/model_analyzer/perf_analyzer/perf_analyzer.py
@@ -38,13 +38,14 @@ from model_analyzer.constants import \
     MEASUREMENT_WINDOW_STEP, PERF_ANALYZER_MEASUREMENT_WINDOW, \
     PERF_ANALYZER_MINIMUM_REQUEST_COUNT
 
-from subprocess import Popen, PIPE, DEVNULL
+from subprocess import Popen, STDOUT
 import psutil
 import re
 import logging
 import signal
 import os
 import csv
+import tempfile
 
 logger = logging.getLogger(LOGGER_NAME)
 
@@ -100,7 +101,7 @@ class PerfAnalyzer:
         self._config = config
         self._max_retries = max_retries
         self._timeout = timeout
-        self._output = None
+        self._output = ""
         self._perf_records = {}
         self._max_cpu_util = max_cpu_util
 
@@ -232,11 +233,12 @@ class PerfAnalyzer:
         return perf_analyzer_env
 
     def _create_process(self, cmd, perf_analyzer_env):
+        self._cmd_log = tempfile.NamedTemporaryFile()
         try:
             process = Popen(cmd,
                             start_new_session=True,
-                            stdout=PIPE,
-                            stderr=DEVNULL,
+                            stdout=self._cmd_log,
+                            stderr=STDOUT,
                             encoding='utf-8',
                             env=perf_analyzer_env)
         except FileNotFoundError as e:
@@ -271,7 +273,7 @@ class PerfAnalyzer:
 
         while current_timeout > 0:
             if process.poll() is not None:
-                self._output, _ = process.communicate()
+                self._output = self._get_process_output()
                 break
 
             # perf_analyzer using too much CPU?
@@ -280,8 +282,8 @@ class PerfAnalyzer:
                 logger.info(
                     f'perf_analyzer used significant amount of CPU resources ({cpu_util}%), killing perf_analyzer'
                 )
+                self._output = self._get_process_output()
                 process.kill()
-                self._output, _ = process.communicate()
 
                 return self.PA_FAIL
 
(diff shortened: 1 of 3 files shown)


matthewkotila (changes requested):
> @tgerdesnv: but we don't use the error output for now

https://github.com/triton-inference-server/model_analyzer/blob/main/model_analyzer/perf_analyzer/perf_analyzer.py#L302-L304

What about something like above? Isn't that MA using stderr output from PA? Wouldn't piping PA stderr to `/dev/null` prevent this logic from working as intended?

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -280,8 +280,8 @@ def _poll_perf_analyzer(self, process):
                 logger.info(
                     f'perf_analyzer used significant amount of CPU resources ({cpu_util}%), killing perf_analyzer'
                 )
-                self._output = process.stdout.read()
                 process.kill()
+                self._output, _ = process.communicate()
```
Tabrizian, line 284: Does this need to be before process.kill? I think process.kill might be blocked because it cannot right anything to stdout/stderr?

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -290,6 +290,7 @@ def _poll_perf_analyzer(self, process):
             logger.info(
                 'perf_analyzer took very long to exit, killing perf_analyzer')
             process.kill()
+            self._output, _ = process.communicate()
```
Tabrizian, line 293: Same here.

model_analyzer/perf_analyzer/perf_analyzer.py
tgerdesnv: kill will kill immediately, after which communicate will still work
communicate will block until the process is done, which is not what we want.

From the official docs:
image

tgerdesnv:
> > @tgerdesnv: but we don't use the error output for now
> 
> https://github.com/triton-inference-server/model_analyzer/blob/main/model_analyzer/perf_analyzer/perf_analyzer.py#L302-L304
> 
> What about something like above? Isn't that MA using stderr output from PA? Wouldn't piping PA stderr to `/dev/null` prevent this logic from working as intended?

MA uses the STDOUT output, not the STDERR. Those 'errors' are printed to stdout by PA

model_analyzer/perf_analyzer/perf_analyzer.py
Tabrizian: I see, thanks for the clarification.

matthewkotila:
> > > @tgerdesnv: but we don't use the error output for now
> > 
> > 
> > https://github.com/triton-inference-server/model_analyzer/blob/main/model_analyzer/perf_analyzer/perf_analyzer.py#L302-L304
> > What about something like above? Isn't that MA using stderr output from PA? Wouldn't piping PA stderr to `/dev/null` prevent this logic from working as intended?
> 
> MA uses the STDOUT output, not the STDERR. Those 'errors' are printed to stdout by PA

They are printed to stderr [here](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/perf_analyzer/inference_profiler.cc#L455-L553) and [here](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/perf_analyzer/perf_analyzer.cc#L1832)

Tabrizian (approved):

tgerdesnv:
> They are printed to stderr [here](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/perf_analyzer/inference_profiler.cc#L455-L553) and [here](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/perf_analyzer/perf_analyzer.cc#L1832)
> 
> Somewhat related: do we have testing for the MA logic I originally linked that I was concerned would be broken from this change?

Another reason we should be using error codes. Needing to scrape both stdout and stderr for specific strings is extremely fragile.

Yes, the L0_perf_analyzer appears to test this (and is failing with my changes. I will debug)

We also have unit testing in test_perf_analyzer.py that confirms that we behave correctly when "Please use a larger time window" is in stdout: https://github.com/triton-inference-server/model_analyzer/blob/tgerdes-fix-pa-hang/tests/test_perf_analyzer.py#L418-L424

tgerdesnv:
> They are printed to stderr [here](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/perf_analyzer/inference_profiler.cc#L455-L553) and [here](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/perf_analyzer/perf_analyzer.cc#L1832)

You are correct. The old code was piping stderr->stdout, and thus the stdout PIPE had both in it. My original solution only captured stdout. I am now using a tempfile to capture.

matthewkotila (approved):

PR #454, opened 2022-07-11
Updates to quick algo

tgerdesnv:
- Zero throughput instead of ignore and continue on bad PA result
- Change algorithm to be 2D for single-model
- Add variable step magnitude
- Update radius/magnitude/min_init to 3/4*5/3

### DIFF
--- a/tests/test_neighborhood.py
+++ b/tests/test_neighborhood.py
@@ -206,10 +206,11 @@ class TestNeighborhood(trc.TestResultCollector):
 
         n = Neighborhood(nc, cd, Coordinate([0, 0]))
         magnitude = 20
-        new_coord = n.calculate_new_coordinate(magnitude)
 
         # Run it multiple times to make sure no values are changing
+        new_coord = n.calculate_new_coordinate(magnitude)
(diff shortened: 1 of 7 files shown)


tests/test_neighborhood.py
```diff
@@ -201,13 +205,12 @@ def test_calculate_new_coordinate_larger_magnitude(self):
...
+        magnitude = 20
+        new_coord = n.calculate_new_coordinate(magnitude)
 
         # Run it multiple times to make sure no values are changing
-        self.assertEqual(Coordinate([2, 2]),
-                         n.calculate_new_coordinate(magnitude))
-        self.assertEqual(Coordinate([2, 2]),
-                         n.calculate_new_coordinate(magnitude))
+        self.assertEqual(new_coord, Coordinate([2, 2]))
+        self.assertEqual(new_coord, Coordinate([2, 2]))
```
nv-hwoo, line 213: If we are not re-running `calculate_new_coordinate`, then I think we can remove this line.

tgerdesnv: good catch. Fixed

nv-hwoo (approved):

PR #460, opened 2022-07-14
Update experiment data to factor in client batch size

tgerdesnv:


experiments/experiment_data.py
```diff
@@ -125,6 +128,24 @@ def _extract_perf_config_key(self, perf_config):
...
+
+        if len(concurrencies) != len(batch_sizes):
+            raise Exception(f"concurrencies don't match batch sizes")
+
+        for i in range(len(concurrencies)):
+            tmp_int = int(concurrencies[i]) * int(batch_sizes[i])
+            clamped_int = self._clamp_to_power_of_two(tmp_int)
+            concurrencies[i] = str(clamped_int)
+
+        return ';'.join(concurrencies)
```
nv-hwoo, line 142: Should we attach the `--concurrency-range=` prefix back before we return?

tgerdesnv: It doesn't matter. All reads (getting experiment data) and writes (loading checkpoint data into this class) go through this function.

nv-hwoo (approved):

PR #461, opened 2022-07-14
Run analysis after profile

nv-braf:
Refactored code in analyzer
Fixed bugs in metrics manger and result table manager
Updated relevant L0 tests so they only run profile

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -35,8 +35,6 @@ from .model_analyzer_exceptions \
 from .triton.client.client import TritonClient
 from .device.gpu_device import GPUDevice
 
-from typing import Union
-
 import logging
 
 logger = logging.getLogger(LOGGER_NAME)
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -97,6 +97,11 @@ class Analyzer:
         self._get_server_only_metrics(client)
 
         self._profile_models()
+        # TODO: TMA-792: This won't be needed once analysis is part of profile
+        self._analyze_model()
+
+        # TODO: TMA-790: Skip this step if the CLI option indicates the user doesn't want summary reports
+        self._create_summary_reports()
 
         logger.info(self._get_profile_complete_string())
         logger.info("")
@@ -222,6 +227,31 @@ class Analyzer:
                     self._model_manager.run_models(models=[model])
                 finally:
                     self._state_manager.save_checkpoint()
+
+    def _analyze_model(self):
+        gpu_info = self._state_manager.get_state_variable('MetricsManager.gpus')
+        if not gpu_info:
+            gpu_info = {}
+
+        # Create result tables, put top results and get stats
+        self._result_manager.create_tables()
+        self._result_manager.compile_and_sort_results()
+
+        # Dump to tables and write to disk
+        self._result_manager.tabulate_results()
+        self._result_manager.export_results()
+        if verbose:
+            self._result_manager.write_results()
+
+    def _create_summary_reports(self):
+        self._report_manager = ReportManager(
+            mode=mode,
+            config=self._config,
+            gpu_info=gpu_info,
+            result_manager=self._result_manager)
+
+        self._report_manager.create_summaries()
+        self._report_manager.export_summaries()
 
     def _should_profile_multiple_models_concurrently(self):
         return (self._config.run_config_profile_models_concurrently_enable and
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -247,10 +247,6 @@ class Analyzer:
         self._config._fields["analysis_models"] = self._config._fields[
             "profile_models"]
 
-        gpu_info = self._state_manager.get_state_variable('MetricsManager.gpus')
-        if not gpu_info:
-            gpu_info = {}
-
         self._result_manager.compile_and_sort_results()
 
     def _create_summary_tables(self, verbose: bool):
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -104,7 +104,7 @@ class Analyzer:
 
         self._profile_models()
 
-        # The message is in interrupt_hanlder(), so we can just exit
+        # The message is in interrupt_handler(), so we can just exit
         if (self._state_manager.exiting()):
             sys.exit(1)
 
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -97,8 +97,11 @@ class Analyzer:
         self._get_server_only_metrics(client)
 
         self._profile_models()
-        # TODO: TMA-792: This won't be needed once analysis is part of profile
-        self._analyze_model()
+
+        # TODO: TMA-792: This won't be needed once the Results class is used in profile
+        self._analyze_models()
+
+        self._create_summary_tables()
 
         # TODO: TMA-790: Skip this step if the CLI option indicates the user doesn't want summary reports
         self._create_summary_reports()
@@ -228,20 +231,27 @@ class Analyzer:
                 finally:
                     self._state_manager.save_checkpoint()
 
-    def _analyze_model(self):
+    def _analyze_models(self):
+        # TODO: Until we get rid of analysis we need to copy some values from profile
+        self._config._fields["analysis_models"] = self._config._fields[
+            "profile_models"]
+
         gpu_info = self._state_manager.get_state_variable('MetricsManager.gpus')
         if not gpu_info:
             gpu_info = {}
 
-        # Create result tables, put top results and get stats
-        self._result_manager.create_tables()
         self._result_manager.compile_and_sort_results()
 
-        # Dump to tables and write to disk
-        self._result_manager.tabulate_results()
-        self._result_manager.export_results()
+    def _create_summary_tables(self):
+        self._result_table_manager = ResultTableManager(self._config,
+                                                        self._result_manager)
+
+        self._result_table_manager.create_tables()
+        self._result_table_manager.tabulate_results()
+        self._result_table_manager.export_results()
+
         if verbose:
-            self._result_manager.write_results()
+            self._result_table_manager.write_results()
 
     def _create_summary_reports(self):
         self._report_manager = ReportManager(
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import List
 import sys
 from model_analyzer.constants import LOGGER_NAME
 from .model_manager import ModelManager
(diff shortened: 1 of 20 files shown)


model_analyzer/analyzer.py
```diff
@@ -64,7 +70,8 @@ def __init__(self, config, server, state_manager, checkpoint_required):
         self._result_manager = ResultManager(config=config,
                                              state_manager=self._state_manager)
 
-    def profile(self, client, gpus):
+    def profile(self, client: Union[TritonGRPCClient, TritonHTTPClient],
+                gpus: list, mode: str, verbose: bool):
```
nv-hwoo, line 74: Type hint also supports `List` that allows us to specify the object types inside the list as well.

model_analyzer/entrypoint.py
```diff
@@ -260,7 +260,10 @@ def main():
                                 server,
                                 state_manager,
                                 checkpoint_required=False)
-            analyzer.profile(client=client, gpus=gpus)
+            analyzer.profile(client=client,
+                             gpus=gpus,
+                             mode=args.mode,
+                             verbose=bool(args.verbose))
```
nv-hwoo, line 266: I think `args.verbose` is already boolean type.

model_analyzer/analyzer.py
```diff
@@ -223,6 +241,42 @@ def _profile_models(self):
...
                     self._state_manager.save_checkpoint()
 
+    def _analyze_models(self):
+        # TODO: Until we get rid of analysis we need to copy some values from profile
+        self._config._fields["analysis_models"] = self._config._fields[
+            "profile_models"]
+
+        gpu_info = self._state_manager.get_state_variable('MetricsManager.gpus')
+        if not gpu_info:
+            gpu_info = {}
```
nv-hwoo, line 251: Is `gpu_info` used here?

model_analyzer/analyzer.py
```diff
@@ -98,11 +105,22 @@ def profile(self, client, gpus):
...
+        if (self._state_manager.exiting()):
+            sys.exit(1)
+
+        # TODO: TMA-792: This won't be needed once the Results class is used in profile
+        self._analyze_models()
+
+        self._create_summary_tables(verbose)
+
+        # TODO: TMA-790: Skip this step if the CLI option indicates the user doesn't want summary reports
+        self._create_summary_reports(mode)
```
nv-hwoo, line 117: Does the order matter here? Just making sure since `analyze` seems to run these in the opposite order.

model_analyzer/result/result_table_manager.py
```diff
@@ -74,7 +75,8 @@ def tabulate_results(self):
 
         # Fill rows in descending order
         for model in self._result_manager.get_model_names():
-            result_heap = self._result_manager.get_model_sorted_results(model)
+            result_heap = deepcopy(
+                self._result_manager.get_model_sorted_results(model))
```
nv-hwoo, line 79: Do we need to copy result_heap?

model_analyzer/analyzer.py
```diff
@@ -98,11 +105,22 @@ def profile(self, client, gpus):
 
         self._profile_models()
 
+        if (self._state_manager.exiting()):
+            sys.exit(1)
```
nv-hwoo, line 109: Is it okay to terminate without any message?

qa/L0_config/test.sh
```diff
@@ -73,32 +72,24 @@ for config in ${LIST_OF_CONFIG_FILES[@]}; do
         cat $ANALYZER_LOG
         RET=1
     fi
-    MODEL_ANALYZER_ARGS="$MODEL_ANALYZER_ANALYZE_BASE_ARGS"
-    MODEL_ANALYZER_SUBCOMMAND="analyze"
-    run_analyzer
```
nv-hwoo, line 78: Should we still check if the analyze subcommand works in the L0 tests since we haven't entirely deprecated it yet?

model_analyzer/analyzer.py
nv-braf: No, there is no ordering which is why I grouped them this way.

model_analyzer/result/result_table_manager.py
nv-braf: Yes. Because the code below pops all the entries off the heap. Otherwise, you have to be sure that this method is the last thing you call, which I don't think is a very "clean" way of coding things.

qa/L0_config/test.sh
nv-braf: There's separate tests (see L0_analyze) that still use the analyze subcommand.

model_analyzer/analyzer.py
```diff
@@ -64,7 +70,8 @@ def __init__(self, config, server, state_manager, checkpoint_required):
         self._result_manager = ResultManager(config=config,
                                              state_manager=self._state_manager)
 
-    def profile(self, client, gpus):
+    def profile(self, client: Union[TritonGRPCClient, TritonHTTPClient],
```
tgerdesnv, line 73: For the case of multiple implementations of a class, we should just type-hint the parent/interface. In this case, that is just TritonClient.

model_analyzer/analyzer.py
```diff
@@ -98,11 +105,22 @@ def profile(self, client, gpus):
...
 
+        if (self._state_manager.exiting()):
+            sys.exit(1)
+
+        # TODO: TMA-792: This won't be needed once the Results class is used in profile
+        self._analyze_models()
+
+        self._create_summary_tables(verbose)
+
+        # TODO: TMA-790: Skip this step if the CLI option indicates the user doesn't want summary reports
```
tgerdesnv, line 116: I still think profile-only should also skip the csv output as well (which I think is 2 lines above this?)

model_analyzer/analyzer.py
```diff
@@ -223,6 +241,42 @@ def _profile_models(self):
                 finally:
                     self._state_manager.save_checkpoint()
 
+    def _analyze_models(self):
+        # TODO: Until we get rid of analysis we need to copy some values from profile
```
tgerdesnv, line 245: Tag every TODO with a story.

model_analyzer/analyzer.py
```diff
@@ -223,6 +241,42 @@ def _profile_models(self):
...
+
+        self._result_table_manager.create_tables()
+        self._result_table_manager.tabulate_results()
+        self._result_table_manager.export_results()
+
+        if verbose:
+            self._result_table_manager.write_results()
+
+    def _create_summary_reports(self, mode: str):
+        gpu_info = self._state_manager.get_state_variable('MetricsManager.gpus')
```
tgerdesnv, line 267: Let's avoid reaching into state manager though a hardcoded string. Ideally this goes through a public function in ResultManager, although this path seems to be MetricsManager?

model_analyzer/result/result_table_manager.py
tgerdesnv: Can we not come up with a better way to iterate through sorted results without popping?

tgerdesnv (changes requested):

qa/L0_config/test.sh
nv-hwoo: Gotcha. Thanks!

model_analyzer/analyzer.py
nv-braf: Updated to use List

model_analyzer/analyzer.py
nv-braf: Yes, the message is in the interrupt handler(). I've added a comment indicating this.

model_analyzer/analyzer.py
nv-braf: Done

model_analyzer/analyzer.py
nv-braf: Added TMA-792 tag

model_analyzer/analyzer.py
nv-braf: I've filed TMA-809 to track this work

model_analyzer/analyzer.py
nv-braf: Done

model_analyzer/entrypoint.py
nv-braf: Removed the bool wrapper

model_analyzer/analyzer.py
nv-braf: No, I've removed it. Good catch.

model_analyzer/analyzer.py
```diff
@@ -30,6 +32,11 @@
 from .model_analyzer_exceptions \
     import TritonModelAnalyzerException
 
+from .triton.client.client import TritonClient
+from .device.gpu_device import GPUDevice
+
+from typing import Union
```
nv-hwoo, line 38: We can merge the two imports

tgerdesnv (dismissed):

tgerdesnv:
approved to remove my request for changes, but please don't merge until it is actually ready and fully reviewed

nv-hwoo (dismissed):

model_analyzer/analyzer.py
```diff
@@ -98,11 +104,22 @@ def profile(self, client, gpus):
...
+        self._analyze_models()
+
+        # TODO: TMA-790: Skip these steps if the CLI option indicates the user doesn't want summary reports
+        self._create_summary_tables(verbose)
+        self._create_summary_reports(mode)
+
         logger.info(self._get_profile_complete_string())
         logger.info("")
-        logger.info(self._get_analyze_command_help_string())
+        logger.info(self._get_report_command_help_string())
```
tgerdesnv, line 120: have we verified this still works correctly?

model_analyzer/record/metrics_manager.py
```diff
@@ -117,7 +117,7 @@ def _init_state(self):
             if self._gpus[i].device_uuid() not in gpu_info:
                 device_info = {}
                 device = numba.cuda.list_devices()[i]
-                device_info['name'] = device.name
+                device_info['name'] = str(device.name, encoding='utf-8')
```
tgerdesnv, line 120: why was this needed?

model_analyzer/analyzer.py
```diff
@@ -98,11 +104,22 @@ def profile(self, client, gpus):
 
         self._profile_models()
 
+        # The message is in interrupt_hanlder(), so we can just exit
```
tgerdesnv, line 107: typo

model_analyzer/analyzer.py
nv-braf: Yes, I've run profile and verified the output is correct

model_analyzer/record/metrics_manager.py
nv-braf: The device name is returned as a byte-stream - this wasn't an issue before because it was stored in the checkpoint and then read back (which automatically converted it to a string when it went into json format).

model_analyzer/analyzer.py
nv-braf: Fixed

nv-hwoo (approved):

PR #462, opened 2022-07-14
Fix duplicate experiments

tgerdesnv:
Don't generate non-default case for remote mode

### DIFF
--- a/model_analyzer/config/generate/brute_run_config_generator.py
+++ b/model_analyzer/config/generate/brute_run_config_generator.py
@@ -69,12 +69,11 @@ class BruteRunConfigGenerator(ConfigGeneratorInterface):
 
     def _get_next_config(self):
         yield from self._generate_subset(0, default_only=True)
-        if self._do_search():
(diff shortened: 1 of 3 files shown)


model_analyzer/config/generate/brute_run_config_generator.py
```diff
@@ -70,8 +69,12 @@ def get_configs(self):
 
     def _get_next_config(self):
         yield from self._generate_subset(0, default_only=True)
-        self._default_returned = True
-        yield from self._generate_subset(0, default_only=False)
+        if self._do_search():
```
tgerdesnv, line 72: this function name might be misleading.  search_disable=true for us means that we still do manual mode (and do_search would still return true in that case)

I could say call it _should_generate_non_default_only_configs(), but that seems long? Thoughts?

nv-braf: How about `_only_generate_non_default_configs()`?

nv-braf (approved):

PR #466, opened 2022-07-19
Improvements to early exit in manual mode

nv-braf:


nv-hwoo (dismissed):
LGTM.

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -188,6 +189,7 @@ def _reset_concurrencies(self):
 
     def _reset_batch_sizes(self):
         self._curr_batch_size_index = 0
+        self._batch_size_warning_printed = False
```
tgerdesnv, line 192: technically this variable does nothing, because batch_size is the outer loop and _reset_batch_sizes() is never called. Once we exit client batch size sweep we are done.

We should probably just remove the variable and this function. You can leave the warning below

tgerdesnv (changes requested):

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: Done

tgerdesnv: I don't see the changes pushed up

nv-braf: My commit worked, but my push somehow failed. All good now.

tgerdesnv (approved):

PR #467, opened 2022-07-19
Add skip-summary-reports CLI option for profile

nv-hwoo:
* Add new CLI option that skips generating summary reports after profile step.
* Add and fix test cases.

### DIFF
--- a/model_analyzer/config/input/config_command_profile.py
+++ b/model_analyzer/config/input/config_command_profile.py
@@ -29,7 +29,7 @@ from .config_command import ConfigCommand
 from .config_defaults import \
     DEFAULT_BATCH_SIZES, DEFAULT_CHECKPOINT_DIRECTORY, \
     DEFAULT_CLIENT_PROTOCOL, DEFAULT_DURATION_SECONDS, \
-    DEFAULT_GPUS, DEFAULT_MAX_RETRIES, \
+    DEFAULT_GPUS, DEFAULT_SKIP_SUMMARY_REPORTS, DEFAULT_MAX_RETRIES, \
     DEFAULT_MONITORING_INTERVAL, DEFAULT_COLLECT_CPU_METRICS, DEFAULT_OFFLINE_OBJECTIVES, \
     DEFAULT_OUTPUT_MODEL_REPOSITORY, DEFAULT_OVERRIDE_OUTPUT_REPOSITORY_FLAG, \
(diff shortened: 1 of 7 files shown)


nv-braf (commented):
Have you run profile w/ this option enabled? 
Also, we should enable this option for the L0_profile test to ensure we have live coverage of this option

model_analyzer/config/input/config_command_profile.py
```diff
@@ -221,6 +222,14 @@ def _fill_config(self):
                 default_value=DEFAULT_GPUS,
                 description="List of GPU UUIDs to be used for the profiling. "
                 "Use 'all' to profile all the GPUs visible by CUDA."))
+        self._add_config(
+            ConfigField(
+                'skip_summary_reports',
```
Tabrizian, line 227: We used to have this option but it was removed in this commit: https://github.com/triton-inference-server/model_analyzer/commit/93991747f579fe26b0adc914d6de8228e63df0d3#diff-bfa8f90a391c1ffdc5f96eac924441f3941ab2aeb9d7784b89ab3f503cd37c8dL156-L158 Maybe we can use the same name now that we are re-adding it?

nv-braf: We don't want the same name as the default behavior is opposite. We want to create summary reports/tables by default, as that is the common case usage.

We could call this '--skip-summarize', but I'd much rather be slightly more verbose and more explicit with the naming of the flag. I like '--skip-summary-reports' and unless there are strong objections I'd like to keep this new name.

Tabrizian: Sounds good. Resolving.

model_analyzer/config/input/config_command_profile.py
```diff
@@ -221,6 +222,14 @@ def _fill_config(self):
...
+                flags=['--skip-summary-reports'],
+                field_type=ConfigPrimitive(bool),
+                parser_args={'action': 'store_true'},
+                default_value=DEFAULT_SKIP_SUMMARY_REPORTS,
+                description='Specify whether to generate summary reports or not.'))
```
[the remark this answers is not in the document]
nv-hwoo: Fixed.

nv-hwoo:
@nv-braf Yes. I have checked with the option enabled in the profile step.

nv-braf (approved):

PR #468, opened 2022-07-22
RunConfigMeasurement new public methods

nv-braf:
Added three new public methods to support quick algorithm:

- compare_measurements(other) - takes a second RCM and returns a positive value if new RCM is better, negative if worse. The greater the magnitude the better/worse the measurement is
- is_passing_constraints() - returns a bool indicating the RCM is passing all constraints
- set_model_config_constraints(constraints) - sets the constraints for the RCM

Added basic unit testing to cover the new methods

### DIFF
--- a/model_analyzer/result/run_config_measurement.py
+++ b/model_analyzer/result/run_config_measurement.py
@@ -16,6 +16,7 @@ from model_analyzer.constants import COMPARISON_SCORE_THRESHOLD
 from model_analyzer.constants import LOGGER_NAME
 
 from model_analyzer.result.model_config_measurement import ModelConfigMeasurement
+from model_analyzer.result.constraint_manager import ConstraintManager
 from model_analyzer.record.record import RecordType
 
 from statistics import mean
@@ -394,7 +395,12 @@ class RunConfigMeasurement:
         return not self.is_better_than(other)
 
     def is_passing_constraints(self):
-        NotImplemented
+        """
+        Returns true if all model measurements pass
+        their respective constraints
+        """
+        return ConstraintManager.satisfies_constraints(
+            self._model_config_constraints, self)
 
     def compare_measurements(self, other):
         """
--- a/model_analyzer/result/run_config_measurement.py
+++ b/model_analyzer/result/run_config_measurement.py
@@ -50,6 +50,7 @@ class RunConfigMeasurement:
 
         self._model_config_measurements = []
         self._model_config_weights = []
+        self._model_config_constraints = []
 
     @classmethod
     def from_dict(cls, run_config_measurement_dict):
@@ -75,11 +76,15 @@ class RunConfigMeasurement:
         run_config_measurement._model_config_weights = run_config_measurement_dict[
             '_model_config_weights']
 
+        run_config_measurement._model_config_constraints = run_config_measurement_dict[
+            '_model_config_constraints']
+
         return run_config_measurement
 
     def set_model_config_weighting(self, model_config_weights):
         """
         Sets the model config weightings used when calculating 
+        
         weighted metrics
         
         Parameters
@@ -92,6 +97,17 @@ class RunConfigMeasurement:
             model_config_weight / sum(model_config_weights)
             for model_config_weight in model_config_weights
         ]
+
+    def set_model_config_constraints(self, model_config_constraints):
+        """
+
+        Parameters
+        ----------
+            model_config_constraints: dict (single model) or list of dicts (multi-model)
+            Used to determine if an ModelConfigMeasurement passes or fails
+        """
+        self._model_config_constraints = [model_config_constraints] if type(
+            model_config_constraints) is dict else model_config_constraints
 
     def add_model_config_measurement(self, model_config_name,
                                      model_specific_pa_params, non_gpu_data):
@@ -379,6 +395,36 @@ class RunConfigMeasurement:
         """
 
         return not self.is_better_than(other)
+
+    def is_passing_constraints(self):
+        NotImplemented
+
+    def compare_measurements(self, other):
+        """
+        Compares two RunConfig measurements based on each
+        ModelConfigs weighted metric objectives and the
+        ModelConfigs weighted value within the RunConfig
+
+        Parameters
+        ----------
+            other: RunConfigMeasurement
+            
+        Returns
+        -------
+        float
+           Positive value if other is better
+           Negaive value is self is better
+           Zero if they are equal 
+        """
+        # Step 1: for each ModelConfig determine the weighted score
+        weighted_mcm_scores = self._calculate_weighted_mcm_score(other)
+
+        # Step 2: combine these using the ModelConfig weighting
+        weighted_rcm_score = self._calculate_weighted_rcm_score(
+            weighted_mcm_scores)
+
+        # Positive value means self is better than other, so invert this
+        return -1 * weighted_rcm_score
 
     def _compare_measurements(self, other):
         """
--- a/model_analyzer/result/run_config_measurement.py
+++ b/model_analyzer/result/run_config_measurement.py
@@ -76,9 +76,6 @@ class RunConfigMeasurement:
         run_config_measurement._model_config_weights = run_config_measurement_dict[
             '_model_config_weights']
 
-        run_config_measurement._model_config_constraints = run_config_measurement_dict[
-            '_model_config_constraints']
-
         return run_config_measurement
 
     def set_model_config_weighting(self, model_config_weights):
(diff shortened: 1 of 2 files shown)


model_analyzer/result/run_config_measurement.py
```diff
@@ -93,6 +96,17 @@ def set_model_config_weighting(self, model_config_weights):
...
+    def set_model_config_constraints(self, model_config_constraints):
+        """
+
+        Parameters
+        ----------
+            model_config_constraints: dict (single model) or list of dicts (multi-model)
+            Used to determine if an ModelConfigMeasurement passes or fails
+        """
+        self._model_config_constraints = [model_config_constraints] if type(
+            model_config_constraints) is dict else model_config_constraints
```
nv-hwoo, line 108: I think method should stick with a single input type, and it should be the caller's job to match the type.

model_analyzer/result/run_config_measurement.py
```diff
@@ -380,6 +394,41 @@ def __lt__(self, other):
...
+           Zero if they are equal 
+        """
+        # Step 1: for each ModelConfig determine the weighted score
+        weighted_mcm_scores = self._calculate_weighted_mcm_score(other)
+
+        # Step 2: combine these using the ModelConfig weighting
+        weighted_rcm_score = self._calculate_weighted_rcm_score(
+            weighted_mcm_scores)
+
+        # Positive value means self is better than other, so invert this
```
nv-hwoo, line 429: I think it's better to remove this comment since the positive value in this comment does not correspond to the positive value in the docstring, which can be a bit confusing.

model_analyzer/result/run_config_measurement.py
```diff
@@ -380,6 +394,41 @@ def __lt__(self, other):
...
+
+    def compare_measurements(self, other):
+        """
+        Compares two RunConfig measurements based on each
+        ModelConfigs weighted metric objectives and the
+        ModelConfigs weighted value within the RunConfig
+
+        Parameters
+        ----------
+            other: RunConfigMeasurement
```
nv-hwoo, line 413: Remove the indentation.

tests/test_run_config_measurement.py
```diff
@@ -208,6 +210,43 @@ def test_is_better_than_consistency(self):
...
+        """
+        Test to ensure compare measurement function returns
+        the correct magnitude
+        # RCM4's throughput is 1000
+        # RCM5's throughput is 2000
+        # Therefore, the magnitude is (RCM5 - RCM4) / avg throughput
+        #                             (2000 - 1000) / 1500
+        """
+        rcm4_vs_rcm5 = self.rcm4.compare_measurements(self.rcm5)
+        self.assertEqual(rcm4_vs_rcm5, 1000 / 1500)
```
nv-hwoo, line 223: Can we also add `rcm5_vs_rcm4` case to make sure it covers both ways?

model_analyzer/result/run_config_measurement.py
```diff
@@ -93,6 +96,17 @@ def set_model_config_weighting(self, model_config_weights):
             for model_config_weight in model_config_weights
         ]
 
+    def set_model_config_constraints(self, model_config_constraints):
+        """
+
+        Parameters
+        ----------
+            model_config_constraints: dict (single model) or list of dicts (multi-model)
```
nv-hwoo, line 104: Remove the indentation.

tests/test_run_config_measurement.py
```diff
@@ -208,6 +210,43 @@ def test_is_better_than_consistency(self):
...
+        Test to ensure constraints are reported as passing
+        if none were specified
+        """
+        self.assertTrue(self.rcm5.is_passing_constraints())
+
+    def test_is_passing_constraints(self):
+        """
+        Test to ensure constraints are reported as
+        passing/failing if model is above/below
+        throughput threshold
```
nv-hwoo, line 236: Should we also check for max latency?

model_analyzer/result/run_config_measurement.py
nv-braf: Understood. I had just copied what was needed when setting constraints for the RunConfigResult class, but yes I agree this is not needed here. I will just expect a list.

model_analyzer/result/run_config_measurement.py
nv-braf: My goal here was to explain why I'm inverting the value before returning. It's because the inner functions return opposite polarity from what we want, I've rewritten the comment to (hopefully) make it less confusing.

tests/test_run_config_measurement.py
nv-braf: Yes! Done.

tests/test_run_config_measurement.py
nv-braf: This isn't needed. This functionality is covered in test_constraint_manager().

nv-hwoo (dismissed):

model_analyzer/result/run_config_measurement.py
```diff
@@ -80,6 +82,7 @@ def from_dict(cls, run_config_measurement_dict):
     def set_model_config_weighting(self, model_config_weights):
         """
         Sets the model config weightings used when calculating 
+
```
tgerdesnv, line 85: random endl?

model_analyzer/result/run_config_measurement.py
```diff
@@ -380,6 +393,41 @@ def __lt__(self, other):
...
+        their respective constraints
+        """
+        return ConstraintManager.satisfies_constraints(
+            self._model_config_constraints, self)
+
+    def compare_measurements(self, other):
+        """
+        Compares two RunConfig measurements based on each
+        ModelConfigs weighted metric objectives and the
+        ModelConfigs weighted value within the RunConfig
```
tgerdesnv, line 408: do we have a RunConfig anywhere here?  Let's not couple the functionality with the expected usage

model_analyzer/result/run_config_measurement.py
```diff
@@ -380,6 +393,41 @@ def __lt__(self, other):
...
 
+    def is_passing_constraints(self):
+        """
+        Returns true if all model measurements pass
+        their respective constraints
+        """
+        return ConstraintManager.satisfies_constraints(
+            self._model_config_constraints, self)
+
+    def compare_measurements(self, other):
```
tgerdesnv, line 404: add type hints everywhere

tests/test_run_config_measurement.py
```diff
@@ -208,6 +210,48 @@ def test_is_better_than_consistency(self):
...
+        """
+        self.rcm5.set_model_config_constraints([{
```
tgerdesnv, line 241: general python testing question -- does this affect the value of self.rcm5 for all future tests in this file? Could this fail if the test runner decides to run this before test_is_passing_constraints_none?

tgerdesnv (changes requested):

model_analyzer/result/run_config_measurement.py
nv-braf: Removed

model_analyzer/result/run_config_measurement.py
nv-braf: This should be RCM. Updated the comment

tests/test_run_config_measurement.py
nv-braf: It should not as they are created in the setup of each test (and theoretically destroyed in the teardown)

model_analyzer/result/run_config_measurement.py
nv-braf: Done

model_analyzer/result/run_config_measurement.py
```diff
@@ -12,10 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Any, Dict, List, Union
+
+from numpy import intc
```
nv-hwoo, line 17: Remove unused import.

model_analyzer/result/run_config_measurement.py
```diff
@@ -380,11 +405,46 @@ def __lt__(self, other):
...
-    def _compare_measurements(self, other):
+    def is_passing_constraints(self) -> bool:
+        """
+        Returns true if all model measurements pass
+        their respective constraints
+        """
+        return ConstraintManager.satisfies_constraints(
+            self._model_config_constraints, self)
+
+    def compare_measurements(self, other: 'RunConfigMeasurement') -> float:
```
nv-hwoo, line 416: I think [PEP 484](https://peps.python.org/pep-0484/#forward-references) recommends using string literals only when there's cyclic import issue or the class has not been defined yet. But since I'm also pretty new to type hints, this may not be a big issue. Any thoughts?

tgerdesnv: This is the exact case they used in your link -- a function in class RunConfigMeasurement that takes a RunConfigMeasurement as an argument.

nv-hwoo: Oh! Haha totally forgot that this was inside RunConfigMeasurement 😅

tgerdesnv (approved):

PR #469, opened 2022-07-24
Remove measurement-request-count from representation

nv-braf:
Removes measurement-request-count from PA representation field. This way on a checkpoint restore we will treat find/match on this entry even if a non-default MRC was needed.

### DIFF
--- a/model_analyzer/config/run/model_run_config.py
+++ b/model_analyzer/config/run/model_run_config.py
@@ -93,10 +93,29 @@ class ModelRunConfig:
 
     def representation(self):
         """
-        Returns a representation string for the ModelRunConfig that can be used
+        Returns a representation string, containing model name, batch size, 
+        and concurrency for the ModelRunConfig that can be used
         as a key to uniquely identify it
         """
-        return self.perf_config().representation()
+        pc_string = self.perf_config().representation()
+        pc_list = pc_string.split()
+
+        repr_list = []
+        if '-m' in pc_list:
+            repr_list.append(pc_list[pc_list.index('-m')])
+            repr_list.append(pc_list[pc_list.index('-m') + 1])
+
+        if '-b' in pc_list:
+            repr_list.append(pc_list[pc_list.index('-b')])
+            repr_list.append(pc_list[pc_list.index('-b') + 1])
+
+        if '--concurrency' in pc_list:
+            repr_list.append(
+                next((str for str in pc_list if '--concurrency' in str), None))
+
+        repr_str = ' '.join(repr_list)
+
+        return repr_str
 
     def is_legal_combination(self):
         """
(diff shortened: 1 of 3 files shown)


model_analyzer/config/run/model_run_config.py
```diff
@@ -96,7 +97,13 @@ def representation(self):
         Returns a representation string for the ModelRunConfig that can be used
         as a key to uniquely identify it
         """
-        return self.perf_config().representation()
+        repr_string = self.perf_config().representation()
```
tgerdesnv, line 100: Should we just change this in PerfConfig instead?

nv-braf: I didn't feel like it belonged in PerfConfig because I don't think that class should be mucking w/ the representation. It should just be returning it unmutated (what if another class wanted repr fields that we chose to remove because of how we used it in ModelRunConfig.

tgerdesnv: Yeah it all depends on how we expect it to be used. This is the only place that function code is used, and longer term I would like it to be even more stripped down. In my opinion it should have the minimum information needed to uniquely differentiate it from other runs.

model_analyzer/config/run/model_run_config.py
```diff
@@ -96,7 +97,13 @@ def representation(self):
...
+        repr_list = self._remove_mrc_from_list(repr_string.split())
+
+        return ' '.join(repr_list)
+
+    def _remove_mrc_from_list(self, repr_list: List[str]):
```
tgerdesnv, line 105: what is the input here?
Is it something like this?
["-optionA", 10, "-optionB", 15]

or like this?
["-optionA=10", "-optionB=15"]

If it is the first one, this code won't work, because you are only removing the key but not the value (which is the real problem)

The test below implies this is working fine, but I wanted to double check

tgerdesnv: The code seems to assume that there are no spaces between the option and the value, or inside of the option itself.  I doubt that would change, but it doesn't seem robust.  If the input was somehow something with a space like `--measurement-request-count="5, 10"` then this wouldn't work?

nv-braf: It's both! Some options like batch size are '-b <number>', which will get split into two fields. Other's are like MRC, which have the format "measurement-request-count=<number>' all as one string.

There's no easy way to make this robust - you need knowledge of how it's going to be represented in the string from PA (which I admit is ugly).

model_analyzer/config/run/model_run_config.py
nv-braf: I've moved the logic into perf_config.py

tgerdesnv (approved):

PR #470, opened 2022-07-25
Integrate RunConfigMeasurement into quick algo

nv-hwoo:
Summary: Integrate RunConfigMeasurements into the quick search algorithm. Change the coordinate update procedures accordingly using the RunConfigMeasurements. 

* Update Neighborhood class to compute new coordinate based on the collected RunConfigMeasurements.
* Change CoordinateData to store RunConfigMeasurements instead of just throughput values.
* Handle `None` measurements by setting their weights to zeros.
* Introduce back-off mechanism for the case when home coordinate has `None` measurement.
* Decay step magnitude every time we encounter back-offs.
* Clip the step vectors to avoid over-stepping.
* Add/Fix test cases.
* Add type hints.


model_analyzer/config/generate/coordinate_data.py
```diff
@@ -12,43 +12,48 @@
...
 
-    def increment_visit_count(self, coordinate):
+    def increment_visit_count(self, coordinate: Coordinate):
         """
         Increase the visit count for the given coordinate by 1
         """
         key = tuple(coordinate)
         new_count = self.get_visit_count(coordinate) + 1
-        self._visit_counts[key] = new_count
+        self._visit_counts[key] = new_count
```
nv-braf, line 59: Add newline

model_analyzer/config/generate/run_config_generator_factory.py
```diff
@@ -92,7 +92,7 @@ def _create_search_config(command_config):
 
         search_config = SearchConfig(dimensions=dimensions,
                                      radius=3,
-                                     step_magnitude=20,
+                                     step_magnitude=3,
```
nv-braf, line 95: Remove magic numbers and replace with defines

nv-hwoo: Should we define these under `constants.py` since these aren't going to be specified through config (at least for now)?

nv-braf: Yes, that would be my preference

tests/test_neighborhood.py
```diff
@@ -78,30 +102,68 @@ def test_num_initialized(self):
...
+        Test _get_step_vector method that determines the direction to step
+        towards from the home coordinate using the collected coordinates
+        and their measurements.
+
+          1. Get vectors from home to the candidate coordinates
+             and their measurements.
+                [2, 1, 1] - [1, 1, 1] = [1, 0, 0]
+                [1, 2, 1] - [1, 1, 1] = [0, 1, 0]
+
+          2. Multiply each vectors by its weights:
```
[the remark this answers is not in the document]
nv-hwoo: Good point. Changed it to apply different weight.

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -62,26 +78,32 @@ def __init__(self, search_config, config, gpus, models, client):
...
         # updated every step of this generator
-        self._coordinate_to_measure = self._current_coordinate
+        self._coordinate_to_measure = self._home_coordinate
+
+        # Track the best coordinate seen so far that can be used during
+        # the back-off stage.
+        self._best_coordinate = self._origin
+        self._best_measurement = None
 
         # TODO: Add cases to use these
```
tgerdesnv, line 93: remove the todo and the radius_offset

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -122,43 +146,80 @@ def set_last_results(self, measurements):
         """
 
         self._coordinate_data.increment_visit_count(self._coordinate_to_measure)
```
tgerdesnv, line 148: do we have two coordinate datas now?

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -122,43 +146,80 @@ def set_last_results(self, measurements):
...
+        self._neighborhood.coordinate_data.increment_visit_count(
+            coordinate=self._coordinate_to_measure)
 
         if measurements is not None and measurements[0] is not None:
             assert len(measurements) == 1
 
             throughput = measurements[0].get_non_gpu_metric_value(
```
tgerdesnv, line 155: move the debug printing into a helper function that is called before the if statement. The if/else itself should be tiny.

tgerdesnv: for that matter, is the if/else even needed here once the debug code is extracted?

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -122,43 +146,80 @@ def set_last_results(self, measurements):
...
+        logger.debug(
+            f"Stepping back: {self._home_coordinate}->{new_coordinate}"
+        )
+        self._home_coordinate = new_coordinate
+        self._coordinate_to_measure = new_coordinate
+        self._recreate_neighborhood()
+
+        self._magnitude_scaler *= MAGNITUDE_DECAY_RATE
+
+    def _is_home_measurement_none(self):
```
tgerdesnv, line 213: the name doesn't match the functionality. This checks if we are currently measuring home and it is none.

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -257,3 +322,22 @@ def _get_next_perf_analyzer_config(self, model_variant_name, model_num):
         perf_analyzer_config.update_config(
             self._models[model_num].perf_analyzer_flags())
         return perf_analyzer_config
+
+    def _create_zero_throughput_measurement(self) -> RunConfigMeasurement:
```
tgerdesnv, line 326: used?

tests/test_coordinate_data.py
```diff
@@ -12,37 +12,53 @@
...
 from .common import test_result_collector as trc
 
 
 class TestCoordinateData(trc.TestResultCollector):
 
+    def _construct_rcm(self,
+                       throughput: float,
+                       latency: float,
+                       config_name: str = "modelA_config_0"):
+        model_name = "modelA"
```
tgerdesnv, line 31: this hardcoding seems wrong for the modelB case below

tests/test_coordinate_data.py
```diff
@@ -12,37 +12,53 @@
...
+from .common.test_utils import construct_run_config_measurement
 from .common import test_result_collector as trc
 
 
 class TestCoordinateData(trc.TestResultCollector):
 
+    def _construct_rcm(self,
+                       throughput: float,
+                       latency: float,
+                       config_name: str = "modelA_config_0"):
```
tgerdesnv, line 30: remove the default

tests/test_coordinate_data.py
```diff
@@ -59,3 +75,23 @@ def test_visit_count(self):
...
+
+        coordinate0 = Coordinate([0, 0, 0])
+        coordinate1 = Coordinate([0, 4, 1])
+
+        rcm0 = self._construct_rcm(10, 5, config_name="modelA_config_0")
+        rcm1 = self._construct_rcm(20, 8, config_name="modelB_config_0")
+
+        coordinate_data.set_measurement(coordinate0, rcm0)
+        measurement0 = coordinate_data.get_measurement(coordinate0)
+        self.assertEqual("modelA_config_0", measurement0.model_variants_name())
```
tgerdesnv, line 93: this is only testing the name, and isn't actually testing anything about the measurement values?

tests/test_quick_run_config_generator.py
```diff
@@ -244,20 +244,29 @@ def test_get_next_run_config_multi_model(self):
...
+
+    def test_magnitude(self):
+        """
+        Test that _get_magnitude works correctly.
+        """
+        qrcg = self._qrcg
+        self.assertEqual(qrcg._get_magnitude(), 7)  # initial value
+
+        qrcg._magnitude_scaler *= 0.5
+        self.assertAlmostEqual(qrcg._get_magnitude(), 3.5)
```
tgerdesnv, line 266: you should be able to do equals here instead of almost equals

tests/test_quick_run_config_generator.py
```diff
@@ -244,20 +244,29 @@ def test_get_next_run_config_multi_model(self):
...
 
-        self.assertEqual(urcg._get_radius(), 6)
-        self.assertEqual(urcg._get_magnitude(), 8)
+        qrcg._magnitude_scaler *= 0.2
```
tgerdesnv, line 268: just set it to 0.1 instead of multiplying. Otherwise it is confusing

tests/test_neighborhood.py
```diff
@@ -12,18 +12,46 @@
...
 from model_analyzer.config.generate.coordinate_data import CoordinateData
 from model_analyzer.config.generate.coordinate import Coordinate
 
+from .common.test_utils import construct_run_config_measurement
 from .common import test_result_collector as trc
 
 
 class TestNeighborhood(trc.TestResultCollector):
 
+    def _construct_rcm(self, throughput: float, latency: float):
```
tgerdesnv, line 30: 2 files with the same function. Let's move it to utils.

tests/test_neighborhood.py
```diff
@@ -111,121 +173,149 @@ def test_weighted_center(self):
         ])
 
         nc = NeighborhoodConfig(dims, radius=2, min_initialized=3)
-        cd = CoordinateData()
+        n = Neighborhood(nc, home_coordinate=Coordinate([1, 1, 1]))
 
-        n = Neighborhood(nc, cd, Coordinate([1, 1, 1]))
+        rcm0 = self._construct_rcm(throughput=1, latency=5)
```
tgerdesnv, line 178: do these measurements actually get you exact weights of 1.0 and 1.5?

tests/test_neighborhood.py
```diff
@@ -111,121 +173,149 @@ def test_weighted_center(self):
...
+        n.coordinate_data.set_measurement(Coordinate([1, 0]), rcm1)
+        n.coordinate_data.increment_visit_count(Coordinate([1, 0]))
+
+        n.coordinate_data.set_measurement(Coordinate([0, 1]), rcm2)
+        n.coordinate_data.increment_visit_count(Coordinate([0, 1]))
+
+        magnitude = 20
+        new_coord = n.calculate_new_coordinate(magnitude,
+                                               disable_clipping=True)
+        self.assertEqual(new_coord, Coordinate([10, 0]))
```
tgerdesnv, line 225: immediately after try with clipping enabled

tests/test_neighborhood.py
```diff
@@ -111,121 +173,149 @@ def test_weighted_center(self):
...
+        n.coordinate_data.increment_visit_count(Coordinate([0, 0]))
 
-        self.assertEqual(new_coord, Coordinate([3, 0]))
+        n.coordinate_data.set_measurement(Coordinate([1, 0]), rcm1)
+        n.coordinate_data.increment_visit_count(Coordinate([1, 0]))
+
+        n.coordinate_data.set_measurement(Coordinate([0, 1]), rcm2)
+        n.coordinate_data.increment_visit_count(Coordinate([0, 1]))
+
+        magnitude = 20
```
tgerdesnv, line 222: let's put this to a reasonable value (5?)

tests/test_neighborhood.py
```diff
@@ -111,121 +173,149 @@ def test_weighted_center(self):
...
+        n.coordinate_data.increment_visit_count(Coordinate([0, 0]))
 
-        self.assertEqual(new_coord, Coordinate([1, 1]))
+        n.coordinate_data.set_measurement(Coordinate([1, 0]), rcm1)
+        n.coordinate_data.increment_visit_count(Coordinate([1, 0]))
+
+        n.coordinate_data.set_measurement(Coordinate([0, 1]), rcm2)
+        n.coordinate_data.increment_visit_count(Coordinate([0, 1]))
+
+        magnitude = 20
```
tgerdesnv, line 254: 5, and retry with clipping

tests/test_neighborhood.py
```diff
@@ -111,121 +173,149 @@ def test_weighted_center(self):
...
+        n.coordinate_data.set_measurement(Coordinate([0, 0]), rcm0)  # home coordinate
+        n.coordinate_data.increment_visit_count(Coordinate([0, 0]))
+
+        n.coordinate_data.set_measurement(Coordinate([1, 0]), rcm1)
+        n.coordinate_data.increment_visit_count(Coordinate([1, 0]))
+
+        n.coordinate_data.set_measurement(Coordinate([0, 1]), rcm2)
+        n.coordinate_data.increment_visit_count(Coordinate([0, 1]))
+
+        magnitude = 30
```
tgerdesnv, line 287: 7?

tests/test_neighborhood.py
nv-hwoo: Added a separate testing for clipping.

model_analyzer/config/generate/quick_run_config_generator.py
tgerdesnv: The way the interface is designed, we may get a list of measurements. For example, in the brute force code we pick a model config, and then sweep over PA configs. That whole list of results is passed to the model config generator so it can make a decision

tests/test_quick_run_config_generator.py
tgerdesnv: But that isn't what this is testing. You're coupling the two, and making the test confusing as a result.

What you are testing here is how get_magnitude is related to the magnitude_scalar. We don't care how or why magnitude_scalar is updated.

nv-hwoo: Gotcha. Changed to setting the magnitude_scalar to a value instead of multiplying.

model_analyzer/config/generate/quick_run_config_generator.py
nv-hwoo: I see. Thanks for the explanations. I have extracted the debug prints and removed the if/else statements (you're right, we don't need them anymore :))

tests/test_neighborhood.py
tgerdesnv: I think other tests would benefit from a slimmed down method like this, but we can leave it as is for now.

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -144,31 +144,14 @@ def set_last_results(self, measurements: List[RunConfigMeasurement]):
...
-                f"throughput = {throughput}, avg_latency = {avg_latency}"
-            )
-        else:
-            self._neighborhood.coordinate_data.set_measurement(
-                coordinate=self._coordinate_to_measure, measurement=None)
-            logger.debug(
-                f"Measurement for {self._coordinate_to_measure}: None."
-            )
+        self._neighborhood.coordinate_data.set_measurement(
+            coordinate=self._coordinate_to_measure, measurement=measurements[0])
```
tgerdesnv, line 154: if measurements is none this will fail. I may have lead you astray in telling you to completely remove the if/else

tgerdesnv (changes requested):

model_analyzer/config/generate/quick_run_config_generator.py
nv-hwoo: Is there a code in the codebase that passes none to this method? Both model_manager.py and evaluate_config_generator.py passes a list (even when the measurement is none).

tgerdesnv: I think you're right. It is always a list, but the member(s) of the list may be none

tgerdesnv (approved):

PR #473, opened 2022-07-28
Skip unload/reload and server restart

tgerdesnv:
When calling PA, if the model config is the same as the previous one (and we are only changing PA attributes like concurrency) then do not tear down the server and start from scratch. Instead utilizing the server that is already up with the proper model config loaded.

### DIFF
--- a/model_analyzer/record/metrics_manager.py
+++ b/model_analyzer/record/metrics_manager.py
@@ -89,6 +89,7 @@ class MetricsManager:
         self._server = server
         self._result_manager = result_manager
         self._state_manager = state_manager
+        self._loaded_models = None
 
         self._cpu_warning_printed = False
 
@@ -182,14 +183,20 @@ class MetricsManager:
                 "Existing measurement found for run config. Skipping profile")
             return measurement
 
-        self._server.start(env=run_config.triton_environment())
-        if not self._load_model_variants(run_config):
+        current_model_variants = run_config.model_variants_name()
+        if current_model_variants != self._loaded_models:
+            logger.debug(f"Loaded model: {self._loaded_models}")
+            logger.debug(f"Next model to load: {current_model_variants}")
             self._server.stop()
-            return
+            self._server.start(env=run_config.triton_environment())
+
+            if not self._load_model_variants(run_config):
+                self._loaded_models = None
+                return
+
+            self._loaded_models = current_model_variants
 
         measurement = self.profile_models(run_config)
-
-        self._server.stop()
 
         return measurement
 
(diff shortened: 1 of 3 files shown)


qa/L0_perf_analyzer/test.sh
```diff
@@ -63,7 +63,7 @@ for CONFIG_FILE in ${LIST_OF_CONFIG_FILES[@]}; do
         TEST_NAME="count_window"
     elif [[ "$CONFIG_FILE" == "config-additive-args-count-no-adjust.yml" ]]; then
         TEST_NAME="count_window"
-    elif [[ "$CONFIG_FILE" == "config_perf_output_timeout.yml" ]]; then
+    elif [[ "$CONFIG_FILE" == "config-perf-output-timeout.yml" ]]; then
```
nv-hwoo, line 66: Is this an intended change?

model_analyzer/record/metrics_manager.py
```diff
@@ -182,14 +183,19 @@ def execute_run_config(self, run_config):
                 "Existing measurement found for run config. Skipping profile")
             return measurement
 
-        self._server.start(env=run_config.triton_environment())
-        if not self._load_model_variants(run_config):
+        current_model_variants = run_config.model_variants_name()
+        if current_model_variants != self._loaded_models:
             self._server.stop()
-            return
+            self._server.start(env=run_config.triton_environment())
```
nv-hwoo, line 189: In the very first call to this method, it seems like `self._server.stop()` gets called before the server has been started. What happens in such case?

qa/L0_perf_analyzer/test.sh
tgerdesnv: Yes. I have no idea how this ever passed before. You'll notice all other files have dashes instead of underscores. Without this TEST_NAME would be undefined and the check_results script will fail.

model_analyzer/record/metrics_manager.py
tgerdesnv: All stop() calls are gated at the start to make sure there is actually a process. For example:
https://github.com/triton-inference-server/model_analyzer/blob/main/model_analyzer/triton/server/server_local.py#L110

qa/L0_perf_analyzer/test.sh
tgerdesnv: I investigated how it was "passing" before. The TEST_NAME was unmodified from the previous test in the loop. I only discovered this because I locally edited the test to only execute a single test, so it ended up being undefined.

nv-hwoo (approved):

PR #479, opened 2022-07-29
Clean up cpu-only mode

tgerdesnv:
- Remove mention of DCGM requirement in docs
- Downgrade "stopped triton server" print to debug for docker mode (just like it is for local mode)
- Only print "CPU metrics are not being collected" once
- Refactor csv printing. Don't print files with no data
- Removed warning about "No non-gpu metric", as it provided no value
 
I did NOT fix the following warning, because I think it is legit as it comes from the options of what fields to print in the csv files:
- Analysis step printed a ton of warnings about GPU and Server output fields gpu_used_memory, gpu_utilization, and gpu_power_usage

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -49,7 +49,7 @@ for more info on these
 
 ## GPU metrics
 
-These are metrics currently captured using DCGM. They are recorded for each GPU
+These are metrics captured by the tritonserver using DCGM. They are recorded for each GPU
 in fixed intervals during perf analyzer runs and then aggregated across all the
 records for a run. 
 
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -49,7 +49,7 @@ for more info on these
 
 ## GPU metrics
 
-These are metrics captured by the tritonserver using DCGM. They are recorded for each GPU
+These are metrics captured by the tritonserver. They are recorded for each GPU
 in fixed intervals during perf analyzer runs and then aggregated across all the
 records for a run. 
 
--- a/model_analyzer/record/metrics_manager.py
+++ b/model_analyzer/record/metrics_manager.py
@@ -563,9 +563,11 @@ class MetricsManager:
         collect_cpu_metrics_expect = cpu_only or len(self._gpus) == 0
         collect_cpu_metrics_actual = len(self._cpu_metrics) > 0
         if collect_cpu_metrics_expect and not collect_cpu_metrics_actual:
-            logger.info(
-                "CPU metric(s) are not being collected, while this profiling will run on CPU(s)."
-            )
+            if not self._cpu_warning_printed:
+                self._cpu_warning_printed = True
+                logger.warning(
+                    "CPU metric(s) are not being collected, while this profiling will run on CPU(s)."
+                )
         # Warn user about CPU monitor performance issue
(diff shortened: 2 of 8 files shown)


docs/metrics.md
```diff
@@ -49,7 +49,7 @@ for more info on these
 
 ## GPU metrics
 
-These are metrics currently captured using DCGM. They are recorded for each GPU
+These are metrics captured by the tritonserver using DCGM. They are recorded for each GPU
```
nv-braf, line 52: I thought DCGM was being removed from the docs?

model_analyzer/record/metrics_manager.py
```diff
@@ -563,9 +563,11 @@ def _print_run_config_info(self, run_config):
         collect_cpu_metrics_expect = cpu_only or len(self._gpus) == 0
         collect_cpu_metrics_actual = len(self._cpu_metrics) > 0
         if collect_cpu_metrics_expect and not collect_cpu_metrics_actual:
-            logger.info(
-                "CPU metric(s) are not being collected, while this profiling will run on CPU(s)."
-            )
+            if not self._cpu_warning_printed:
+                self._cpu_warning_printed = True
+                logger.warning(
+                    "CPU metric(s) are not being collected, while this profiling will run on CPU(s)."
```
nv-braf, line 569: Is this proper English? Seems like some words are transposed in the warning message?

tgerdesnv: reworded

model_analyzer/record/metrics_manager.py
```diff
@@ -563,16 +563,17 @@ def _print_run_config_info(self, run_config):
...
+                    "One or more models are running on the CPU, but CPU metric(s) are not being collected"
+                )
         # Warn user about CPU monitor performance issue
         if collect_cpu_metrics_actual:
             if not self._cpu_warning_printed:
                 self._cpu_warning_printed = True
-                logger.warning("CPU metric(s) are being collected.")
                 logger.warning(
-                    "Collecting CPU metric(s) can affect the latency or throughput numbers reported by perf analyzer."
+                    "CPU metrics are being collected. This can affect the latency or throughput numbers reported by perf analyzer."
```
nv-hwoo, line 576: Can we make it multiple lines for better readability?

nv-braf (approved):

model_analyzer/result/result_table_manager.py
```diff
@@ -94,30 +94,35 @@ def export_results(self):
...
+                             dir=results_export_directory,
+                             filename=self._config.filename_model_inference,
+                             key=self.model_inference_table_key)
+
+        self._export_results(name="GPU",
+                             dir=results_export_directory,
+                             filename=self._config.filename_model_gpu,
+                             key=self.model_gpu_table_key)
+
+    def _export_results(self, name, dir, filename, key):
```
nv-hwoo, line 116: Can we add docstrings for the parameters? Also, it's missing type hints.

model_analyzer/record/metrics_manager.py
tgerdesnv: yapf/format.py should be taking care of this for us. I'm looking into it.

PR #481, opened 2022-08-02
Don't enable dynamic batching if no max_batch_size in config

tgerdesnv:
If max_batch_size is not specified or is set to 0 in the default config, do not sweep over max_batch_size, and do not enable dynamic batching.

### DIFF
--- a/model_analyzer/config/generate/automatic_model_config_generator.py
+++ b/model_analyzer/config/generate/automatic_model_config_generator.py
@@ -53,7 +53,7 @@ class AutomaticModelConfigGenerator(BaseModelConfigGenerator):
         self._curr_instance_count = self._min_instance_count
         self._curr_max_batch_size = 0
 
-        self._sweep_max_batch_size_disabled = self._determine_sweep_max_batch_size_disabled(
+        self._max_batch_size_disabled = self._determine_max_batch_size_disabled(
         )
 
         self._reset_max_batch_size()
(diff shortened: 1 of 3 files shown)


model_analyzer/config/generate/automatic_model_config_generator.py
```diff
@@ -116,19 +116,19 @@ def _get_curr_param_combo(self):
...
             'instance_group': [{
                 'count': self._curr_instance_count,
                 'kind': self._instance_kind
             }]
         }
 
-        if not self._sweep_max_batch_size_disabled:
+        if not self._max_batch_size_disabled:
             config['max_batch_size'] = self._curr_max_batch_size
+            config['dynamic_batching'] = {}
```
nv-braf, line 127: What if the user doesn't want to enabled dynamic batching?
For example in the L0_config test - the first run has a batch size of 128, but no dynamic batching. And, then the second run enabled dynamic batching. Now aren't you going to produce the same config?

nv-braf (changes requested):

model_analyzer/config/generate/automatic_model_config_generator.py
tgerdesnv: The user has full control if they manually specify what they want to search over. If they don't provide us with anything, then the automatic search does what it can to find the best config -- which always involves turning on dynamic batching.

Yes, this could result in the default config being part of the search space. This already can happen and is fine.

nv-braf (approved):

PR #483, opened 2022-08-03
Fix bugs in config variant mappings when continuing profile from a checkpoint

nv-braf:
When restoring and continuing to profile from a checkpoint the conifg_# are incorrect in many cases, due to the fact that the mappings from model -> config_# are not restored (from the checkpoint) when the re-profile occurs. If the user just increased a max_value, this wouldn't be an issue, but if a min_value was decreased then the numbering wouldn't be consistent and it two model configs with the same number would be different. 

To fix this is required a complete re-write of the ModelVaraintNameManager class. Here are the changes:

- Capture the entire model config dict and compare against it
- Store (and restore) from the class in the checkpoint

This also required a regeneration of all the unit and L0 checkpoints, and it exposed this exact bug in the L0_config test.


model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,79 @@ class ModelVariantNameManager:
...
+                                   model_config_dict: Dict) -> Dict:
+        mcd = deepcopy(model_config_dict)
+        mcd['name'] = model_name
 
-        return self._model_name_index[base_model_name]
+        return mcd
 
-    def _get_hashable_key(self, obj):
-        obj_copy = deepcopy(obj)
+    def _find_existing_variant(self, model_name: str,
```
nv-hwoo, line 76: Input parameter `model_name` is not used.

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,79 @@ class ModelVariantNameManager:
...
+            self._model_config_dicts[model_config_name] = model_config_dict
         else:
-            raise TypeError(type(obj))
+            new_index = self._model_name_index[model_name] + 1
+            self._model_name_index[model_name] = new_index
+            model_config_name = model_name + '_config_' + str(new_index)
+            self._model_config_dicts[model_config_name] = model_config_dict
 
-        return key
+        return model_config_name
```
nv-hwoo, line 100: Missing newline?

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,79 @@ class ModelVariantNameManager:
 
     def __init__(self):
 
-        # Stored as _variant_names[base_model_name][param_combo] = variant_name_string
-        self._variant_names = {}
+        # Dict of {model_config_name: model_config_dict}
+        self._model_config_dicts = {}
```
nv-hwoo, line 26: We could also add type hints like `self._model_config_dicts: Dict[str, dict] = {}`.

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,79 @@ class ModelVariantNameManager:
...
 
-        # Stored as _model_name_index[base_model_name] = current_count_integer
+        # Dict of {base_model_name: current_count_integer}
         self._model_name_index = {}
 
-    def get_model_variant_name(self, base_model_name, param_combo):
+    @classmethod
+    def _from_dict(
```
nv-hwoo, line 32: If it is a classmethod, should we remove the underscore?

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -169,29 +169,38 @@ def make_model_config(param_combo, config, client, gpus, model,
...
+
+        model_config_dict['name'] = variant_name
+        logger.info("")
+        if variant_found:
+            logger.info(
+                f"Found existing model config: {model_config_dict['name']}")
+        else:
+            logger.info(f"Creating model config: {model_config_dict['name']}")
+        for str in logger_str:
+            logger.info(str)
```
nv-hwoo, line 203: Is there a reason why we are logging in this manner?

tests/test_plot_manager.py
```diff
@@ -15,7 +15,7 @@
 import unittest
 
 from .common import test_result_collector as trc
-from .common.test_utils import evaluate_mock_config, ROOT_DIR
+from .common.test_utils import evaluate_mock_config, convert_to_bytes, default_encode, ROOT_DIR
```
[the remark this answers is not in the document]
nv-braf: Yes, but only in the commented out code that is used to regenerate the plots (see line 49).

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,79 @@ class ModelVariantNameManager:
...
+            cls,
+            model_variant_name_manager_dict: Dict) -> "ModelVariantNameManager":
```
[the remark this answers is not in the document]
nv-braf: Agreed, but that's what our formatter chooses to do

model_analyzer/config/generate/model_variant_name_manager.py
nv-braf: Removed

model_analyzer/config/generate/model_variant_name_manager.py
nv-braf: Added

model_analyzer/config/generate/model_variant_name_manager.py
nv-braf: Good catch.

model_analyzer/config/generate/model_variant_name_manager.py
nv-braf: Yes. I've removed it.

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: Yes, we now create the param combo before the variant name, but the logger print order is goes variant name then param combo.

nv-hwoo (approved):

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,78 @@ class ModelVariantNameManager:
...
+            '_model_name_index']
+
+        return model_variant_name_manager
+
+    def get_model_variant_name(self, model_name: str, model_config_dict: Dict,
+                               param_combo: Dict) -> Tuple[bool, str]:
         """
         Given a base model name and a dict of parameters to be applied
-        to the base model config, return the name of the model variant
+        to the base model config, return the name of the model variant
```
tgerdesnv, line 48: the order of the english should match the order of the return values. Bool first.

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,78 @@ class ModelVariantNameManager:
...
-        elif isinstance(obj, collections.abc.Mapping):
-            key = frozenset(
-                (k, self._get_hashable_key_helper(v)) for k, v in obj.items())
-        elif isinstance(obj, collections.abc.Iterable):
-            key = tuple(self._get_hashable_key_helper(item) for item in obj)
+    def _create_new_model_variant(self, model_name: str,
+                                  model_config_dict: Dict) -> str:
+        if model_name not in self._model_name_index:
+            self._model_name_index[model_name] = 0
+            model_config_name = model_name + '_config_0'
```
tgerdesnv, line 91: combine some of the if/else.  Only the determining of new_index and incrementing of model_name_index{} needs to be in the if/else

model_analyzer/model_manager.py
```diff
@@ -55,6 +56,13 @@ def __init__(self, config, gpus, client, server, metrics_manager,
         self._result_manager = result_manager
         self._state_manager = state_manager
 
+        if state_manager.starting_fresh_run():
+            self._init_state()
+
+        self._model_variant_name_manager = ModelVariantNameManager.from_dict(
```
tgerdesnv, line 62: Is there something else that should handle this?  It seems wrong for model manager to need to get MVNM and state_manager to interact in such a precise way.

model_analyzer/model_manager.py
```diff
@@ -55,6 +56,13 @@ def __init__(self, config, gpus, client, server, metrics_manager,
         self._result_manager = result_manager
         self._state_manager = state_manager
 
+        if state_manager.starting_fresh_run():
```
tgerdesnv, line 59: Is there a required ordering between this and the setting of mvnm a few lines lower?

qa/L0_config/config_generator.py
```diff
@@ -52,8 +52,8 @@ def _get_sweep_configs(profile_models):
             } for model in profile_models
         },
     }
-    # total_param=5 because the default config will also be generated
-    model_config['total_param'] = 5
+    # total_param=7 due to dynamic batching
```
tgerdesnv, line 55: explain better.  5 from the previous run, and 4 configs here, but 2 are the same as the previous run

tests/test_model_variant_name_manager.py
```diff
@@ -17,188 +17,109 @@
...
 
     def test_default(self):
         """
-        If no param is passed in, it is considered the default config
+        Check that default config is returned
         """
-        mvnm = ModelVariantNameManager()
-        name = mvnm.get_model_variant_name("modelA", {})
-        self.assertEqual(name, "modelA_config_default")
+        name = self._mvnm.get_model_variant_name("modelA", {'A': 1},
```
tgerdesnv, line 37: call this "result" instead of "name", since it is a tuple

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -21,95 +22,78 @@ class ModelVariantNameManager:
...
+        if variant_found:
+            return (True, model_variant_name)
 
-    def _get_model_variant_name(self, base_model_name, param_combo):
-        """
-        Returns the model variant name that already exists for this combination
-        """
-        hashable_key = self._get_hashable_key(param_combo)
-        return self._variant_names[base_model_name][hashable_key]
+        if self._is_default_config(param_combo):
```
tgerdesnv, line 61: it is weird that we are still passing in the param combo for only this reason. Seems like we should be able to completely operate in the model_config_dict space instead of param_combo, but I don't think this class has enough information. Not sure of anything better. Just take a min to look to see if any better solution

tests/test_model_variant_name_manager.py
```diff
@@ -17,188 +17,109 @@
 import unittest
 from unittest.mock import patch
 
+from tests.common.test_utils import default_encode
+from model_analyzer.constants import DEFAULT_CONFIG_PARAMS
+
 
 class TestModelVariantNameManager(trc.TestResultCollector):
```
tgerdesnv, line 24: Add a test that has more complicated dict matching than a single string-int dict.  I'd like to see multiple dicts inside of a list, and then have the list reordered on a 2nd try (but still be the same)

tests/test_plot_manager.py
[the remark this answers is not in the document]
tgerdesnv: fyi you can have imports in the middle of the code. You could move them into the commented out block.

model_analyzer/config/generate/model_variant_name_manager.py
nv-braf: Done

model_analyzer/config/generate/model_variant_name_manager.py
nv-braf: Done

model_analyzer/model_manager.py
nv-braf: We've talked about this coupling in the past and I agree it's awkward, but I don't of any other way to do this with how things currently exist. it's beyond the scope of this story to make a change, and I'm not sure it's worth the $/! with so few classes needing checkpoints.

model_analyzer/model_manager.py
nv-braf: Yes

tests/test_model_variant_name_manager.py
nv-braf: Calling this a# to be consistent with the other tests

model_analyzer/config/generate/model_variant_name_manager.py
nv-braf: Without param combo I couldn't figure out how to tell what the default config was from the model_config_dict. And, I even if you could I think it would be even more awkward to figure out.

tests/test_model_variant_name_manager.py
nv-braf: I can create a test with nested dicts. But, lists aren't going to work if they're reordered, as they wouldn't be the same.  The actual model_config does have lists, but the order they are in is always consistent.

qa/L0_config/config_generator.py
nv-braf: Done - also updated based on the merge with main

tests/test_model_variant_name_manager.py
nv-braf: Done

tgerdesnv (approved):

PR #486, opened 2022-08-04
Fail if an option is specified in both the CLI and the yaml file

tgerdesnv:
If a user specifies an option in both the yaml config as well as the CLI, fail with an error message.

This has been common in the past with `profile_models`, where the CLI option overwrites the entire set of options specified in the YAML.  Now we will immediately fail and tell the user instead of prioritizing what is on the CLI.


tgerdesnv:
This is apparently "working as intended"

PR #489, opened 2022-08-10
New method to determine which RCM is closer to passing constraints

nv-braf:
New public method in RCM class: compare_constraints()
New static method in constraint_manager: constraint_failure_percentage()

### DIFF
--- a/model_analyzer/result/constraint_manager.py
+++ b/model_analyzer/result/constraint_manager.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Dict
+from typing import List, Dict, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from model_analyzer.result.run_config_measurement import RunConfigMeasurement
 
 
 class ConstraintManager:
@@ -45,7 +48,7 @@ class ConstraintManager:
     @staticmethod
     def satisfies_constraints(
             constraints: List[Dict[str, Dict[str, int]]],
-            run_config_measurement: "RunConfigMeasurement") -> bool:
+            run_config_measurement: 'RunConfigMeasurement') -> bool:
         """
         Checks that the measurements, for every model, satisfy 
         the provided list of constraints
@@ -82,7 +85,7 @@ class ConstraintManager:
     @staticmethod
     def constraint_failure_percentage(
             constraints: List[Dict[str, Dict[str, int]]],
-            run_config_measurement: "RunConfigMeasurent") -> float:
+            run_config_measurement: 'RunConfigMeasurement') -> float:
         """
         Additive percentage, for every measurement, in every model, of how much 
         the RCM is failing the constraints by
--- a/model_analyzer/result/constraint_manager.py
+++ b/model_analyzer/result/constraint_manager.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Dict, Tuple, Union, TYPE_CHECKING
+from typing import List, Dict, TYPE_CHECKING
 
 from model_analyzer.record.record import Record
 
@@ -72,11 +72,10 @@ class ConstraintManager:
         if constraints:
             for (i, model_metrics) in enumerate(run_config_measurement.data()):
                 for metric in model_metrics:
-                    matches, constraint = ConstraintManager._metric_matches_constraint(
-                        metric, constraints[i])
-                    if matches:
+                    if ConstraintManager._metric_matches_constraint(
+                            metric, constraints[i]):
                         if ConstraintManager._get_failure_percentage(
-                                metric, constraint) > 0:
+                                metric, constraints[i][metric.tag]) > 0:
                             return False
 
         return True
@@ -93,27 +92,25 @@ class ConstraintManager:
         -------
         float
         """
-        failure_percentage = 0.0
+        failure_percentage: float = 0
 
         if constraints:
             for (i, model_metrics) in enumerate(run_config_measurement.data()):
                 for metric in model_metrics:
-                    matches, constraint = ConstraintManager._metric_matches_constraint(
-                        metric, constraints[i])
-                    if matches:
+                    if ConstraintManager._metric_matches_constraint(
+                            metric, constraints[i]):
                         failure_percentage += ConstraintManager._get_failure_percentage(
-                            metric, constraint)
+                            metric, constraints[i][metric.tag])
 
         return failure_percentage * 100
 
     @staticmethod
     def _metric_matches_constraint(
-        metric: Record, constraint: Dict[str, Dict[str, int]]
-    ) -> Tuple[bool, Union[Dict[str, int], None]]:
+            metric: Record, constraint: Dict[str, Dict[str, int]]) -> bool:
         if constraint is not None and metric.tag in constraint:
-            return (True, constraint[metric.tag])
+            return True
         else:
-            return (False, None)
+            return False
 
     @staticmethod
     def _get_failure_percentage(metric: Record, constraint: Dict[str,
--- a/model_analyzer/result/constraint_manager.py
+++ b/model_analyzer/result/constraint_manager.py
@@ -87,7 +87,6 @@ class ConstraintManager:
         Additive percentage, for every measurement, in every model, of how much 
         the RCM is failing the constraints by
         
-        
         Returns
         -------
         Float
--- a/model_analyzer/result/constraint_manager.py
+++ b/model_analyzer/result/constraint_manager.py
@@ -70,9 +70,9 @@ class ConstraintManager:
         if constraints:
             for (i, model_metrics) in enumerate(run_config_measurement.data()):
                 for metric in model_metrics:
-                    if constraints[i] is not None and type(
-                            metric).tag in constraints[i]:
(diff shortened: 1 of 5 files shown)


model_analyzer/result/run_config_measurement.py
```diff
@@ -448,6 +448,34 @@ def compare_measurements(self, other: 'RunConfigMeasurement') -> float:
...
+
+        Parameters
+        ----------
+        other: RunConfigMeasurement
+            
+        Returns
+        -------
+        float
+           Positive value if other is closer to passing constraints
+           Negative value is self is closer to passing constraints
```
nv-hwoo, line 465: Typo. Change is -> if

nv-hwoo, line 465: Typo. Change is -> if

model_analyzer/result/constraint_manager.py
```diff
@@ -75,3 +81,36 @@ def satisfies_constraints(constraints, run_config_measurement):
...
+    def constraint_failure_percentage(
+            constraints: List[Dict[str, Dict[str, int]]],
+            run_config_measurement: 'RunConfigMeasurement') -> float:
+        """
+        Additive percentage, for every measurement, in every model, of how much 
+        the RCM is failing the constraints by
+        
+        Returns
+        -------
+        Float
```
nv-hwoo, line 95: Lower case to `float`

tests/test_constraint_manager.py
```diff
@@ -210,6 +210,148 @@ def test_multi_model_constraint_checks(self):
...
+            "perf_throughput": 0
+        }, {
+            "perf_latency_p99": 0,
+            "perf_throughput": 90
+        }])
+        self.assertEqual(
+            ConstraintManager.constraint_failure_percentage(constraints, rcm),
+            50)
+
+    def test_single_model_max_failure_percentage(self):
```
nv-hwoo, line 325: Duplicate function.

model_analyzer/result/constraint_manager.py
```diff
@@ -75,3 +81,36 @@ def satisfies_constraints(constraints, run_config_measurement):
...
+        failure_percentage = 0
+
+        if constraints:
+            for (i, model_metrics) in enumerate(run_config_measurement.data()):
+                for metric in model_metrics:
+                    if constraints[i] is not None and metric.tag in constraints[
+                            i]:
+                        constraint = constraints[i][metric.tag]
+                        if 'min' in constraint:
+                            if metric.value() < constraint['min']:
```
tgerdesnv, line 106: 6 levels of nesting -> smells like it needs a refactor

model_analyzer/result/constraint_manager.py
```diff
@@ -75,3 +81,36 @@ def satisfies_constraints(constraints, run_config_measurement):
...
+            run_config_measurement: 'RunConfigMeasurement') -> float:
+        """
+        Additive percentage, for every measurement, in every model, of how much 
+        the RCM is failing the constraints by
+        
+        Returns
+        -------
+        Float
+        """
+        failure_percentage = 0
```
tgerdesnv, line 97: what are the bounds on this? 0 to 100? 0 to 100 * num_constraints?

model_analyzer/result/constraint_manager.py
tgerdesnv: maybe something like `_get_failure_percentage(metric, constraint)`, which can also be reused in satisfies_constraints()?

model_analyzer/result/run_config_measurement.py
```diff
@@ -448,6 +448,34 @@ def compare_measurements(self, other: 'RunConfigMeasurement') -> float:
...
+        """
+        if self.is_passing_constraints() or other.is_passing_constraints():
+            return None
+
+        self_failing_pct = ConstraintManager.constraint_failure_percentage(
+            self._model_config_constraints, self)
+        other_failing_pct = ConstraintManager.constraint_failure_percentage(
+            other._model_config_constraints, other)
+
+        return (self_failing_pct - other_failing_pct) / 100
```
tgerdesnv, line 477: What is the range of return values?

tests/test_constraint_manager.py
```diff
@@ -210,6 +210,148 @@ def test_multi_model_constraint_checks(self):
...
+        constraints = [
+            ConstraintManager.get_constraints_for_all_models(config)['model_A']
+        ]
+
+        # Constraint is P99 Latency max of 100
+        rcm = self._construct_rcm({"perf_latency_p99": 225})
+        self.assertEqual(
+            ConstraintManager.constraint_failure_percentage(constraints, rcm),
+            125)
+        rcm = self._construct_rcm({"perf_latency_p99": 150})
```
tgerdesnv, line 228: Is there any concern that max-style constraints can be off by an infinite percentage, while min-style constraints are limited to 100%? I suppose that would only matter if there are BOTH min and max style constraints enabled at the same time?

tests/test_run_config_measurement.py
```diff
@@ -252,6 +252,71 @@ def test_is_passing_constraints(self):
         }])
         self.assertFalse(self.rcm5.is_passing_constraints())
 
+    def test_compare_constraints_none(self):
+        """
+        Checks case where either self or other is passing constraints
```
tgerdesnv, line 257: Does it "pass" constraints if there is no constraint?

model_analyzer/result/constraint_manager.py
```diff
@@ -75,3 +81,36 @@ def satisfies_constraints(constraints, run_config_measurement):
...
+
+        if constraints:
+            for (i, model_metrics) in enumerate(run_config_measurement.data()):
+                for metric in model_metrics:
+                    if constraints[i] is not None and type(
```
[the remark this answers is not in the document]
nv-braf: Tim helped me out and this is now fixed (removed type())

model_analyzer/result/constraint_manager.py
nv-braf: Technically, It's 0 to infinity, but in practice it's:

max constraint: 0 to infinity
min constraint: 0 to 100

And then each constraint is additive, along with being per model

model_analyzer/result/run_config_measurement.py
nv-braf: 0 to infinity, see my previous comment

But in practice, it will be 0.01 to a single digit number most of the time

tests/test_constraint_manager.py
nv-braf: I'm not too concerned about that as I don't think that's a very common way of specifying constraints. My thoughts are the user will either want a minimum throughput or a max latency

tests/test_run_config_measurement.py
nv-braf: Yes

model_analyzer/result/constraint_manager.py
nv-braf: I've refactored, but now running into a type check failure I can't solve:

`../../model_analyzer/result/constraint_manager.py:79: error: Argument 2 to "_get_failure_percentage" of "ConstraintManager" has incompatible type "Optional[Dict[str, int]]"; expected "Dict[str, int]"
`

The issue is that `_metric_matches_constraint` can return either: (True, constraint) or (False, None). Any suggestions about how I can get around this?

tgerdesnv: It is pointing out a legit issue. If _metric_matches_constraint() can return <False, None>, then the downstream _get_failure_percentage(metric, constraint) needs to be able to handle "None" for the constraint input arg. I understand that will never actually happen the way the code is structured, but it should still be cleaned up. 

Why do you have to return a tuple from _metric_matches_constraint? Just return a bool, and then if that is true the caller can extract the constraint.

```
if (ConstraintManager._metric_matches_constraint(metric, constraints[i])):
  constraint = constraints[i][metric.tag]
  failure_percentage += ...
```

model_analyzer/result/constraint_manager.py
tgerdesnv: Is it possible to bound to 100 so the scale is the same as the code that compares objectives?

nv-braf: No, I don't think you want to have a bound. Especially, in the situation where you can have multiple models and or constraints.

model_analyzer/result/constraint_manager.py
nv-braf: Understood. I've refactored.

model_analyzer/result/constraint_manager.py
tgerdesnv: So we want to use the same code to take steps via objective differences or constraint differences. Ideally we get return values in the same range. It isn't the end of the world if we are unbound, because we will clamp the step size.

Still, we were able to bound objectives by doing (A-B)/(A+B).  I haven't thought about it too closely, but can we do the same for constraints? We are doing A-B/A or B-A/A.  Can we not have A+B on the bottom?  That math probably doesn't make any sense, but just wanted to throw it out there.

nv-braf: There not really analogusWe're doing A-B/100

The difference here is that we are comparing A against C (the constraint) and B against C. There's no good/easy way to compare A & B directly due to the fact that multiple constraints can be in play (across multiple models).

tgerdesnv (approved):

PR #490, opened 2022-08-10
Check if gpus match checkpoint

nv-braf:
Check to ensure GPUs match their checkpoint counterparts.
If so, we can skip getting the server metrics.
If not, we take an exception.

Enhanced L0_config_search to check for both conditions.


model_analyzer/analyzer.py
```diff
@@ -213,6 +217,12 @@ def _create_model_manager(self, client, gpus):
             state_manager=self._state_manager)
 
     def _get_server_only_metrics(self, client):
+        if self._do_checkpoint_server_metrics_match():
```
tgerdesnv, line 220: bare minimum this should be inside of the `!= c_api` code below. However, I think this will also not work for remote mode.  Then again, remote mode server will already be up and we might be able to solve that case slightly differently?

model_analyzer/analyzer.py
```diff
@@ -344,3 +354,21 @@ def _get_top_n_model_config_names(self, n=-1):
             x.run_config().model_variants_name()
             for x in self._result_manager.top_n_results(n=n)
         ]
+
+    def _do_checkpoint_server_metrics_match(self) -> bool:
```
tgerdesnv, line 358: name this better. We aren't checking the metrics, but rather the GPUs.

model_analyzer/analyzer.py
```diff
@@ -344,3 +354,21 @@ def _get_top_n_model_config_names(self, n=-1):
...
+        foo = nvmlInit()
```
tgerdesnv, line 359: I think you need to also shutdown when you are done?

tgerdesnv (changes requested):

tgerdesnv:
as we discussed offline, it looks like the GPU information is already available to you in the function that calls get_server_only_metrics

nv-braf:
I've removed pynvml. Please re-review

model_analyzer/analyzer.py
```diff
@@ -212,8 +213,14 @@ def _create_model_manager(self, client, gpus):
             metrics_manager=self._metrics_manager,
             state_manager=self._state_manager)
 
-    def _get_server_only_metrics(self, client):
+    def _get_server_only_metrics(self, client, gpus):
         if self._config.triton_launch_mode != 'c_api':
+            if self._do_checkpoint_gpus_match(gpus):
```
tgerdesnv, line 218: What should we do if the GPU's DONT match? Should we error and tell the user to toss away the checkpoint?

model_analyzer/analyzer.py
```diff
@@ -212,8 +213,14 @@ def _create_model_manager(self, client, gpus):
             metrics_manager=self._metrics_manager,
             state_manager=self._state_manager)
 
-    def _get_server_only_metrics(self, client):
+    def _get_server_only_metrics(self, client, gpus):
         if self._config.triton_launch_mode != 'c_api':
```
tgerdesnv, line 217: Does this work for remote mode? Do the GPUs actually match what is on the remote machine?

model_analyzer/analyzer.py
```diff
@@ -344,3 +351,12 @@ def _get_top_n_model_config_names(self, n=-1):
             x.run_config().model_variants_name()
             for x in self._result_manager.top_n_results(n=n)
         ]
+
+    def _do_checkpoint_gpus_match(self, gpus) -> bool:
+        server_only_data = self._result_manager.get_server_only_data()
+
+        for gpu in gpus:
```
tgerdesnv, line 358: This makes sure that all requested GPUs are in the checkpoint, but does not confirm that they are the same.

The user could have run in docker mode with 2 GPUs visible, and later restarted with only 1 GPU visible. In that case this code will say it is a match, but it is not.

nv-braf: Good point. I've updated the code to check that they exactly match.

model_analyzer/analyzer.py
```diff
@@ -344,3 +351,10 @@ def _get_top_n_model_config_names(self, n=-1):
             x.run_config().model_variants_name()
             for x in self._result_manager.top_n_results(n=n)
         ]
+
+    def _do_checkpoint_gpus_match(self, gpus) -> bool:
+        ckpt_data = self._result_manager.get_server_only_data()
+        ckpt_uuids = [ckpt_uuid for ckpt_uuid in ckpt_data.keys()]
+        gpu_uuids = [gpu._device_uuid for gpu in gpus]
+
+        return sorted(ckpt_uuids) == sorted(gpu_uuids)
```
tgerdesnv, line 360: endl

model_analyzer/analyzer.py
nv-braf: Yes. I've added that check.

model_analyzer/analyzer.py
nv-braf: I've tried this in remote mode and it does work as we are able to acquire the gpuids

model_analyzer/analyzer.py
```diff
@@ -212,8 +213,18 @@ def _create_model_manager(self, client, gpus):
...
-    def _get_server_only_metrics(self, client):
+    def _get_server_only_metrics(self, client, gpus):
         if self._config.triton_launch_mode != 'c_api':
+            if self._do_checkpoint_gpus_match(gpus):
+                logger.info(
+                    "GPU devices match checkpoint - skipping server metric acquisition"
+                )
+                return
+            elif gpus is not None:
+                sys.exit(
```
tgerdesnv, line 224: why sys.exit instead of raising an exception?

requirements.txt
```diff
@@ -24,4 +24,4 @@ pdfkit>=0.6.1
 cryptography>=3.3.2
 httplib2>=0.19.0
 tritonclient[all]>=2.4.0
-gevent>=21.12.0
+gevent>=21.12.0
```
tgerdesnv, line 27: is anything different here? if not, revert

model_analyzer/analyzer.py
nv-braf: I've changed this to be an exception

requirements.txt
nv-braf: Reverted

qa/L0_config_search/test.sh
```diff
@@ -122,6 +142,18 @@ for launch_mode in $TRITON_LAUNCH_MODES; do
             RET=1
         fi
 
+        # Check that GPUs don't match and we return an error
+        if [ $launch_mode != 'remote' ]; then
```
tgerdesnv, line 146: why not do this for remote?

qa/L0_config_search/test.sh
```diff
@@ -28,7 +28,7 @@ EXPORT_PATH="`pwd`/results"
 FILENAME_SERVER_ONLY="server-metrics.csv"
 FILENAME_INFERENCE_MODEL="model-metrics-inference.csv"
 FILENAME_GPU_MODEL="model-metrics-gpu.csv"
-TRITON_LAUNCH_MODES="remote local"
+TRITON_LAUNCH_MODES="local remote"
```
tgerdesnv, line 31: any reason why you chose L0_config_search instead of L0_server_launch_modes?

qa/L0_config_search/test.sh
```diff
@@ -122,6 +142,18 @@ for launch_mode in $TRITON_LAUNCH_MODES; do
             RET=1
         fi
 
+        # Check that GPUs don't match and we return an error
+        if [ $launch_mode != 'remote' ]; then
+            sed -i 's/GPU-/GPU-1-/g' $CHECKPOINT_DIRECTORY/1.ckpt
+            
+            run_analyzer
+            if [ $? -e 0 ]; then
```
tgerdesnv, line 150: Also re-do the grep above and make sure it does NOT match

tgerdesnv (changes requested):

qa/L0_config_search/test.sh
nv-braf: Would have been more code complexity to get this to work with keeping the remote server up and then shutting it down correctly.

And, I only need to cover this case for one (non-c_api) mode to exercise this code.

qa/L0_config_search/test.sh
nv-braf: I couldn't easily figure out how to change that test and this gets us most of what we need (local and remote).

qa/L0_config_search/test.sh
nv-braf: Done

tgerdesnv (approved):

PR #491, opened 2022-08-10
Fix type checking failures in quick search algo

nv-hwoo:
* Fix or add missing type hints
* Make Coordinate class to be iterable for compatibility with tuple type hints
* Avoid `round` operator overloading in Coordinate class since Python expects round operator to return integer
* Fix test case


tgerdesnv (changes requested):

nv-braf (approved):
Looks good

tgerdesnv (approved):

PR #493, opened 2022-08-11
Remove wkhtmltopdf requirement from docs

tgerdesnv:
wkhtmltopdf is available in sdk container and will be installed with a wheel or pip install. There is no need to mention it in any docs


nv-braf (approved):
Looks good

PR #497, opened 2022-08-12
Update string representation for instance_group

tgerdesnv:
This means two instances PER gpu:

```
  instance_group [{
      count: 2
      kind: KIND_GPU
    }]
```
 
This means two instances TOTAL:
```
  instance_group [{
      count: 2
      kind: KIND_CPU
    }]
```
 
In our reports we were printing `2/GPU` and `2/CPU`, which implies "2 per gpu" and "2 per cpu". The second one is wrong and confusing.

With the changes, we now print TOTAL instance count. For example, `count:2 kind:KIND_GPU` on an 8-GPU machine used to print `2/GPU`, but will now print `16:GPU`

In the case of instances on both CPU and GPU, we will put a plus in-between. For example: `4:GPU + 1:CPU`


nv-braf (approved):
Looks good

PR #498, opened 2022-08-12
Enable constraints-aided search in quick algo

nv-hwoo:
* Allow quick algo to adjust search based on constraints provided
* Compare constraints when both home and neighbors are all failing the constraints ([#489](https://github.com/triton-inference-server/model_analyzer/pull/489))
* Add test cases

### DIFF
--- a/model_analyzer/config/generate/neighborhood.py
+++ b/model_analyzer/config/generate/neighborhood.py
@@ -97,11 +97,11 @@ class Neighborhood:
             The new coordinate computed based on the neighborhood measurements.
         """
         step_vector = self._get_step_vector() * magnitude
-        step_vector.round()
 
         if enable_clipping:
             step_vector = self._clip_vector_values(vector=step_vector,
                                                    clip_value=clip_value)
+        step_vector.round()
 
         tmp_new_coordinate = self._home_coordinate + step_vector
         new_coordinate = self._clamp_coordinate_to_bounds(tmp_new_coordinate)
@@ -134,7 +134,7 @@ class Neighborhood:
 
         if max_value > clip_value and max_value != 0:
             for i in range(len(vector)):
-                vector[i] = round(clip_value * vector[i]/max_value)
+                vector[i] = clip_value * vector[i]/max_value
         return vector
 
     def pick_coordinate_to_initialize(self) -> Optional[Coordinate]:
--- a/model_analyzer/config/generate/neighborhood.py
+++ b/model_analyzer/config/generate/neighborhood.py
@@ -111,10 +111,9 @@ class Neighborhood:
                             clip_value: int) -> Coordinate:
         """
         Clip the values of the vector to be within the range of
-        [-clip_value, clip_value]. The clipping **approximately** preserves
-        the direction of the vector (e.g. if the clip_value is 2, then
-        [10, 5] will be clipped to [2, 1] instead of [2, 2]). It may not
-        be exact due to the rounding of the values at the end.
+        [-clip_value, clip_value]. The clipping preserves the direction
+        of the vector (e.g. if the clip_value is 2, then [10, 5] will be
+        clipped to [2, 1] instead of [2, 2]).
 
         Parameters
         ----------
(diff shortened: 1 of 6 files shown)


model_analyzer/config/generate/neighborhood.py
```diff
@@ -228,38 +227,129 @@ def _get_visited_coordinates(self) -> List[Coordinate]:
...
+                vectors=vectors,
+                measurements=measurements)
+        else:
+            step_vector = self._optimize_for_passing_constraints(
+                home_measurement=home_measurement,
+                vectors=vectors,
+                measurements=measurements)
+        return step_vector
+
+    def _optimize_for_better_objectives(self,
```
tgerdesnv, line 271: rename to optimize_for_objectives

model_analyzer/config/generate/neighborhood.py
```diff
@@ -228,38 +227,129 @@ def _get_visited_coordinates(self) -> List[Coordinate]:
...
 
+        assert home_measurement is not None, "Home measurement cannot be NoneType."
+
+        if home_measurement.is_passing_constraints():
+            step_vector = self._optimize_for_better_objectives(
+                home_measurement=home_measurement,
+                vectors=vectors,
+                measurements=measurements)
+        else:
+            step_vector = self._optimize_for_passing_constraints(
```
tgerdesnv, line 265: rename to optimize_for_constraints

model_analyzer/config/generate/neighborhood.py
```diff
@@ -228,38 +227,129 @@ def _get_visited_coordinates(self) -> List[Coordinate]:
...
+
         Returns
         -------
         step_vector
             a coordinate that tells the direction to move.
         """
-        step_vector = Coordinate([0] * self._config.get_num_dimensions())
-
-        vectors, measurements = self._compile_neighborhood_measurements()
+        vectors, measurements = self._get_constraints_passing_measurements()
```
tgerdesnv, line 253: The function name implies that you are getting constraints that pass measurements. Rename to get_measurements_passing_constraints

model_analyzer/config/generate/neighborhood.py
```diff
@@ -228,38 +227,129 @@ def _get_visited_coordinates(self) -> List[Coordinate]:
...
+            are passing constraints
+
+        Returns
+        -------
+        step_vector
+            a coordinate that tells the direction to move.
+        """
+        step_vector = Coordinate([0] * self._config.get_num_dimensions())
+
+        if not vectors:
```
tgerdesnv, line 334: Either all of the code to determine the vectors should be inside the function, or all should be outside. Currently half is outside and half is inside.

model_analyzer/config/generate/neighborhood.py
```diff
@@ -273,13 +363,39 @@ def _compile_neighborhood_measurements(
         measurements = []
         for coordinate in visited_coordinates:
             measurement = self._coordinate_data.get_measurement(coordinate)
-            vectors.append(coordinate - self._home_coordinate)
-            measurements.append(measurement)
+            if measurement:
```
tgerdesnv, line 366: We are no longer considering NONE measurements when making decisions?

tgerdesnv (changes requested):

model_analyzer/config/generate/neighborhood.py
nv-hwoo: Updated.

model_analyzer/config/generate/neighborhood.py
nv-hwoo: Updated.

model_analyzer/config/generate/neighborhood.py
nv-hwoo: Updated.

model_analyzer/config/generate/neighborhood.py
nv-hwoo: I agree. Vectors and measurements are now handled inside the function.

model_analyzer/config/generate/neighborhood.py
nv-hwoo: We don't. Previously when we were setting the None measurement weights to zeros, it didn't affect the step vector directions anyway (since adding zero vector does not change the step vector). Only thing that gets affected was its scale since the denominator will be slightly larger if we considered them. However, the fact that None measurement affects the step vector seemed a bit strange, so we simply just ignore them.

tgerdesnv (dismissed):

tgerdesnv (approved):

PR #499, opened 2022-08-16
Updating printing of GPU names and max memory

tgerdesnv:
Old code did not specify number of GPUs, so it looked like only one GPU was running in the reports.
image

image

New code cleans that up:
image

image

### DIFF
--- a/model_analyzer/reports/report_manager.py
+++ b/model_analyzer/reports/report_manager.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Tuple
+from typing import List, Tuple, TYPE_CHECKING
 
-from model_analyzer.result.run_config_measurement import RunConfigMeasurement
+if TYPE_CHECKING:
+    from model_analyzer.result.run_config_measurement import RunConfigMeasurement
 from model_analyzer.constants import LOGGER_NAME, TOP_MODELS_REPORT_KEY
 from model_analyzer.result.constraint_manager import ConstraintManager
 from model_analyzer.record.metrics_manager import MetricsManager
@@ -873,7 +874,8 @@ class ReportManager:
         return len(self._gpu_info)
 
     def _get_gpu_stats(
-            self, measurements: List[RunConfigMeasurement]) -> Tuple[str, str]:
+            self,
+            measurements: List["RunConfigMeasurement"]) -> Tuple[str, str]:
         """
         Gets names and max total memory of GPUs used in measurements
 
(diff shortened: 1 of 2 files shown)


model_analyzer/reports/report_manager.py
```diff
@@ -873,20 +873,39 @@ def _build_detailed_info(self, model_config_name):
...
 
-    def _get_gpu_stats(self, measurements):
+    def _get_gpu_stats(
+            self,
+            measurements: List["RunConfigMeasurement"]) -> Tuple[str, str]:
         """
-        Gets names and memory infos of GPUs used in measurements
+        Gets names and max total memory of GPUs used in measurements
+
+        Returns Tuple(Names(string), MaxMemory(string))
```
nv-hwoo, line 882: nit: Following the return docstring template
```
Returns
-------
(gpu_names, max_memory_str)
    ...
```

model_analyzer/reports/report_manager.py
```diff
@@ -857,10 +858,9 @@ def _build_detailed_info(self, model_config_name):
         gpu_cpu_string = "CPU"
 
         if not run_config.cpu_only():
-            gpu_dict = self._get_gpu_stats(measurements=measurements)
-            gpu_names = ','.join(list(gpu_dict.keys()))
-            max_memories = ','.join([str(x) + ' GB' for x in gpu_dict.values()])
-            gpu_cpu_string = f"GPU(s) {gpu_names} with memory limit(s) {max_memories}"
+            (gpu_names,
+             max_memories) = self._get_gpu_stats(measurements=measurements)
```
nv-hwoo, line 862: nit: I think we can make it a single line for better readability: 
`gpu_names, max_memories = self._get_gpu_stats(measurements)`

tests/test_report_manager.py
```diff
@@ -433,6 +433,61 @@ def _test_instance_group_helper(self, model_config_dict, expected_output):
...
+            model_config_names=MagicMock(),
+            model_specific_pa_params=MagicMock(),
+            gpu_metric_values=avg_gpu_metrics,
+            non_gpu_metric_values=MagicMock(),
+            metric_objectives=MagicMock(),
+            model_config_weights=MagicMock())
+
+        measurements = [measurement, measurement]
+        (names,
+         max_gpu) = report_manager._get_gpu_stats(measurements=measurements)
```
nv-hwoo, line 487: Same as above

tests/test_report_manager.py
```diff
@@ -433,6 +433,61 @@ def _test_instance_group_helper(self, model_config_dict, expected_output):
...
+        measurement = construct_run_config_measurement(
+            model_name=MagicMock(),
+            model_config_names=MagicMock(),
+            model_specific_pa_params=MagicMock(),
+            gpu_metric_values=avg_gpu_metrics,
+            non_gpu_metric_values=MagicMock(),
+            metric_objectives=MagicMock(),
+            model_config_weights=MagicMock())
+
+        measurements = [measurement, measurement]
```
nv-hwoo, line 485: What happens when the two measurements have different gpu ids?

tgerdesnv: updated the test to have one measurement use GPUs 1 and 4, and the other use 1 and 2.

nv-hwoo (approved):

PR #501, opened 2022-08-17
Set custom objectives to RunConfigMeasurements

nv-hwoo:
When the custom objective is defined by the user using yaml config file, make sure we pass that information to the measurements so that quick search algorithm can use it for the search.

### DIFF
--- a/experiments/experiment_config_command_creator.py
+++ b/experiments/experiment_config_command_creator.py
@@ -26,9 +26,6 @@ class ExperimentConfigCommandCreator:
 
     @staticmethod
     def make_config(data_path, model_name, other_args):
-        mock_model_config = MockModelConfig("")
-        mock_model_config.start()
-
         checkpoint_dir = f"{data_path}/{model_name}"
 
         #yapf: disable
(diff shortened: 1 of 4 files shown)


experiments/experiment_config_command_creator.py
```diff
@@ -26,22 +26,30 @@ class ExperimentConfigCommandCreator:
...
             '--profile-models', model_name,
             '--model-repository', data_path,
-            '--checkpoint-directory', checkpoint_dir,
-            '-f', 'path-to-config-file'
+            '--checkpoint-directory', checkpoint_dir
         ]
         args += other_args
 
-        yaml_content = convert_to_bytes("")
+        if '-f' not in args and '--config-file' not in args:
```
nv-braf, line 40: Seems like there are two separate issues being solved in this story? What does this have to do with the failure to set metric weightings?

nv-hwoo: Commit message may be a bit misleading. I don't think I fixed it, but merely just bypassed the issue. For some unknown reason in ArgumentParser in CLI class, specifying `--config-file` flag in command line causes an TypeError issue. This only happens for running experiments folder, so I'm bypassing it by reading the file and passing it to mock_config.

nv-braf: I let Tim, weigh in on how clean we want to keep the experiments folder, but this seems hacky to me. Can you post the error message?

tgerdesnv: I think this was to allow us to specify a custom yaml so we could test with custom constraints/objectives. The existing code just mocked out an empty yaml file, so this was his workaround.

Can you push the real fixes into main and this change into my side branch?

nv-hwoo: Sure. I'll only leave the fixes for main on this PR.

tgerdesnv (approved):

PR #503, opened 2022-08-18
Add to result heaps during profile

nv-braf:
This changes our paradigm and adds measurement results to the heaps as each is profiled, versus waiting until they have all completed.

This will allow the quick algorithm to determine rankings of a node during it's walk.

The big change here is the removal of `compile_and_sort_results()`. Now, checkpoint results are loaded when the result manager is initialized and both the Results class and the heaps are appended when an RCM is added during profile.


tests/test_result_manager.py
```diff
@@ -132,45 +129,44 @@ def test_get_single_model_sorted_results(self):
...
-        fake_model.model_name.return_value = "FakeModel"
-        old_analysis_models = result_manager._config.analysis_models
-        old_analysis_models.append(fake_model)
-        result_manager._config.analysis_models = old_analysis_models
-
-        fake_run_config = MagicMock()
-        fake_run_config.models_name.return_value = "FakeModel"
-
-        result_manager.add_run_config_measurement(fake_run_config, MagicMock())
+    # FIXME: This doesn't work anymore since model names are loaded at __init__
```
nv-braf, line 145: @tgerdesnv - This unit test is problematic and I can't think of a good way of easily fixing it. The issue is that all models must now be known when the result manager is created. Thoughts?

tests/test_report_manager.py
```diff
@@ -254,9 +252,10 @@ def subtest_build_summary_table(self, mode, cpu_only):
             self.assertGreaterEqual(current_row[throughput_index],
                                     next_row[throughput_index])
 
-    def test_build_detailed_info(self):
-        for cpu_only in [True, False]:
-            self._subtest_build_detailed_info(cpu_only)
+    # FIXME: You can't add results if the config is REPORT
```
nv-braf, line 255: @tgerdesnv  - This unit test is problematic and I can't think of a good way of easily fixing it. The issue is that you can't add new measurements if the report subcommand is used (as we don't have any run_comparators)

tests/test_analyzer.py
```diff
@@ -57,19 +57,14 @@ def mock_get_state_variable(self, name):
     def mock_get_list_of_models(self):
         return ['model1']
 
-    @patch.multiple(
```
tgerdesnv, line 60: Are there any real changes to this file? Or just formatting?

tests/test_result_manager.py
tgerdesnv: We need to be able to say "get top 3 configs across all models". Is this protected anywhere else? Does that functionality still work?

tests/test_report_manager.py
tgerdesnv: We need to be able to test this text creation. There must be a way to do it if the live code is working.

model_analyzer/result/result_manager.py
```diff
@@ -252,6 +247,18 @@ def _init_state(self):
         self._state_manager.set_state_variable('ResultManager.server_only_data',
                                                {})
 
+    def _setup_for_analysis(self):
+        # Skip if the report subcommand is being executed
+        if 'analysis_models' in self._config._fields:
```
tgerdesnv, line 252: This is ugly. What is the purpose?

model_analyzer/result/result_manager.py
```diff
@@ -44,13 +45,22 @@ def __init__(self, config, state_manager):
...
         # Data structures for sorting results
         self._per_model_sorted_results = defaultdict(ResultHeap)
         self._across_model_sorted_results = ResultHeap()
 
+        # TODO: TMA-792: Until we get rid of analysis we need to model names from profile
+        if 'profile_models' in self._config._fields:
+            self._config._fields["analysis_models"] = self._config._fields[
+                "profile_models"]
+
+        if state_manager.starting_fresh_run():
```
tgerdesnv, line 57: This isn't clicking for me. Let's chat

tests/test_analyzer.py
nv-braf: analysis_models was removed

tests/test_report_manager.py
nv-braf: I fixed the unit test

tests/test_result_manager.py
nv-braf: I've fixed the unit test

model_analyzer/result/result_manager.py
nv-braf: I've refactored, please re-review.

model_analyzer/result/result_manager.py
nv-braf: I've refactored, please re-review.

model_analyzer/result/result_manager.py
```diff
@@ -318,13 +364,13 @@ def _add_results_to_heaps(self):
         results = self._state_manager.get_state_variable(
             'ResultManager.results')
 
-        for model_name in self._analysis_model_names:
-            model_measurements = results.get_model_measurements_dict(model_name)
+        for model_name in results.get_list_of_models():
```
tgerdesnv, line 367: is this correct for multi-model?

Also, the only difference here is that we no longer assert? Is that assert handled elsewhere now for the case it matters?

model_analyzer/result/result_manager.py
```diff
@@ -98,46 +105,31 @@ def add_server_data(self, data):
...
-        results = self._state_manager.get_state_variable(
-            'ResultManager.results')
+        run_config_result = RunConfigResult(
+            model_name=model_name,
+            run_config=run_config,
+            comparator=self._run_comparators[model_name],
+            constraints=self._run_constraints[model_name])
 
-        results.add_run_config_measurement(run_config, run_config_measurement)
+        run_config_measurement.set_metric_weightings(
```
tgerdesnv, line 122: I don't love that we have to reach into private variables here.

also, is there ever a time we would want to set metric weights and not model weights, or vice-versa?

nv-braf: Yeah, I agree, but this is an issue with the RunConfigResultComparator class. I'll add a public get function to the class.

model_analyzer/result/result_manager.py
nv-braf: Yes, I believe so. The Results class already has the correct multi-model name (comes from the checkpoint).

There's no assertion here. This is new logic to ensure that we don't add checkpointed models to the heap that the user didn't specify on a rerun. Note: they are still present in the Result class, so these other model results won't be lost when the new checkpoint is created.

model_analyzer/result/result_manager.py
nv-braf: Added public class.

model_analyzer/result/result_manager.py
tgerdesnv: re-add assert as discussed

model_analyzer/result/result_manager.py
```diff
@@ -319,9 +365,10 @@ def _add_results_to_heaps(self):
             'ResultManager.results')
 
         for model_name in self._analysis_model_names:
-            model_measurements = results.get_model_measurements_dict(model_name)
-
-            if not model_measurements:
+            try:
```
tgerdesnv, line 368: IS this correct? Does it now raise instead of returning None if it does not exist?

tgerdesnv: can you add a unit test to confirm if there is not one already/

model_analyzer/result/result_manager.py
nv-braf: Assert added. I've had to fix a bunch of unit tests as a fallout from this. Please review the latest change.s

model_analyzer/result/result_manager.py
nv-braf: We (inadvertently) have multiple unit tests covering this case. I've fixed the unit tests so that they run profile instead of analysis, before adding in their measurements.

model_analyzer/result/result_manager.py
```diff
@@ -252,6 +244,38 @@ def _init_state(self):
         self._state_manager.set_state_variable('ResultManager.server_only_data',
                                                {})
 
+    def _complete_setup(self):
+        # The Report subcommand can init, but nothing needs to be done
+        if isinstance(self._config, ConfigCommandProfile):
+            self._complete_profile_setup()
+        elif isinstance(self._config, ConfigCommandAnalyze):
+            self._complete_analyze_setup()
```
tgerdesnv, line 252: elif report:
  pass
else:
  raise exception

tests/test_report_manager.py
```diff
@@ -137,9 +154,14 @@ def setUp(self):
         self.json_mock = MockJSONMethods()
         self.json_mock.start()
 
+        # if not os.path.exists("/test/export/path"):
```
tgerdesnv, line 157: needed?

tests/test_report_manager.py
```diff
@@ -519,6 +547,7 @@ def tearDown(self):
         self.os_mock.stop()
         self.json_mock.stop()
         patch.stopall()
+        # shutil.rmtree("/test")
```
tgerdesnv, line 550: needed?

tests/test_result_table_manager.py
```diff
@@ -104,6 +110,9 @@ def test_create_inference_table_with_backend_parameters(self):
             'backend_parameter/parameter_2'
         ])
 
+    @patch(
```
tgerdesnv, line 113: @patch('...path...', MagicMock(return_value=True))

tests/test_report_manager.py
nv-braf: Removed

tests/test_report_manager.py
nv-braf: Removed

model_analyzer/result/result_manager.py
nv-braf: Updated to add pass & exception

tests/test_result_table_manager.py
nv-braf: Fixed

tgerdesnv (dismissed):

tgerdesnv (approved):

PR #504, opened 2022-08-22
Sweep experiments script

tgerdesnv:
New script to run the experiment script on all models in all checkpoints in a directory (recursively) 

Also, make the following updates to the experiment script:
- support for running experiment without default config (it can get it from the checkpoint)
- support for custom constraints and objectives
- support for exponential instance count
- debug print throughput/latency map

### DIFF
--- a/experiments/scripts/sweep_checkpoints.py
+++ b/experiments/scripts/sweep_checkpoints.py
@@ -137,8 +137,10 @@ def run_all_models():
                 continue
 
             percentile = float(result_data['Percentile'])
-            num_measurement = int(result_data['Generator num configs'])
 
+            num_measurement = int(result_data['Generator num measurements'])
+            num_measurement += int(
+                result_data['Generator missing num measurements'])
             print(
                 f"  {model}: Percentile = {percentile}, measurements = {num_measurement}"
             )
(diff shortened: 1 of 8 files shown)


experiments/scripts/sweep_checkpoints.py
```diff
@@ -0,0 +1,197 @@
...
+
+            if 'Percentile' not in result_data:
+                print(f"{model} failed! Skipping")
+                continue
+            if 'Generator num configs' not in result_data:
+                print(f"{model} failed! Skipping")
+                continue
+
+            percentile = float(result_data['Percentile'])
+            num_measurement = int(result_data['Generator num configs'])
```
nv-hwoo, line 140: Should this use "num measurements" and "missing num measurements" to get number of measurements?

experiments/scripts/sweep_checkpoints.py
```diff
@@ -0,0 +1,197 @@
...
+    NUM_MEASUREMENTS_MAX = 15
+    NUM_MEASUREMENTS_MIN = 5
+    ckpts = find_all_checkpoints(PATH)
+    total_models = 0
+
+    # FIXME
+    #ckpts = ckpts[80:82]
+    for i, ckpt in enumerate(ckpts):
+
+        # FIXME -- infinite loop
```
nv-hwoo, line 106: We can remove this since infinite loop is fixed.

nv-hwoo (approved):

PR #506, opened 2022-08-29
No batching support

tgerdesnv:
Don't turn on dynamic batching if it is not supported (or if sequence batching is already enabled)
Only search instance count in quick search for models that don't support batching

### DIFF
--- a/model_analyzer/config/generate/base_model_config_generator.py
+++ b/model_analyzer/config/generate/base_model_config_generator.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from model_analyzer.result.run_config_measurement import RunConfigMeasurement
 from .config_generator_interface import ConfigGeneratorInterface
-
+from typing import List, Union
 from model_analyzer.constants import LOGGER_NAME
 from model_analyzer.triton.model.model_config import ModelConfig
 from .model_profile_spec import ModelProfileSpec
@@ -55,11 +56,11 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
         self._model_name_index = 0
         self._generator_started = False
         self._max_batch_size_warning_printed = False
-        self._last_results = []
+        self._last_results: List[RunConfigMeasurement] = []
         # Contains the max throughput from each provided list of measurements
         # since the last time we stepped max_batch_size
         #
-        self._curr_max_batch_size_throughputs = []
+        self._curr_max_batch_size_throughputs: List[float] = []
 
     def _is_done(self):
         """ Returns true if this generator is done generating configs """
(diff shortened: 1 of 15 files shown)


model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -16,13 +16,19 @@
...
 from model_analyzer.triton.model.model_config import ModelConfig
+from typing import NamedTuple
 
 import abc
 import logging
 
 logger = logging.getLogger(LOGGER_NAME)
 
 
+class ModelBatchingConfig(NamedTuple):
```
tgerdesnv, line 27: Does this deserve its own file?
Should it be expanded to something more generic (that could also contain things like cpu_only)?

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -119,7 +119,7 @@ def _last_results_increased_throughput(self):
             lastest_throughput > prev_throughput
             for prev_throughput in self._curr_max_batch_size_throughputs[:-1])
 
-    def _get_last_results_max_throughput(self):
+    def _get_last_results_max_throughput(self) -> Union[float, None]:
```
nv-braf, line 122: It looks cleaner to say Optional[float] instead of Union[float, None] and I've switched over to this in my code:
https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint

model_analyzer/config/generate/model_profile_spec.py
```diff
@@ -0,0 +1,48 @@
...
+            return False
+        return True
+
+    def supports_dynamic_batching(self) -> bool:
+        """ Returns True if this model supports dynamic batching. Else False """
+        supports_dynamic_batching = self.supports_batching()
+
+        if "sequence_batching" in self._default_model_config:
+            supports_dynamic_batching = False
+        return supports_dynamic_batching
```
nv-braf, line 48: missing newline

model_analyzer/config/generate/model_profile_spec.py
```diff
@@ -0,0 +1,48 @@
...
+from model_analyzer.config.input.objects.config_model_profile_spec import ConfigModelProfileSpec
+from model_analyzer.triton.model.model_config import ModelConfig
+
+
+class ModelProfileSpec(ConfigModelProfileSpec):
+    """
+    The profile configuration and default model config for a single model to be profiled
+    """
+
+    def __init__(self, spec: ConfigModelProfileSpec, config, client, gpus):
```
nv-braf, line 25: add type checking for config, client, gpus

model_analyzer/config/generate/base_model_config_generator.py
tgerdesnv: Nice! I've boyscouted that everywhere I could find

model_analyzer/config/generate/model_profile_spec.py
tgerdesnv: done

PR #508, opened 2022-08-31
Adding concurrency sweep after quick search completes

nv-braf:
After quick search completes we will now sweep concurrency over the top N configs, as well as the default config.

Changes include:

- New high level generator class that wraps around the quick search
- Fix to the MVNM class to check for default config before a match in the mapper (quick search can create a default config, but give it a number)
- New flag to skip generation of default configs in the brute sweep
- Removing rate limiter from quick search - this was causing us not to match (and create new config #'s), when sweeping concurrencies using brute

These are results from a run of Resnet50 using quick + brute:

Capture1
Capture2


model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,188 @@
...
+        self._search_config = search_config
+        self._config = config
+        self._gpus = gpus
+        self._models = models
+        self._client = client
+        self._result_manager = result_manager
+        self._model_variant_name_manager = model_variant_name_manager
+        self._rcg = None
+
+    def set_last_results(self, measurements):
```
nv-braf, line 90: I tried to add typing here: ```measurements: List[Union[RunConfigMeasurement,None]]``` (like in quick search)

But got the following error that I couldn't figure out how to fix: ```../../model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py:92: error: "None" has no attribute "set_last_results"```

I'm pretty sure this is because I initialize `self._rcg to None`, but that seems like the correct thing to do. Any suggestions?

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        yield from self._execute_quick_search()
+        yield from self._sweep_concurrency_over_top_results()
+
+    def _execute_quick_search(self):
+        self._rcg = self._create_quick_run_config_generator()
+
+        yield from self._rcg.get_configs()
+
+    def _create_quick_run_config_generator(self) -> QuickRunConfigGenerator:
```
tgerdesnv, line 108: Don't we have a factory that does this?

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,188 @@
...
+            search_config=self._search_config,
+            config=self._config,
+            gpus=self._gpus,
+            models=self._models,
+            client=self._client,
+            model_variant_name_manager=self._model_variant_name_manager)
+
+    def _sweep_concurrency_over_top_results(self):
+        top_results = self._result_manager.top_n_results(
+            n=DEFAULT_NUM_CONFIGS_PER_MODEL)
```
tgerdesnv, line 119: This is wrong. This default is fed into the ConfigCommandProfile, and is used if not specified on the command line. You need to use the value in the config object.

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,188 @@
...
+    def _create_new_config_command_profile(
+            self, result: RunConfigResult) -> ConfigCommandProfile:
+        new_config = deepcopy(self._config)
+
+        self._set_search_mode(new_config)
+        self._set_parameters(result, new_config)
+
+        return new_config
+
+    def _create_brute_run_config_generator(
```
tgerdesnv, line 137: existing factory

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,188 @@
...
+                new_config, skip_default_config=(count != 0))
+
+            yield from self._rcg.get_configs()
+
+    def _create_new_config_command_profile(
+            self, result: RunConfigResult) -> ConfigCommandProfile:
+        new_config = deepcopy(self._config)
+
+        self._set_search_mode(new_config)
+        self._set_parameters(result, new_config)
```
tgerdesnv, line 133: I don't love that result is an input, and new_config is an input and output. Can we make them all return config so the code is more obvious what is being modified?

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,188 @@
...
+        return BruteRunConfigGenerator(
+            config=new_config,
+            gpus=self._gpus,
+            models=self._models,
+            client=self._client,
+            model_variant_name_manager=self._model_variant_name_manager,
+            skip_default_config=skip_default_config)
+
+    def _set_search_mode(self, config: ConfigCommandProfile):
+        config._fields['run_config_search_mode']._field_type._value = 'brute'
```
tgerdesnv, line 149: We need to hide this magic inside of the base ConfigCommand class

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,188 @@
...
+        batch_size = self._find_batch_size(result)
+        self._set_batch_size(config, batch_size)
+
+        instance_count = self._find_instance_count(result)
+        self._set_instance_count(config, instance_count)
+
+        self._set_concurrency(config)
+
+    def _find_batch_size(self, result: RunConfigResult) -> int:
+        return result.run_config().model_run_configs()[0].model_config(
```
tgerdesnv, line 164: hardcoded 0? Will this break for multi-model?

tgerdesnv (changes requested):

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: The factory now has the quick+sweep generator

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: I tried to use the RCG factory, but I ran into a circular import issue

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: Done

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: Yes, this is only going to work for single model, but didn't think this was an issue (at the moment), since quick search is only for single model.

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: I agree this isn't clean, but I don't know how to easily do this and wasn't sure it was worth the effort to figure this out.

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: I didn't think the user had the ability to specify min/max concurrencies with a quick search, so I just figured I would use the defaults.

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: For multi-model, this approach isn't going to work as we can't use the search space to set batch size or instance count if each model has a different value.

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
tgerdesnv: This isn't min/max concurrencies. It is top-N. It should be using config.num_configs_per_model:
```
            ConfigField(
                'num_configs_per_model',
                flags=['--num-configs-per-model'],
                field_type=ConfigPrimitive(int),
                default_value=DEFAULT_NUM_CONFIGS_PER_MODEL,
                description=
                'The number of configurations to plot per model in the summary.'
            ))
```

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
tgerdesnv: looks like you fixed most cases, but not this line

model_analyzer/config/input/config_command.py
```diff
@@ -179,3 +179,9 @@ def __deepcopy__(self, memo):
 
     def __getattr__(self, name):
         return self._fields[name].value()
+
+    def __setattr__(self, name, value):
+        if name == '_fields':
+            self.__dict__[name] = value
+        else:
+            self._fields[name]._field_type._value = value
```
tgerdesnv, line 187: is _fields[name] a ConfigValue or a ConfigField?  Regardless, I think they have a set_value() function that should be used here.

tgerdesnv (changes requested):

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
tgerdesnv: It seems we picked the wrong approach then. We can keep this for now, but create a story to convert to a brute force config file in the multi-model epic, and put a TODO tag for these [0] lines

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: Fixed

model_analyzer/config/input/config_command.py
nv-braf: Using set_value()

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: Understood and fixed

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Returns
+        -------
+        RunConfig
+            The next RunConfig generated by this class
+        """
+        yield from self._execute_quick_search()
+        yield from self._sweep_concurrency_over_top_results()
+
+    def _execute_quick_search(self):
```
tgerdesnv, line 103: type check return val

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -0,0 +1,195 @@
...
+            skip_default_config: bool) -> BruteRunConfigGenerator:
+        return BruteRunConfigGenerator(
+            config=new_config,
+            gpus=self._gpus,
+            models=self._models,
+            client=self._client,
+            model_variant_name_manager=self._model_variant_name_manager,
+            skip_default_config=skip_default_config)
+
+    def _set_search_mode(self, config: ConfigCommandProfile):
```
tgerdesnv, line 148: type check ret val

tgerdesnv (changes requested):

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: I've fixed this

tgerdesnv (dismissed):

tgerdesnv (approved):

PR #509, opened 2022-08-31
More experiment sweep changes

tgerdesnv:
Major changes to the ability to sweep through experiments of quick search algo:

sweep_checkpoints.py:
- New classes to hold data
- Ability to sweep through 3 modes:
  - max throughput
  - max throughput with latency budget
  - min latency with throughput floor

checkpoint_experiment_data: 
- Ability to load only measurements visible to algorithm
- Clean up map printing

config_command_experiment:
- Ability to set minimum max batch size index 

evaluate_config_generator:
- Patch loading checkpoint so it is ok with the lack of mvnm

experiment_evaluator:
- Calculate percentile differently for minimize-latency case

generator_experiment_factory:
- Patch some functions and then use the real factory

main.py:
- no longer have generator as an option, as we use the real CLI option

minimize_latency.yml:
- Basic config to set objective to minimize latency

run_config_generator_factory:
- Hook to patch in dimension changes for instance_count and max_batch_size


experiments/checkpoint_experiment_data.py
```diff
@@ -97,9 +112,38 @@ def _print_map(self):
...
+        ma_key = f"instance_count=3,max_batch_size=1"
+        pa_key = "8"
+        measurement = self._get_run_config_measurement_from_keys(ma_key,
+                                                                 pa_key,
+                                                                 skip_warn=True)
+        if not measurement:
+            ret = True
+        return ret
+
+    def _are_keys_visable_to_algorithm(self, ma_key, pa_key):
```
nv-braf, line 141: add `-> bool`

experiments/checkpoint_experiment_data.py
```diff
@@ -97,9 +112,38 @@ def _print_map(self):
...
-            print(row_str)
+
+            # Print at least the first 7 rows, and then stop when we hit an empty row
+            # (Some databases don't have data for the first 6 rows)
+            if row_has_data or i < 7:
+                print(row_str)
+            else:
+                break
+
+    def _has_exponential_inst_count(self) -> bool:
```
nv-braf, line 129: Why 3? Just because it's the first non-exponent? If this isn't present does this guarantee all the remaining instances will be exponents?

experiments/evaluate_config_generator.py
```diff
@@ -78,3 +84,18 @@ def _run_generator(self, cg):
                 run_config, run_config_measurement)
 
             cg.set_last_results([run_config_measurement])
+
+    def _patch_checkpoint_load(self):
```
nv-braf, line 88: I don't understand this code. Is this a method inside of a method that you are patching?

nv-braf (commented):
RCG factory looks clean. Did a cursory review of the experiments work, but don't understand everything you're doing fully.

experiments/checkpoint_experiment_data.py
tgerdesnv: Yeah, in the databases, it is either 1,2,3,4,5,... or 1,2,4,8,16,...
I picked the first one to make sure early exit didn't impact

experiments/evaluate_config_generator.py
tgerdesnv: So, the databases don't have a mvnm.  If we run without these changes it hits a key error.

Before, to run these experiments, we had to comment out the code that loaded the mvnm from the database.  This is a proper workaround that patches the code (basically it patches the checkpoint data) to have an empty mvnm.

nv-braf (approved):

PR #510, opened 2022-08-31
Add slow mode support

tgerdesnv:
Results show it is slower, but more accurate, as expected:

**Baseline (without these changes):**
normal Average/Median/Mode percentiles = 0.93 / 0.98 / 1.00
normal Average/Median/Mode measurements = 11.85 / 11.50 / 7.00
normal: 10 out of 128 are below the cutoff of 70.0%:

latency_budget Average/Median/Mode percentiles = 0.94 / 0.97 / 1.00
latency_budget Average/Median/Mode measurements = 10.47 / 10.00 / 4.00
latency_budget: 4 out of 127 are below the cutoff of 70.0%:

min_throughput Average/Median/Mode percentiles = 0.94 / 1.00 / 1.00
min_throughput Average/Median/Mode measurements = 5.55 / 5.00 / 4.00
min_throughput: 9 out of 125 are below the cutoff of 70.0%:

**New (with these changes):**
normal Average/Median/Mode percentiles = 0.95 / 0.99 / 1.00
normal Average/Median/Mode measurements = 13.23 / 12.00 / 7.00
normal: 7 out of 128 are below the cutoff of 70.0%:

latency_budget Average/Median/Mode percentiles = 0.97 / 1.00 / 1.00
latency_budget Average/Median/Mode measurements = 11.74 / 10.00 / 10.00
latency_budget: 2 out of 127 are below the cutoff of 70.0%:

min_throughput Average/Median/Mode percentiles = 0.97 / 1.00 / 1.00
min_throughput Average/Median/Mode measurements = 6.55 / 5.00 / 4.00
min_throughput: 3 out of 125 are below the cutoff of 70.0%:

### DIFF
--- a/model_analyzer/config/generate/neighborhood.py
+++ b/model_analyzer/config/generate/neighborhood.py
@@ -497,12 +497,6 @@ class Neighborhood:
         return num_uncovered
 
     def _is_slow_mode(self):
-        """ 
-        Returns true if any of the following are true:
-        - force_slow_mode() has been called
-        - Home is passing and anyone in the neighborhood is failing
-        - Home is failing and anyone in the neighborhood is passing
-        """
         if self._force_slow_mode:
             return True
 
--- a/model_analyzer/config/generate/neighborhood.py
+++ b/model_analyzer/config/generate/neighborhood.py
@@ -74,11 +74,11 @@ class Neighborhood:
         Returns true if enough coordinates inside of the neighborhood
         have been initialized. Else false
 
-        If the neighborhood is in slow mode, this means all next door neighbors
+        If the neighborhood is in slow mode, this means all adjacent neighbors
         must be visited
         """
         if self._is_slow_mode():
-            return self._are_all_next_door_neighbors_visited()
+            return self._are_all_adjacent_neighbors_visited()
         else:
             min_initialized = self._config.get_min_initialized()
             num_initialized = len(self._get_initialized_coordinates())
@@ -193,7 +193,7 @@ class Neighborhood:
         Based on the initialized coordinate values, pick an unvisited
         coordinate to initialize next.
 
-        If the neighborhood is in slow mode, only pick from within the next door neighbors
+        If the neighborhood is in slow mode, only pick from within the adjacent neighbors
         """
 
         if self._is_slow_mode():
@@ -202,7 +202,7 @@ class Neighborhood:
             return self._pick_fast_mode_coordinate_to_initialize()
 
     def _pick_slow_mode_coordinate_to_initialize(self):
-        for neighbor in self._get_all_next_door_neighbors():
+        for neighbor in self._get_all_adjacent_neighbors():
             if not self._is_coordinate_visited(neighbor):
                 return neighbor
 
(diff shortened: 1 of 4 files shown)


model_analyzer/config/generate/neighborhood.py
```diff
@@ -71,31 +73,81 @@ def enough_coordinates_initialized(self) -> bool:
...
+        home_measurement = self._get_home_measurement()
+
+        if home_measurement.is_passing_constraints():
+            vectors.append(Coordinate([0] * self._config.get_num_dimensions()))
+            measurements.append(home_measurement)
+
+        best_vector = vectors[0]
+        best_measurement = measurements[0]
+
+        for v, m in zip(vectors, measurements):
```
nv-braf, line 140: Can't you just make this a tuple and then call sort?

model_analyzer/config/generate/neighborhood.py
```diff
@@ -133,14 +185,30 @@ def _clip_vector_values(self, vector: Coordinate,
...
-                vector[i] = clip_value * vector[i]/max_value
+                vector[i] = clip_value * vector[i] / max_value
         return vector
 
     def pick_coordinate_to_initialize(self) -> Optional[Coordinate]:
         """
         Based on the initialized coordinate values, pick an unvisited
         coordinate to initialize next.
+
+        If the neighborhood is in slow mode, only pick from within the next door neighbors
```
nv-braf, line 196: Is "next door" defined anywhere? Would it be better to be explicit and say adjacent neighbors?

model_analyzer/config/generate/neighborhood.py
```diff
@@ -428,3 +495,64 @@ def _get_num_uncovered_values(
                 num_uncovered += 1
 
         return num_uncovered
+
+    def _is_slow_mode(self):
+        """ 
+        Returns true if any of the following are true:
```
nv-braf, line 501: This comment is somewhat redundant (as you're just explaining what the code does), but I'm not sure if it should be removed

nv-braf (commented):
Like all the unit tests!

model_analyzer/config/generate/neighborhood.py
tgerdesnv: updated

model_analyzer/config/generate/neighborhood.py
nv-braf: Lines 137-143 can be replaced with one line: 
best_vector, best_measurement = zip(vectors, measurements).sort()[-1]

(or you can change the sort to be descending so that best is the first element, if that is more readable)

tgerdesnv: The zip has to be measurements first (since that is what we want to sort on, or else we need to do extra stuff to tell it to sort differently)

Zip doesn't have sort, but sorted() supports it.

Thus, the line is 
```
        _, best_vector = sorted(zip(measurements, vectors))[-1]
```

I have updated it to that, and confirmed the results of the entire sweep are still exactly the same

nv-braf (dismissed):

nv-braf (approved):

PR #512, opened 2022-09-02
only use one coordinate data

tgerdesnv:
No longer create a coordinateData inside of neighborhood. Instead, pass it in. Without this change, we were ignoring some measurements and wasting time gathering new different measurements.

Also, update some CoordinateData concepts around visiting/measuring to differentiate between home visits vs random coordinate measurements

Results: Significant drop off in the number of measurements taken, but much of this was probably fake savings (we would re-measure a coordinate, but the ResultData had a cache so it didn't cost us anything to re-measure). However, this still looks really good.

**Old Results (current main with slow mode changes):**
normal Average/Median/Mode percentiles = 0.95 / 0.99 / 1.00
normal Average/Median/Mode measurements = 13.23 / 12.00 / 7.00
normal: 7 out of 128 are below the cutoff of 70.0%:

latency_budget Average/Median/Mode percentiles = 0.97 / 1.00 / 1.00
latency_budget Average/Median/Mode measurements = 11.74 / 10.00 / 10.00
latency_budget: 2 out of 127 are below the cutoff of 70.0%:

min_throughput Average/Median/Mode percentiles = 0.97 / 1.00 / 1.00
min_throughput Average/Median/Mode measurements = 6.55 / 5.00 / 4.00
min_throughput: 3 out of 125 are below the cutoff of 70.0%:

**New Results:**
normal Average/Median/Mode percentiles = 0.95 / 1.00 / 1.00
normal Average/Median/Mode measurements = 9.97 / 10.00 / 5.00
normal: 6 out of 128 are below the cutoff of 70.0%:

latency_budget Average/Median/Mode percentiles = 0.98 / 1.00 / 1.00
latency_budget Average/Median/Mode measurements = 10.54 / 10.00 / 3.00
latency_budget: 1 out of 127 are below the cutoff of 70.0%:

min_throughput Average/Median/Mode percentiles = 0.98 / 1.00 / 1.00
min_throughput Average/Median/Mode measurements = 6.34 / 5.00 / 4.00
min_throughput: 2 out of 123 are below the cutoff of 70.0%:

### DIFF
--- a/model_analyzer/config/generate/coordinate_data.py
+++ b/model_analyzer/config/generate/coordinate_data.py
@@ -26,6 +26,7 @@ class CoordinateData:
     def __init__(self):
         self._measurements = {}
         self._visit_counts = {}
+        self._is_measured = {}
 
     def get_measurement(
             self, coordinate: Coordinate) -> Optional[RunConfigMeasurement]:
@@ -42,6 +43,20 @@ class CoordinateData:
         """
         key: Tuple[Coordinate, ...] = tuple(coordinate)
         self._measurements[key] = measurement
+        self._is_measured[key] = True
+
+    def is_measured(self, coordinate: Coordinate) -> bool:
+        """
+        Returns true if a measurement has been set for the given Coordinate
+        """
+        key: Tuple[Coordinate, ...] = tuple(coordinate)
+        return self._is_measured.get(key, False)
+
+    def has_valid_measurement(self, coordinate: Coordinate) -> bool:
+        """
+        Returns true if there is a valid measurement for the given Coordinate
+        """
+        return self.get_measurement(coordinate) is not None
 
     def reset_measurements(self):
         """
--- a/model_analyzer/config/generate/quick_run_config_generator.py
+++ b/model_analyzer/config/generate/quick_run_config_generator.py
@@ -90,7 +90,7 @@ class QuickRunConfigGenerator(ConfigGeneratorInterface):
 
         self._neighborhood = Neighborhood(
             self._search_config.get_neighborhood_config(),
-            self._home_coordinate)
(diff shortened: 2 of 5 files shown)


model_analyzer/config/generate/coordinate_data.py
```diff
@@ -26,6 +26,7 @@ class CoordinateData:
     def __init__(self):
         self._measurements = {}
         self._visit_counts = {}
+        self._is_measured = {}
```
nv-braf, line 29: Don't you need to define what these are for type checking to work correctly?

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -187,8 +183,8 @@ def _update_best_measurement(self, measurement: RunConfigMeasurement):
                 self._best_coordinate = self._coordinate_to_measure
                 self._best_measurement = measurement
 
-    def _get_last_results(self) -> RunConfigMeasurement:
-        return self._neighborhood.coordinate_data.get_measurement(
+    def _get_last_results(self) -> Optional[RunConfigMeasurement]:
```
nv-braf, line 186: Why is this now optional? Is it because it can now return None?

model_analyzer/config/generate/coordinate_data.py
tgerdesnv: type checking passed 🤷‍♂️

model_analyzer/config/generate/quick_run_config_generator.py
tgerdesnv: It always could be None. Some code changes I did made the type checking unhappy.

nv-braf (approved):

PR #513, opened 2022-09-06
Fix live quick search

tgerdesnv:
Fix issue where GPU type was hardcoded
Fix issue where we would report an error when starting a fresh run


model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -87,8 +87,20 @@ def get_configs(self) -> Generator[RunConfig, None, None]:
         RunConfig
             The next RunConfig generated by this class
         """
+
```
nv-braf, line 90: Instead of `("")` why not just include `\n` before after the string

PR #516, opened 2022-09-07
Document quick search

nv-braf:


docs/config_search.md
```diff
@@ -150,7 +151,19 @@ as the range for the `max_batch_size` to `[1]`, it will no longer be a valid
 Triton Model Configuration.
 
 The configuration sweep described above, will sweep through 8 configs = (2
-`max_batch_size`) * (2 `max_queue_delay_microseconds`) * (2 `instance_group`) values.
+`max_batch_size`) _ (2 `max_queue_delay_microseconds`) _ (2 `instance_group`) values.
```
tgerdesnv, line 154: underscores now?

docs/config_search.md
```diff
@@ -131,13 +133,12 @@ run_config_search_disable: True
...
-        dynamic_batching:
-            max_queue_delay_microseconds: [200, 300]
-        instance_group:
-        -
-            kind: KIND_GPU
-            count: [1, 2]
+      max_batch_size: [6, 8]
+      dynamic_batching:
+        max_queue_delay_microseconds: [200, 300]
+      instance_group:
```
tgerdesnv, line 139: is this equivalent/correct?

docs/config_search.md
```diff
@@ -150,7 +151,19 @@ as the range for the `max_batch_size` to `[1]`, it will no longer be a valid
...
+## Quick Search
+
+Quick search is enabled by adding the parameter `--run-config-search-mode quick` to the CLI.
+
+It uses a hill climbing algorithm to search the configuration space, looking for
+the maximal objective value within the specified constraints. In the majority of cases
+this will find greater than 90% of the maximum objective value (that could be found using a brute force search),
+while needing to search less than 10% of the configuration space.
+
+We will then sweep the top-N configurations found (specified by `--num-configs-per-model`) over the default concurrency
```
tgerdesnv, line 165: We vs It

tgerdesnv, line 165: This seems like a technical detail that the user doesn't need to know

docs/config_search.md
```diff
@@ -19,6 +19,10 @@ limitations under the License.
 Model Analyzer's `profile` subcommand supports **automatic** and **manual**
```
tgerdesnv, line 19: Should we differentiate between brute and quick at the top level? And then break brute down into automatic vs manual?

tgerdesnv (changes requested):

docs/config_search.md
nv-braf: This is the result of YAPF formatting. It looks identical when viewing the document

docs/config_search.md
nv-braf: This is the result of YAPF formatting. I don't know if it is equivalent, but it is now correct (this is the formatting I see in yaml files I use when running live tests).

docs/config_search.md
nv-braf: Fixed

nv-braf: Removed

docs/config_search.md
nv-braf: I don't think we need to. Brute is our default, so the user doesn't need to explicitly specifying, therefore I don't think we need to call this out, as our examples work correctly (now if quick becomes our default, then that's a different story...)

docs/config_search.md
tgerdesnv: I think you replied to the wrong comment, and/or did a bad search and replace. These should not be underscores.

image

nv-braf: You are correct. This formatting did need to be fixed. YAPF converted it into underscores, but `/*` preserves them.

docs/config_search.md
tgerdesnv: This first paragraph should introduce the different modes and provide links to the more detailed sections below. 

At a high level, we have Brute and Quick.  Brute has subcategories automatic brute and manual brute. I think the sections should be renamed as such.

docs/config_search.md
```diff
@@ -150,7 +151,19 @@ as the range for the `max_batch_size` to `[1]`, it will no longer be a valid
...
+
+It uses a hill climbing algorithm to search the configuration space, looking for
+the maximal objective value within the specified constraints. In the majority of cases
+this will find greater than 90% of the maximum objective value (that could be found using a brute force search),
+while needing to search less than 10% of the configuration space.
+
+It will then sweep the top-N configurations found (specified by `--num-configs-per-model`) over the default concurrency
+range before generation of the summary reports.
 
 ### Examples of Additional Model Config Parameters
```
tgerdesnv, line 168: This section should stay part of manual brute section

docs/config_search.md
nv-braf: Done

docs/config_search.md
```diff
@@ -16,10 +16,14 @@ limitations under the License.
 
 # Model Config Search
 
-Model Analyzer's `profile` subcommand supports **automatic** and **manual**
+Model Analyzer's `profile` subcommand supports [Brute](config_search.md#brute-search-mode) and [Quick](config_search.md#quick-search-mode) search modes.
+
+## Brute Search Mode
+
+Model Analyzer's brute search mode supposts **automatic** and **manual**
```
tgerdesnv, line 23: This should explain which conditions lead to automatic vs manual, and the automatic and manual words should be links.

Also, typo "supposts"

docs/config_search.md
nv-braf: Fixed

docs/config_search.md
nv-braf: Fixed

docs/cli.md
```diff
@@ -86,8 +86,8 @@ $ model-analyzer profile -h
 
 Depending on the command line or YAML config options provided, the `profile`
 subcommand will either perform a
-[manual](./config_search.md#manual-configuration-search) or [automatic
-search](./config_search.md#automatic-configuration-search) over perf analyzer
+[manual](./config_search.md#manual-configuration-search), [automatic](./config_search.md#automatic-configuration-search), or
```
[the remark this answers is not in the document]
tgerdesnv: actually the examples below mention automatic and manual, so maybe not

docs/config_search.md
```diff
@@ -16,26 +16,24 @@ limitations under the License.
...
-Analyzer. To enable manual configuration search, the `--run-config-search-disable` flag
-should be provided in the CLI.
-
-### Automatic Configuration Search
+### Automatic Brute Search
 
 Automatic configuration search is the default behavior when running Model
-Analyzer. This mode is enabled when there is not any parameters specified for the
-`model_config_parameters` section of the Model Analyzer Config. The parameters
+Analyzer without specifying manually specifying what values to search. The parameters
```
nv-braf, line 31: "specifying manually specifying" -> "manually specifying"

tgerdesnv (approved):

PR #519, opened 2022-09-08
Capture GPU metrics from PA

nv-braf:
Captures new GPU metrics that PA provides and puts them into Records

Added a bunch of unit tests to cover the various scenarios of multi and sparse GPU reporting

### DIFF
--- a/model_analyzer/perf_analyzer/perf_analyzer.py
+++ b/model_analyzer/perf_analyzer/perf_analyzer.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Any, Dict, List, Union, Tuple
 from model_analyzer.model_analyzer_exceptions \
     import TritonModelAnalyzerException
 from model_analyzer.record.types.perf_latency_avg import PerfLatencyAvg
@@ -33,6 +34,7 @@ from model_analyzer.record.types.perf_server_compute_infer \
 from model_analyzer.record.types.perf_server_compute_output \
     import PerfServerComputeOutput
 
+from model_analyzer.record.record import Record
 from model_analyzer.record.types.gpu_utilization import GPUUtilization
 from model_analyzer.record.types.gpu_power_usage import GPUPowerUsage
 from model_analyzer.record.types.gpu_used_memory import GPUUsedMemory
@@ -69,18 +71,18 @@ class PerfAnalyzer:
 
     METRIC_TAG,                        CSV_STRING,             RECORD_CLASS,             REDUCTION_FACTOR = 0, 1, 2, 3
     perf_metric_table = [
-        ["perf_latency_avg",           "Avg latency",           PerfLatencyAvg,          1000],
-        ["perf_latency_p90",           "p90 latency",           PerfLatencyP90,          1000],
-        ["perf_latency_p95",           "p95 latency",           PerfLatencyP95,          1000],
-        ["perf_latency_p99",           "p99 latency",           PerfLatencyP99,          1000],
-        ["perf_throughput",            "Inferences/Second",     PerfThroughput,             1],
-        ["perf_client_send_recv",      "request/response",      PerfClientSendRecv,      1000],
-        ["perf_client_send_recv",      "send/recv",             PerfClientSendRecv,      1000],
-        ["perf_client_response_wait",  "response wait",         PerfClientResponseWait,  1000],
-        ["perf_server_queue",          "Server Queue",          PerfServerQueue,         1000],
-        ["perf_server_compute_infer",  "Server Compute Infer",  PerfServerComputeInfer,  1000],
-        ["perf_server_compute_input",  "Server Compute Input",  PerfServerComputeInput,  1000],
-        ["perf_server_compute_output", "Server Compute Output", PerfServerComputeOutput, 1000]
+        ["perf_latency_avg",           "Avg latency",           PerfLatencyAvg,          "1000"],
+        ["perf_latency_p90",           "p90 latency",           PerfLatencyP90,          "1000"],
+        ["perf_latency_p95",           "p95 latency",           PerfLatencyP95,          "1000"],
+        ["perf_latency_p99",           "p99 latency",           PerfLatencyP99,          "1000"],
+        ["perf_throughput",            "Inferences/Second",     PerfThroughput,             "1"],
+        ["perf_client_send_recv",      "request/response",      PerfClientSendRecv,      "1000"],
+        ["perf_client_send_recv",      "send/recv",             PerfClientSendRecv,      "1000"],
+        ["perf_client_response_wait",  "response wait",         PerfClientResponseWait,  "1000"],
+        ["perf_server_queue",          "Server Queue",          PerfServerQueue,         "1000"],
+        ["perf_server_compute_infer",  "Server Compute Infer",  PerfServerComputeInfer,  "1000"],
+        ["perf_server_compute_input",  "Server Compute Input",  PerfServerComputeInput,  "1000"],
+        ["perf_server_compute_output", "Server Compute Output", PerfServerComputeOutput, "1000"]
     ]
 
     gpu_metric_table = [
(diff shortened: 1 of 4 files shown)


model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -402,26 +427,104 @@ def _parse_outputs(self, metrics):
...
+        return gpu_metrics, gpu_values
+
+    def _add_gpu_metrics_to_records(self, gpu_metrics: List[List[Any]],
+                                    gpu_values: List[List[Tuple[str, str]]],
+                                    perf_records: List[Record]):
+        for i, gpu_metric in enumerate(gpu_metrics):
+            for gpu_tuple in gpu_values[i]:
+                perf_records.append(gpu_metric[PerfAnalyzer.RECORD_CLASS](
+                    value=float(gpu_tuple[PerfAnalyzer.GPU_METRIC_VALUE]),
+                    device_uuid=int(gpu_tuple[PerfAnalyzer.GPU_METRIC_UUID])))
```
tgerdesnv, line 520: uuid is not an integer. It is a string:

```
01:57:38 [Model Analyzer] Using GPU 0 NVIDIA TITAN RTX with UUID GPU-aaf4fea0-a814-e92a-6162-5852d329510f
```

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -402,26 +427,104 @@ def _parse_outputs(self, metrics):
...
+            if self._is_metric_requested_and_in_row(perf_metric,
+                                                    requested_metrics,
+                                                    row_metrics):
+                value = float(row_metrics[str(
+                    perf_metric[PerfAnalyzer.CSV_STRING])])
+                reduction_factor = float(
+                    str(perf_metric[PerfAnalyzer.REDUCTION_FACTOR]))
+                perf_value = value / reduction_factor
+
+                perf_metrics.append(perf_metric)
```
tgerdesnv, line 464: why append into two different lists that then need to be handled in lockstep? Can't you just call the record constructor here with perf_metric and perf_value? And return a list of records directly from this function?

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -402,26 +427,104 @@ def _parse_outputs(self, metrics):
...
+                values.append(perf_value)
+
+        return (perf_metrics, values)
+
+    def _add_perf_metrics_to_records(
```
tgerdesnv, line 469: I don't like perf_records being an input and an output. You have two functions that return a list of records. You don't have to have the appending hidden in the function.

perf_records = extract_perf_records_from_row
gpu_records = extract_gpu_records_from_row
combined_records = perf_records + gpu_records
return combined_records

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -402,26 +427,104 @@ def _parse_outputs(self, metrics):
...
+                gpu_metric_string_tuples = gpu_metric_string.split(';')
+
+                gpu_tuples = []
+                for gpu_metric_string_tuple in gpu_metric_string_tuples:
+                    gpu_metric_tuple = gpu_metric_string_tuple.split(':')
+                    gpu_tuples.append(
+                        (gpu_metric_tuple[PerfAnalyzer.GPU_METRIC_UUID],
+                         gpu_metric_tuple[PerfAnalyzer.GPU_METRIC_VALUE]))
+
+                gpu_metrics.append(gpu_metric)
```
tgerdesnv, line 508: just create the record here

tests/test_perf_analyzer.py
```diff
@@ -237,7 +243,11 @@ def test_run(self):
         self.server.start()
         self.client.wait_for_server_ready(num_retries=1)
 
-        pa_csv_mock = """Concurrency,Inferences/Second,Client Send,Network+Server Send/Recv,Server Queue,Server Compute Input,Server Compute Infer,Server Compute Output,Client Recv,p50 latency,p90 latency,p95 latency,p99 latency,Avg latency,request/response,response wait\n1,46.8,2,187,18,34,65,16,1,4600,4700,4800,4900,5000,3,314"""
+        pa_csv_mock = """Concurrency,Inferences/Second,Client Send,Network+Server Send/Recv,Server Queue,Server Compute Input,Server Compute Infer,Server Compute Output,"""
+        pa_csv_mock += """Client Recv,p50 latency,p90 latency,p95 latency,p99 latency,Avg latency,request/response,response wait,"""
+        pa_csv_mock += """Avg GPU Utilizations,Avg GPU Power Usages,Max GPU Memory Usages,Total GPU Memory Usages\n"""
+        pa_csv_mock += """1,46.8,2,187,18,34,65,16,1,4600,4700,4800,4900,5000,3,314,"""
+        pa_csv_mock += """1:80.9;2:90.1;3:74.5;,1:91.2;2:100;,1:1000;,7:1500"""
```
tgerdesnv, line 250: update this to use real-ish UUID values (at least make them strings)

tgerdesnv (changes requested):
Testing looks good. I think the main code can be cleaned up. I left some suggestions

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -402,26 +427,104 @@ def _parse_outputs(self, metrics):
...
+            requested_metrics, row_metrics)
+        perf_records = self._add_perf_metrics_to_records(
+            perf_metrics, perf_values, perf_records)
+
+        gpu_metrics, gpu_values = self._find_requested_gpu_metrics(
+            requested_metrics, row_metrics)
+        perf_records = self._add_gpu_metrics_to_records(gpu_metrics, gpu_values,
+                                                        perf_records)
+
+        return perf_records
```
tgerdesnv, line 447: The consuming code is going to have to split these back out to be perf_records and gpu_records. It might be best to just store them separately

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: fixed

tests/test_perf_analyzer.py
nv-braf: fixed

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: It was my attempt not to have a method do two different things, but yes I agree it's much simpler to just find and create the record in one method. It's fixed now.

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: Done

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: Done

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: That very well might be our end solution, but that would break upstream code, which I'd rather not touch as part of this story. It seems a trivial enough change to make in the follow on story if it's needed.

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -402,26 +427,75 @@ def _parse_outputs(self, metrics):
...
+        for gpu_metric in PerfAnalyzer.gpu_metric_table:
+            if self._is_metric_requested_and_in_row(gpu_metric,
+                                                    requested_metrics,
+                                                    row_metrics):
+                gpu_metric_string = row_metrics[str(
+                    gpu_metric[PerfAnalyzer.CSV_STRING])]
+
+                # Return immediately if PA didn't provide data
+                if not gpu_metric_string:
+                    return gpu_records
```
tgerdesnv, line 476: is it possible that one metric header is missing? Should this be continue instead of directly return?

nv-braf: That's a good point. I've made the change.

tgerdesnv (approved):

PR #525, opened 2022-09-14
Replace result heaps with lists

nv-braf:
ResultHeap -> SortedResults

Removed deletion and re-creation of of SortedLists when profile completes.
Added unit testing and re-ran live quick profile to ensure summary PDF is correct on the initial run


model_analyzer/result/sorted_results.py
```diff
@@ -35,53 +40,55 @@ def empty(self):
         """
         Returns
         -------
-        True if this heap has no results
+        True if this list has no results
```
tgerdesnv, line 43: refer to the class/object, not to "this list"

model_analyzer/result/sorted_results.py
```diff
@@ -35,53 +40,55 @@ def empty(self):
...
 
         return not bool(self._sorted_results)
 
-    def results(self):
+    def results(self) -> List[RunConfigResult]:
         """
         Returns
         -------
-        All the results in this result heap
+        All the results in this result list
```
tgerdesnv, line 53: refer to the class/object, not to "this list"

model_analyzer/result/sorted_results.py
```diff
@@ -35,53 +40,55 @@ def empty(self):
...
 
         Returns
         -------
         ModelResult
-            The next best result in this heap
+            The next best result in this list
         """
 
-        return heapq.heappop(self._sorted_results)
+        return self._sorted_results.pop(0)
```
tgerdesnv, line 89: Does a pop still make sense here? Could this be updated to return an iterator?

model_analyzer/result/sorted_results.py
```diff
@@ -91,37 +98,66 @@ def top_n_results(self, n):
...
-
         if len(self._passing_results) == 0:
             logger.warning(
                 f"Requested top {n} configs, but none satisfied constraints. "
                 "Showing available constraint failing configs for this model.")
 
             if n == -1:
-                return heapq.nsmallest(len(self._failing_results),
-                                       self._failing_results)
+                return self._failing_results
```
tgerdesnv, line 111: What is this special -1 case?

model_analyzer/result/sorted_results.py
```diff
@@ -91,37 +98,66 @@ def top_n_results(self, n):
...
                 f"but found only {len(self._passing_results)} passing configs. "
                 "Showing all available constraint satisfying configs for this model."
             )
 
-        return heapq.nsmallest(min(n, len(self._passing_results)),
-                               self._passing_results)
+        result_len = min(n, len(self._passing_results))
+        return self._passing_results[0:result_len]
+
+    def _find_existing_run_config(
```
tgerdesnv, line 134: this name implies it is finding a run config. Is it?

tests/test_sorted_results.py
```diff
@@ -43,42 +43,66 @@ def test_empty(self):
...
                 construct_run_config_result(
                     avg_gpu_metric_values=avg_gpu_metrics,
                     avg_non_gpu_metric_values_list=[avg_non_gpu_metrics],
                     comparator=self.result_comparator))
 
-        results = self.result_heap.results()
+        results = self.sorted_results.results()
         self.assertEqual(len(results), 10)
 
+    def test_add_results_same_config(self):
```
tgerdesnv, line 65: I'd like to see a test that shows/protects the dynamic sorting.  Have 2 entries where A > B.  Add something to B so that now B > A.

tgerdesnv (changes requested):

model_analyzer/result/sorted_results.py
nv-braf: Fixed

model_analyzer/result/sorted_results.py
nv-braf: Fixed

model_analyzer/result/sorted_results.py
nv-braf: See the comment in the method description. I'll boy scout this to remove the magic number and replace it with a constant

model_analyzer/result/sorted_results.py
nv-braf: Yes. Well sort of...it's getting the RunConfigResult, which is the RunConfig + other necessary stuff  (comparator + constraints). 

I omitted the extra result part to try and not have so many things with result in the name. Type checking indicates what this actually is, so I thought the brevity was okay in this case.

tests/test_sorted_results.py
nv-braf: Agreed, bad omission on my part. I've added a unit test to cover this case.

model_analyzer/result/sorted_results.py
tgerdesnv: Is it even possible to hit the -1 case in live code? I suppose it is safer to leave it in just in case.

However, it needs to be fully tested then. Our unit test coverage implies that the -1 case is covered for passing, but not for failing. Can you add a case with only failing tests with -1?

tgerdesnv (changes requested):

model_analyzer/result/sorted_results.py
nv-braf: I've added a case that exercises the failing only case.

model_analyzer/result/sorted_results.py
nv-braf: Like we discussed I've removed the passing/failing lists and now create them only when needed. I've also updated all the names to reflect this and added unit tests to cover the case where a result goes from failing to passing.

model_analyzer/result/sorted_results.py
```diff
@@ -0,0 +1,150 @@
...
+        existing_run_config_result = self._find_existing_run_config_result(
+            run_config_result)
+
+        if existing_run_config_result:
+            self._add_measurements_to_existing_run_config_result(
+                existing_run_config_result, run_config_result)
+        else:
+            self._add_new_run_config_result(run_config_result)
+
+    def top_n_results(self, n: int) -> List[RunConfigResult]:
```
tgerdesnv, line 70: Can you refactor this function? I see 4 return points, and the function is basically the same in the first half and the second half, except for which list it is operating on. You could create a helper function and pass the correct list (passing vs failing) in (as well as a bool to say if it is the passing list, so you can tweak some prints)

model_analyzer/result/sorted_results.py
```diff
@@ -0,0 +1,150 @@
...
+            The number of  top results
+            to retrieve
+
+        Returns
+        -------
+        list of RunConfigResults
+            The n best results for this model,
+            must all be passing results
+        """
+        self._run_config_results.sort()
```
tgerdesnv, line 84: move this sort into the function on the next line, so we are guaranteed that anyone who uses that function will get sorted results.

If you need to, for clarity, you can rename the function to indicate it does a sort

tgerdesnv (changes requested):

model_analyzer/result/sorted_results.py
nv-braf: I don't understand. The other function is private. I believe the sort is in the correct location.

model_analyzer/result/sorted_results.py
nv-braf: I could, but that seems like overkill. I've done a minimal refactor to make it more readable.

model_analyzer/result/sorted_results.py
nv-braf: I've moved it, but I don't understand why it needs to be in the private method.

model_analyzer/result/sorted_results.py
nv-braf: I've refactored

tgerdesnv (approved):

PR #537, opened 2022-09-30
Concurrency sweep early backoff

nv-braf:
Add early backoff functionality to quick search w/ concurrency sweep

### DIFF
--- a/model_analyzer/config/generate/perf_analyzer_config_generator.py
+++ b/model_analyzer/config/generate/perf_analyzer_config_generator.py
@@ -156,9 +156,6 @@ class PerfAnalyzerConfigGenerator(ConfigGeneratorInterface):
             self._last_results = measurement
             self._concurrency_results.extend(measurement)
 
-    def get_last_results(self) -> Optional[RunConfigMeasurement]:
-        return self._last_results[-1]
-
     def _create_concurrency_list(self, cli_config: ConfigCommandProfile,
                                  model_parameters: dict) -> List[int]:
         if model_parameters['concurrency']:
--- a/model_analyzer/config/generate/perf_analyzer_config_generator.py
+++ b/model_analyzer/config/generate/perf_analyzer_config_generator.py
@@ -36,12 +36,9 @@ class PerfAnalyzerConfigGenerator(ConfigGeneratorInterface):
     earlier depending on results that it receives
     """
 
-    def __init__(self, 
-                cli_config: ConfigCommandProfile, 
-                model_name: str, 
-                model_perf_analyzer_flags: dict,
-                model_parameters: dict, 
-                early_exit_enable: bool) -> None:
+    def __init__(self, cli_config: ConfigCommandProfile, model_name: str,
+                 model_perf_analyzer_flags: dict, model_parameters: dict,
+                 early_exit_enable: bool) -> None:
         """
         Parameters
         ----------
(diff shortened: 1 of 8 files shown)


tgerdesnv (changes requested):

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -80,6 +81,9 @@ def set_last_results(
             self, measurements: List[Optional[RunConfigMeasurement]]) -> None:
         self._rcg.set_last_results(measurements)
 
+    def get_last_results(self) -> Optional[RunConfigMeasurement]:
```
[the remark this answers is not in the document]
nv-braf: That didn't seem very clean to me. I thought it better to just keep the measurement stored once in the lower level generator and then add an accessor function to re-acquire it when needed.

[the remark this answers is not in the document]
tgerdesnv: The results are already passed to this object via set_last_results. That is the clean place to grab/keep them if you need them.  Just append them to a local list, and clear it at the start of each concurrency sweep.

Instead what is happening is the results are being passed through inactive(!) generators RCG->MCG->PACG, and then you are creating a new API to get the data from that low level object.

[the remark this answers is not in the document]
nv-braf: I've removed get_last_results(). Please re-review

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -77,8 +74,8 @@ def __init__(self,
         self._generator_started = False
 
         self._last_results: List[RunConfigMeasurement] = []
-        self._concurrency_results: List[RunConfigMeasurement] = []
-        self._batch_size_results: List[RunConfigMeasurement] = []
+        self._concurrency_results: List[Optional[RunConfigMeasurement]] = []
```
tgerdesnv, line 77: Is this still correct? I don't think it is

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -181,8 +206,11 @@ def _step(self) -> None:
             self._step_batch_size()
 
     def _add_best_throughput_to_batch_sizes(self) -> None:
-        if self._concurrency_results:
-            best = max(self._concurrency_results)
+        if self._concurrency_results and not any(
+                x is None for x in self._concurrency_results):
+            # type is List[Optional[RCM]], but static checking doesn't understand
```
tgerdesnv, line 211: is this still correct/needed?

tgerdesnv (changes requested):

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: Yes

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: Ye

tgerdesnv (dismissed):

PR #540, opened 2022-10-06
Multi-model and Quick Search Config Checking

nv-braf:
Checks that quick search is specified when running multi-model and than no search or list values are used when in quick search.

Added unit testing and refactored the code in config command.

### DIFF
--- a/model_analyzer/config/input/config_command.py
+++ b/model_analyzer/config/input/config_command.py
@@ -95,14 +95,13 @@ class ConfigCommand:
             this exception
         """
 
-        # Config file has been specified
         yaml_config = self._load_yaml_config(args)
         self._check_for_illegal_config_settings(args, yaml_config)
         self._set_field_values(args, yaml_config)
         self._preprocess_and_verify_arguments()
         self._autofill_values()
 
-    def _load_yaml_config(self, args: Namespace) -> Dict[str, List]:
+    def _load_yaml_config(self, args: Namespace) -> Optional[Dict[str, List]]:
         if 'config_file' in args:
             yaml_config = self._load_config_file(args.config_file)
             YamlConfigValidator.validate(yaml_config)
@@ -112,13 +111,14 @@ class ConfigCommand:
         return yaml_config
 
     def _check_for_illegal_config_settings(
-            self, args: Namespace, yaml_config: Dict[str, List]) -> None:
+            self, args: Namespace, yaml_config: Optional[Dict[str,
+                                                              List]]) -> None:
         self._check_for_duplicate_profile_models_option(args, yaml_config)
         self._check_for_multi_model_incompatability(args, yaml_config)
         self._check_for_quick_search_incompatability(args, yaml_config)
 
     def _set_field_values(self, args: Namespace,
-                          yaml_config: Dict[str, List]) -> None:
+                          yaml_config: Optional[Dict[str, List]]) -> None:
         for key, value in self._fields.items():
             self._fields[key].set_name(key)
             config_value = self._get_config_value(key, args, yaml_config)
@@ -164,10 +164,11 @@ class ConfigCommand:
                 yaml_config):
             return
 
-        self._check_search_mode(args, yaml_config)
+        self._check_multi_model_search_mode_incompatability(args, yaml_config)
 
-    def _check_search_mode(self, args: Namespace,
-                           yaml_config: Optional[Dict[str, List]]) -> None:
+    def _check_multi_model_search_mode_incompatability(
+            self, args: Namespace, yaml_config: Optional[Dict[str,
+                                                              List]]) -> None:
         if self._get_config_value('run_config_search_mode', args,
                                   yaml_config) != 'quick':
             raise TritonModelAnalyzerException(
--- a/model_analyzer/config/input/config_command.py
+++ b/model_analyzer/config/input/config_command.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Any
 from model_analyzer.model_analyzer_exceptions \
     import TritonModelAnalyzerException
 import yaml
@@ -133,8 +133,7 @@ class ConfigCommand:
 
     def _get_config_value(
             self, key: str, args: Namespace,
-            yaml_config: Optional[Dict[str,
-                                       List]]) -> Optional[Union[int, float]]:
+            yaml_config: Optional[Dict[str, List]]) -> Optional[Any]:
         if key in args:
             return getattr(args, key)
         elif yaml_config is not None and key in yaml_config:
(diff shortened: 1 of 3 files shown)


model_analyzer/config/input/config_command.py
```diff
@@ -95,40 +96,137 @@ def set_config_values(self, args):
         """
 
         # Config file has been specified
```
tgerdesnv, line 98: comment is stale

model_analyzer/config/input/config_command.py
```diff
@@ -95,40 +96,137 @@ def set_config_values(self, args):
...
+        yaml_config = self._load_yaml_config(args)
+        self._check_for_illegal_config_settings(args, yaml_config)
+        self._set_field_values(args, yaml_config)
+        self._preprocess_and_verify_arguments()
+        self._autofill_values()
+
+    def _load_yaml_config(self, args: Namespace) -> Dict[str, List]:
```
tgerdesnv, line 105: Does this pass type checking? We can return None

model_analyzer/config/input/config_command.py
```diff
@@ -95,40 +96,137 @@ def set_config_values(self, args):
...
+            self, args: Namespace, yaml_config: Optional[Dict[str,
+                                                              List]]) -> None:
+        if not self._get_config_value(
+                'run_config_profile_models_concurrently_enable', args,
+                yaml_config):
+            return
+
+        self._check_search_mode(args, yaml_config)
+
+    def _check_search_mode(self, args: Namespace,
```
tgerdesnv, line 169: this function needs a better more specific name. This is not a generic check -- it is very specific (including the print!)

model_analyzer/config/input/config_command.py
```diff
@@ -95,40 +96,137 @@ def set_config_values(self, args):
...
+            )
+        if max_batch_size or min_batch_size:
+            raise TritonModelAnalyzerException(
+                f'\nProfiling of models in quick search mode is not supported with min/max batch size search values.'
+                '\nPlease use brute search mode or remove batch size search values.'
+            )
+
+    def _check_no_list_values(self, args: Namespace,
+                              yaml_config: Optional[Dict[str, List]]) -> None:
+        concurrency = self._get_config_value('concurrency', args, yaml_config)
```
tgerdesnv, line 221: would this catch per-model concurrencies or batch sizes in a yaml config?

tests/test_config.py
```diff
@@ -1755,6 +1758,83 @@ def test_copy(self):
...
+
+        # Tests the case where no search mode is specified (default is brute)
+        with self.assertRaises(TritonModelAnalyzerException):
+            self._evaluate_config(args, yaml_content, subcommand='profile')
+
+        args.append('--run-config-search-mode')
+        args.append('brute')
+
+        with self.assertRaises(TritonModelAnalyzerException):
+            self._evaluate_config(args, yaml_content, subcommand='profile')
```
tgerdesnv, line 1781: add the passing case (where it properly uses quick)

tests/test_config.py
```diff
@@ -1755,6 +1758,83 @@ def test_copy(self):
...
+        Tests that run-config-search options raise exceptions 
+        in quick search mode
+        """
+        new_args = deepcopy(args)
+        new_args.append(rcs_string)
+        new_args.append('1')
+        with self.assertRaises(TritonModelAnalyzerException):
+            self._evaluate_config(new_args, yaml_content, subcommand='profile')
+
+    def _test_quick_search_list_value(self, args: Namespace,
```
tgerdesnv, line 1825: this only tests CLI. Should we also test yaml? Especially for any model-specific args. Same for previous test.

tgerdesnv (changes requested):
Looks good. I like the refactor. A few comments

model_analyzer/config/input/config_command.py
nv-braf: It did pass type checking, but yes, it should be Optional. I've made the change.

model_analyzer/config/input/config_command.py
nv-braf: Removed

model_analyzer/config/input/config_command.py
nv-braf: Updated name

model_analyzer/config/input/config_command.py
nv-braf: No, I don't think it would. I'll look into creating that test, though I'm not exactly sure how to look for that condition yet.

tests/test_config.py
tgerdesnv: agreed, although I think there is a hole in both the testing and the actual code for model-specific args.

```
profile_models:
  model_1:
    concurrency: 1,2,3,128
```

nv-braf: We only need model specific arg testing. We already have a test in place that is exercising the YAML code (test_conflicting_configs)

tests/test_config.py
nv-braf: Added

tests/test_config.py
nv-braf: Added model specific checks and unit testing. Please review the new code

model_analyzer/config/input/config_command.py
```diff
@@ -217,17 +218,38 @@ def _check_no_search_values(self, args: Namespace,
...
+            self, args: Namespace, yaml_config: Optional[Dict[str,
+                                                              List]]) -> None:
+        profile_models = self._get_config_value('profile_models', args,
+                                                yaml_config)
+
+        if type(profile_models) is str:
+            return
+
+        for model in profile_models.values():
+            if not 'parameters' in model:
```
tgerdesnv, line 246: parameters is for PA args. You have that covered. What about model_config_parameters?

```
profile_models:
  model_1:
    model_config_parameters:
      max_batch_size: 2
```

The existence of ANY model_config_parameters probably breaks us for quick search.

model_analyzer/config/input/config_command.py
```diff
@@ -94,41 +95,164 @@ def set_config_values(self, args):
...
+    def _load_yaml_config(self, args: Namespace) -> Optional[Dict[str, List]]:
         if 'config_file' in args:
             yaml_config = self._load_config_file(args.config_file)
             YamlConfigValidator.validate(yaml_config)
         else:
             yaml_config = None
 
+        return yaml_config
+
+    def _check_for_illegal_config_settings(
```
tgerdesnv, line 113: What happens if the user does `run_config_search_disable: True` in the yaml or in the CLI? Do we ignore brute/quick and just run the default config?

tgerdesnv (changes requested):

model_analyzer/config/input/config_command.py
nv-braf: Added per model checking

model_analyzer/config/input/config_command.py
nv-braf: We ignore the search disable. What is the correct thing to do here? Make this an error?

nv-braf: I've made this an error. Please review the new code.

tgerdesnv: Seems like we should honor the disable, but that would require larger changes (maybe in the factory?). Can you file a story to follow up?

tgerdesnv (approved):

model_analyzer/config/input/config_command.py
nv-braf: I've filed TMA-939

PR #541, opened 2022-10-06
Add early access documentation for multi-model

nv-braf:
Also, removing ```Will be ignored if server-launch-mode is not 'remote'``` from the 3 places it doesn't belong

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,6 @@ Triton Inference Server.
 
 ## Features
 
-- [Multi-Model Search](docs/config_search.md#multi-model-search-mode): **EARLY ACCESS** - Model Analyzer can help you
-  find the optimal settings when profiling multiple concurrent models, utilizing our Quick Search alogrithm
 - [Brute and Quick search](docs/config_search.md): Model Analyzer can
   help you automatically find the optimal settings for
   [Max Batch Size](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_configuration.md#maximum-batch-size),
@@ -44,6 +42,9 @@ Triton Inference Server.
   [Performance Analyzer](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/perf_analyzer.md)
   to test the model with different concurrency and batch sizes of requests. Using
   [Manual Config Search](docs/config_search.md#manual-brute-search), you can create manual sweeps for every parameter that can be specified in the model configuration.
+
+- [Multi-Model Search](docs/config_search.md#multi-model-search-mode): **EARLY ACCESS** - Model Analyzer can help you
+  find the optimal settings when profiling multiple concurrent models, utilizing our Quick Search alogrithm
 
 - [Detailed and summary reports](docs/report.md): Model Analyzer is able to generate
   summarized and detailed reports that can help you better understand the trade-offs
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Triton Inference Server.
 
 ## Features
 
-- [Multi-Model Search](docs/config_search.md): _Early Access: this feature is limited in scope_ - Model Analyzer can help you
+- [Multi-Model Search](docs/config_search.md#multi-model-search-mode): **EARLY ACCESS** - Model Analyzer can help you
   find the optimal settings when profiling multiple concurrent models, utilizing our Quick Search alogrithm
 - [Brute and Quick search](docs/config_search.md): Model Analyzer can
   help you automatically find the optimal settings for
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -33,37 +33,39 @@ Triton Inference Server.
...
-
-* [Detailed and summary reports](docs/report.md): Model Analyzer is able to generate
-summarized and detailed reports that can help you better understand the trade-offs
-between different model configurations that can be used for your model.
-
-* [QoS Constraints](docs/config.md#constraint): Constraints can help you
-filter out the Model Analyzer results based on your QoS requirements. For
-example, you can specify a latency budget to filter out model configurations
-that do not satisfy the specified latency threshold.
+- [Multi-Model Search](docs/config_search.md#multi-model-search-mode): **EARLY ACCESS** - Model Analyzer can help you
```
tgerdesnv, line 36: Do we want this to be at the top?  Maybe after `manual config search`?
Also, not sure about the usage of "our". Probably "utilizing the Quick search mode"

docs/config_search.md
```diff
@@ -174,3 +179,21 @@ the maximal objective value within the specified constraints. In the majority of
...
 
 After it has found the best config(s), it will then sweep the top-N configurations found (specified by `--num-configs-per-model`) over the default concurrency range before generation of the summary reports.
+
+## Multi-Model Search Mode
+
+_This mode is in EARLY ACCESS and has the following limitations:_
+
+- Can only be run in `quick` search mode
+- Cannot set limitations on min/max batch size, concurrency or instance count
+- Does not support individual model constraints, only global
```
tgerdesnv, line 189: only global constraints.

docs/config_search.md
```diff
@@ -174,3 +179,21 @@ the maximal objective value within the specified constraints. In the majority of
...
+- Does not support individual model weighting, all models are treated with equal priority when trying to maximize objective value
+- Does not support detailed reporting, only summary reports
+
+Multi-model concurrent search mode can be enabled by adding the parameter `--run-config-profile-models-concurrently-enable` to the CLI.
+
+It uses Quick Search mode's hill climbing algorithm to search all models configurations spaces in parallel, looking for the maximal objective value within the specified constraints. In limited testing, we have observed positive outcomes towards finding the maximum objective value; with runtimes of less than one hour (compared to the days it would take a brute force run to complete).
```
tgerdesnv, line 195: I'm not sure we want to mention limited testing. I would remove the whole last sentence.

docs/config_search.md
```diff
@@ -174,3 +179,21 @@ the maximal objective value within the specified constraints. In the majority of
 this will find greater than 95% of the maximum objective value (that could be found using a brute force search), while needing to search less than 10% of the configuration space.
 
 After it has found the best config(s), it will then sweep the top-N configurations found (specified by `--num-configs-per-model`) over the default concurrency range before generation of the summary reports.
+
+## Multi-Model Search Mode
```
tgerdesnv, line 183: Can we find somewhere to put in an example run command and/or quick start?  Copy an add_sub to add_sub2 and run? Or create a new simple example model.

docs/config.md
```diff
@@ -132,16 +132,16 @@ profile_models: <comma-delimited-string-list>
 # Triton Docker image tag used when launching using Docker mode
 [ triton_docker_image: <string> | default: nvcr.io/nvidia/tritonserver:22.09-py3 ]
 
-# Triton Server HTTP endpoint url used by Model Analyzer client. Will be ignored if server-launch-mode is not 'remote'".
+# Triton Server HTTP endpoint url used by Model Analyzer client.".
```
tgerdesnv, line 135: You're missing 3 more 'ignore' cases from config_command_profile, which is what feeds into the incorrect `--help` output

tgerdesnv (changes requested):

docs/config_search.md
tgerdesnv: Also, what are your thoughts on creating a new multi-model page instead of adding to config_search.md? Do you think it is similar enough to keep it there? Or should we give the new feature its own page (and a clean place to put examples)?

README.md
nv-braf: Moved as suggested

docs/config_search.md
nv-braf: Removed "In limited testing"

docs/config.md
nv-braf: Fixed

docs/config_search.md
tgerdesnv: I'm still not sure about saying "we" here. I don't think we want to personify it.
Also, can you say "well under an hour" or "run times around 15-30 minutes"?

tgerdesnv (approved):

PR #543, opened 2022-10-07
Fix  'custom_ops'  test configuration

mc-nv:


debermudez (commented):
seems fine to me. I will wait for iman to take a look too

Tabrizian (approved):

PR #547, opened 2022-10-12
Add support for model weighting

nv-braf:
- Added 'weighting' field to the YAML. By default the model weighting is set to 1, and can be changed per model by the user.
- Unit testing for both YAML and RunConfigResultComaparator
- Ran quick search with weightings and confirmed that they are being used (no changes needed)


model_analyzer/config/input/config_command_analyze.py
```diff
@@ -414,13 +414,21 @@ def _autofill_values(self):
...
                 new_model['objectives'] = model.objectives()
+
             # Constraints
             if not model.constraints():
                 if 'constraints' in self._fields and self.constraints:
                     new_model['constraints'] = self.constraints
             else:
                 new_model['constraints'] = model.constraints()
 
+            # Weightings
```
tgerdesnv, line 425: do we still need to update both ConfigAnalyze and ConfigProfile? Also, why is the newly added code different in each?

model_analyzer/config/input/objects/config_model_analysis_spec.py
```diff
@@ -80,11 +97,21 @@ def set_constraints(self, constraints):
         Parameters
         -------
         constraints : dict or None
-            A dictionary containing the parameters
+            A dictionary containing the constraints
         """
 
         self._constraints = constraints
 
+    def set_weightings(self, weightings):
```
tgerdesnv, line 105: Do we need this function? I don't see it called, and I don't think we have similar functions for constraints or objectives?

model_analyzer/config/input/objects/config_model_analysis_spec.py
```diff
@@ -55,6 +62,16 @@ def constraints(self):
 
         return self._constraints
 
+    def weightings(self):
```
tgerdesnv, line 65: I can't tell where weights are a single value, and where they are a list. Type checking would help, as would proper pluralization. The name implies it returns multiple weights, but I think this is just a scalar int?

Also, can we populate this with default values so that we don't have to handle "or none" all over the place?

model_analyzer/result/run_config_measurement.py
```diff
@@ -86,7 +86,7 @@ def from_dict(cls,
             run_config_measurement,
             run_config_measurement_dict['_model_config_measurements'])
 
-        # By default setting all models to have equal weighting
+        # By default set all models to have equal weighting
```
tgerdesnv, line 89: Are we going to write out and read in the model weights? Should we?

tests/test_config.py
```diff
@@ -1868,6 +1868,46 @@ def test_quick_search_model_specific(self):
...
+        Test that weightings can be specified only globally in the YAML
+        """
+        args = [
+            'model-analzyer', 'profile', '--model-repository', 'cli-repository',
+            '--run-config-search-mode', 'quick', '-f', 'path-to-config-file'
+        ]
+
+        # Global weightings - legal
+        yaml_content = """
+        weightings: [3, 1]
```
tgerdesnv, line 1882: This does not seem very use friendly to me. The name is very generic and doesn't specify what kind of weights they are (model weights). Also, the user has to know that the order of the values corresponds to the order of the keys in the below profile_models.
Is there any reason we can't just specify a model weight (a single integer value) inside of the model like we do for objectives and constraints, with a default of 1?

tgerdesnv (changes requested):

model_analyzer/result/run_config_measurement.py
nv-braf: We are not, and we should not. Weights, like constraints and objectives are not stored in the checkpoint and must be specified by the user each time MA is run (else the defaults are used).

model_analyzer/config/input/objects/config_model_analysis_spec.py
```diff
@@ -19,7 +19,11 @@ class ConfigModelAnalysisSpec:
     a single model.
     """
 
-    def __init__(self, model_name, objectives=None, constraints=None):
+    def __init__(self,
```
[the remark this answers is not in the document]
nv-braf: We have not combined the two yet. The code is needed in both places.

model_analyzer/config/input/objects/config_model_analysis_spec.py
nv-braf: It is not called, but we do have set functions for objectives and constraints, so I thought it was good to add this. I thought it was good/clean for a class to have get/set methods for private variables, regardless of whether they are currently used.

model_analyzer/config/input/config_command_analyze.py
nv-braf: We do. And, this is just one(more) reason why we should get rid of analyze. This code should be identical to profile. Good catch.

model_analyzer/config/input/objects/config_model_analysis_spec.py
nv-braf: Yes, this should not be plural. I went to add type checking and then got a ton of complaints because not all the variables in the methods were type checked. I tried to boyscout the change, but it was too onerous.

tests/test_config.py
nv-braf: I've changed `weightings` to `weighting` and switched to now flag this if it is set globally. We now take an integer (instead of a list) and the default value is 1 if not specified for a model.

model_analyzer/config/input/objects/config_model_analysis_spec.py
tgerdesnv: I definitely wouldn't call it clean to always have get/set. If the values should be set on construction and never modified then we don't want a set function.

I see now that there are set functions for constraints/objectives/model name. They are never called and should be ripped out. Those aren't things we ever expect to change.

model_analyzer/config/input/objects/config_model_analysis_spec.py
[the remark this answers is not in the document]
tgerdesnv: ok let's prioritize ripping out analysis. Are there any differences in the classes here? Can you create a story?

model_analyzer/result/result_manager.py
```diff
@@ -336,9 +340,13 @@ def _setup_for_sequential_analysis(self):
             model.model_name() for model in self._config.analysis_models
         ]
 
+        # If analyzing sequentially just set all models to equal weighting (since it won't be used)
```
tgerdesnv, line 343: Any reason this code can't be the same as the concurrent code now that we have defaults?  Even if they specify some other value like 2, it won't change anything.

model_analyzer/result/result_manager.py
```diff
@@ -320,11 +320,15 @@ def _setup_for_concurrent_analysis(self):
...
+        model_weighting_list = [
+            model.weighting() for model in self._config.analysis_models
+        ]
 
         self._run_comparators = {
             self._concurrent_analysis_model_name:
                 RunConfigResultComparator(
-                    metric_objectives_list=model_objectives_list)
+                    metric_objectives_list=model_objectives_list,
+                    model_weights=model_weighting_list)
```
tgerdesnv, line 331: remind me why both the measurement (the thing being compared) and the comparator (the thing doing the comparison) both need the objectives and model weights?  Something about comparing between models in the same measurement vs comparing between measurements?

model_analyzer/result/run_config_measurement.py
```diff
@@ -86,9 +86,9 @@ def from_dict(cls,
             run_config_measurement,
             run_config_measurement_dict['_model_config_measurements'])
 
-        # By default setting all models to have equal weighting
-        for mcm in run_config_measurement._model_config_measurements:
-            run_config_measurement._model_config_weights.append(1)
+        # By default set all models to have equal weighting
+        # for mcm in run_config_measurement._model_config_measurements:
```
tgerdesnv, line 90: remove?

model_analyzer/result/run_config_measurement.py
```diff
@@ -104,10 +104,11 @@ def set_model_config_weighting(self,
         Weights are the relative importance of the model_configs 
         with respect to one another
         """
-        self._model_config_weights = [
-            model_config_weight / sum(model_config_weights)
-            for model_config_weight in model_config_weights
-        ]
+        if not None in model_config_weights:
```
tgerdesnv, line 107: is None possible here?

model_analyzer/result/run_config_measurement.py
```diff
@@ -548,7 +549,8 @@ def _calculate_weighted_rcm_score(
             The weighted score. A positive value indicates this 
             RunConfig measurement is better than the other
         """
-        assert len(self._model_config_weights) == len(weighted_mcm_scores)
+        if (len(self._model_config_weights) != len(weighted_mcm_scores)):
```
tgerdesnv, line 552: um... why this change? It looks redundant. Was this for debugging?

tests/test_config.py
```diff
@@ -1868,6 +1868,46 @@ def test_quick_search_model_specific(self):
         with self.assertRaises(TritonModelAnalyzerException):
             self._evaluate_config(args, yaml_content, subcommand='profile')
 
+    def test_model_weightings(self):
+        """
+        Test that weightings can be specified only per model in the YAML
+        """
+        args = [
+            'model-analzyer', 'profile', '--model-repository', 'cli-repository',
```
tgerdesnv, line 1876: your test uses profile. Are we confident it works if the user specifies the analyze step instead?

tests/test_run_config_result_comparator.py
```diff
@@ -126,8 +139,50 @@ def test_equal_weight_multi(self):
             model_name="test_model",
             model_config_names=["test_model_config0", "test_model_config1"])
 
+    def test_unequal_weight_multi(self):
```
tgerdesnv, line 142: I see a test that confirms the input format (yaml only).
I see a test that confirms the weights work properly.
I don't see anything that confirms that we go properly from yaml->comparison objects->correct comparison

tests/test_config.py
```diff
@@ -1868,6 +1868,46 @@ def test_quick_search_model_specific(self):
...
+        """
+
+        with self.assertRaises(TritonModelAnalyzerException):
+            self._evaluate_config(args, yaml_content, subcommand='profile')
+
+        # Model specific weighting - legal
+        yaml_content = """
+        profile_models:
+          resnet50_libtorch:
+            weighting: 3
```
tgerdesnv, line 1900: This looks good and makes a lot more sense now.  What is missing is documentation. Can you add a subsection to the multi-model page about weighting with a simple example, and also document the weighting option (maybe in config.md? I'm not sure where to put something that can only be specified per-model)

tgerdesnv (changes requested):

model_analyzer/config/input/objects/config_model_analysis_spec.py
nv-braf: Removed

model_analyzer/config/input/objects/config_model_analysis_spec.py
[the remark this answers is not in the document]
nv-braf: Yes, analysis is just a small subset of profile now. We can remove this. I've filed TMA-944.

model_analyzer/result/result_manager.py
nv-braf: That's true. I've removed this code.

model_analyzer/result/result_manager.py
nv-braf: RunConfigResultComarator compares groups of RunConfigMeasurements (which are called RunConfigResults)

model_analyzer/result/run_config_measurement.py
nv-braf: Removed

model_analyzer/result/run_config_measurement.py
nv-braf: No, not anymore. Removed

tests/test_config.py
nv-braf: No. I've added unit tests to protect this.

tests/test_config.py
nv-braf: Yes, documentation was going to be my next step. Just wasn't sure if it should be part of this story or a follow on. I can make it part of this story if that makes the most sense.

nv-braf: Please review the added documenation.

docs/config.md
```diff
@@ -244,6 +244,9 @@ profile_models: <comma-delimited-string-list|list|profile_model>
 # List of objectives that user wants to sort the results by it.
 [ objectives: <objective|list> ]
 
+# Weighting used to bias the model's objectives (agianst the other models) in multi-mode mode
```
tgerdesnv, line 247: My only comment is that we need to somehow communicate (here and/or in your other section) that this is not a global option, and must be specified down at the model level inside of profile_models

docs/config.md
```diff
@@ -574,6 +571,12 @@ These parameters will result in testing the concurrency configurations of 2, 10,
 18, 26, 34, 42, 50, 58, and 64, for each of different batch sizes of 1, 2 and 3.
 This will result in 27 individual test runs of the model.
 
+### `<weighting>`
+
+This field is used to bias a model's objective when performing a multi-model search.
+
+See [Multi-Model Search - Model Weighting](docs/config_search.md#model-weighting) for details and an example YAML configuration.
```
tgerdesnv, line 578: dead link (remove docs?)

tgerdesnv (changes requested):

docs/config.md
nv-braf: Do we? There's no example that uses it globally. It's completely non-sensical to have a single number in the global option. And, if a user does try it that get a very nice and descriptive error message telling them that it can't be global.

docs/config.md
nv-braf: Fixed

tgerdesnv (approved):

PR #550, opened 2022-10-18
Removal of analysis subcommand. Passing unit testing

nv-braf:
Completely removes the analysis subcommand from the codebase, L0 testing and documentation.


docs/cli.md
```diff
@@ -194,73 +194,6 @@ analysis_models:
...
-the checkpoint directory. Next, it sorts the models specified in the CLI or
-config YAML, provided they contain measurements in the checkpoint, using the
-objectives specified in the config YAML. Finally, it constructs summary PDFs
-using the top model configs for each model, as well as across models, if
-requested (See the [Reports](./report.md) section for more details). The
-`analyze` subcommand can be run multiple times with different configurations if
-the user would like to sort and filter the results using different objectives or
-under different constraints.
-
-### Examples
```
tgerdesnv, line 223: Are there examples still worth keeping, but showing it in the context of the profile step?

model_analyzer/entrypoint.py
```diff
@@ -161,11 +159,9 @@ def get_cli_and_config_options():
             help=
             'Run model inference profiling based on specified CLI or config options.',
             config=config_profile)
-        cli.add_subcommand(
-            cmd='analyze',
-            help=
-            'Collect and sort profiling results and generate data and summaries.',
-            config=config_analyze)
+        cli.add_subcommand(cmd='analyze',
+                           help='DEPRECATED: Please use profile subcommand.',
```
tgerdesnv, line 163: mention that it is an alias for profile

qa/L0_analyze/test.sh
```diff
@@ -1,103 +0,0 @@
-# Copyright (c) 2021 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tgerdesnv, line 1: Is there a matching MR to remove this test from CI?

tgerdesnv (changes requested):

model_analyzer/entrypoint.py
nv-braf: Done

qa/L0_analyze/test.sh
nv-braf: Yes, I've created this

docs/cli.md
nv-braf: They already exist for the profile step and are redundant (see line 166)

tgerdesnv (approved):

PR #552, opened 2022-10-19
Use NVML to Acquire GPU UUIDs

nv-braf:
Replaces DCGM call with ones from NVML

### DIFF
--- a/model_analyzer/device/gpu_device_factory.py
+++ b/model_analyzer/device/gpu_device_factory.py
@@ -243,8 +243,9 @@ class GPUDeviceFactory:
                         pci_bus_id=
                         '',  # FIXME: TMA-958: PCI bus ID isn't needed and will be removed when DCGM is removed
                         device_uuid=uuid))
-        except NVMLError as error:
-            raise TritonModelAnalyzerException(f"NVML error: {error}")
+        except NVMLError:
+            # This can happen if running on a machine without a GPU (CPU_ONLY)
+            return []
 
         return gpu_devices
 
--- a/model_analyzer/device/gpu_device_factory.py
+++ b/model_analyzer/device/gpu_device_factory.py
@@ -19,6 +19,7 @@ import model_analyzer.monitor.dcgm.dcgm_structs as structs
 from model_analyzer.model_analyzer_exceptions import TritonModelAnalyzerException
 
 import numba.cuda
+import subprocess
 import logging
 
 logger = logging.getLogger(LOGGER_NAME)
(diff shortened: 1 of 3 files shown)


model_analyzer/device/gpu_device_factory.py
```diff
@@ -223,19 +226,28 @@ def get_cuda_visible_gpus(self):
...
-                        f"Device '{str(cuda_device.name, encoding='ascii')}' with "
-                        f"cuda device id {cuda_device.id} is not supported by DCGM."
-                    )
-        return cuda_visible_gpus
+        nvmlInit()
+
+        gpu_devices = []
+        try:
+            devices = nvmlDeviceGetCount()
+            for device_id in range(devices):
```
rmccorm4, line 234: I noticed the previous variable name was `cuda_visible_gpus`, so just wanted to mention this:
 
Might not be an issue for anyone, but `nvml` doesn't respect `CUDA_VISIBLE_DEVICES`. So if you have any tests that exploit this env var for isolation inside the container, it likely won't work.

Isolating gpus in a container through `docker run --gpus '"device=1,3,5"' ...` will work as expected though.

model_analyzer/device/gpu_device_factory.py
```diff
@@ -18,7 +18,10 @@
 import model_analyzer.monitor.dcgm.dcgm_structs as structs
 from model_analyzer.model_analyzer_exceptions import TritonModelAnalyzerException
 
+from pynvml import *
+
 import numba.cuda
+import subprocess
```
tgerdesnv, line 24: are you using this?

model_analyzer/device/gpu_device_factory.py
tgerdesnv: the function name is get_cuda_visible_gpus, so we should honor that. Is there a way to do so? Can we read that env variable here to filter the gpus?

model_analyzer/device/gpu_device_factory.py
```diff
@@ -223,19 +226,28 @@ def get_cuda_visible_gpus(self):
             UUIDs of the DCGM supported devices visible to CUDA
         """
 
-        cuda_visible_gpus = []
-        if numba.cuda.is_available():
-            for cuda_device in numba.cuda.list_devices():
-                try:
-                    cuda_visible_gpus.append(
-                        self.get_device_by_cuda_index(cuda_device.id))
```
tgerdesnv, line 231: it seems that this function should be removed if it is no longer used?

model_analyzer/device/gpu_device_factory.py
```diff
@@ -18,7 +18,10 @@
 import model_analyzer.monitor.dcgm.dcgm_structs as structs
 from model_analyzer.model_analyzer_exceptions import TritonModelAnalyzerException
 
+from pynvml import *
+
 import numba.cuda
```
tgerdesnv, line 23: can this be removed?

tgerdesnv (changes requested):

model_analyzer/device/gpu_device_factory.py
nv-braf: Removed

model_analyzer/device/gpu_device_factory.py
nv-braf: The function is still being used.

model_analyzer/device/gpu_device_factory.py
nv-braf: No, numba.cuda is still being used

model_analyzer/device/gpu_device_factory.py
nv-braf: Ryan, thanks for looking at this! I've added in logic to check for CUDA_VISIBLE_DEVICES and map them to the device ids.

model_analyzer/device/gpu_device_factory.py
```diff
@@ -223,19 +228,47 @@ def get_cuda_visible_gpus(self):
...
+        nvmlInit()
+
+        gpu_devices = []
+        cuda_id_map = []
+        cuda_visiable_devices_set = "CUDA_VISIBLE_DEVICES" in os.environ
+        try:
+            if cuda_visiable_devices_set:
+                cuda_id_map = list(
+                    map(int,
+                        os.environ.get("CUDA_VISIBLE_DEVICES", "").split(",")))
```
rmccorm4, line 240: CUDA_VISIBLE_DEVICES also supports MIG IDs: https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html#cuda-gi which are strings and not just ints.

You may be able to just keep it as a list of strings and instead check something like this:
```
if id in cuda_visible_devices or uuid in cuda_visible_devices:
   ...
```

**disclaimer**: I have no idea what values nvml will return for MIG devices, so you may want to just make this a FIXME or something to support MIG devices as needed based on priority.

model_analyzer/device/gpu_device_factory.py
tgerdesnv: I'm confused. I thought the whole point of this was to remove DCGM, but it doesn't appear that dcgm was used in the function you removed? I do still see it used in init_all_devices() though. Did we change the wrong code?

PR #556, opened 2022-11-04
Fix PDF warning to include command

tgerdesnv:
Despite what https://github.com/triton-inference-server/model_analyzer/pull/493 says, wkhtmltopdf is NOT included in the wheel install. Thus, adding back in clearer directions for the case where the user runs without it installed.

Also, making sure the warning only prints once per run.


nv-braf (dismissed):
Looks good

nv-braf (approved):

PR #558, opened 2022-11-04
Creating partial overlap case using batch size

nv-braf:
Re-runs profile while extending batch size

### DIFF
--- a/qa/L0_profile/test.sh
+++ b/qa/L0_profile/test.sh
@@ -40,7 +40,7 @@ FILENAME_GPU_MODEL="model-metrics-gpu.csv"
 
 rm -rf $OUTPUT_MODEL_REPOSITORY
 
-python3 test_config_generator.py --profile-models $MODEL_NAMES
+python3 test_config_generator.py --profile-models $MODEL_NAMES -b $BATCH_SIZES -c $CONCURRENCY
 
 # Run the analyzer and check the results
 RET=0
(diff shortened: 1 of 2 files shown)


qa/L0_profile/test.sh
```diff
@@ -56,6 +56,40 @@ MODEL_ANALYZER_ARGS="$MODEL_ANALYZER_ARGS -e $EXPORT_PATH --filename-server-only
 MODEL_ANALYZER_ARGS="$MODEL_ANALYZER_ARGS --filename-model-inference=$FILENAME_INFERENCE_MODEL --filename-model-gpu=$FILENAME_GPU_MODEL"
 MODEL_ANALYZER_ARGS="$MODEL_ANALYZER_ARGS --skip-summary-reports"
 MODEL_ANALYZER_SUBCOMMAND="profile"
+run_analyzer
```
tgerdesnv, line 59: Is this just a cut and paste of the same code below? Are there any different checks that we could/should have on the rerun? Or are we just making sure that MA doesn't barf on the rerun?

nv-braf: We are checking that new measurements got added (and that it doesn't barf). It's a rerun with a change to batch sizes (from 1,2 -> 1,2,4). 

I remove the hardcoded values in test_config_generator, so that we will now check for the extra configurations.

tgerdesnv: Any confirmation that the old values remain and aren't remeasured?

nv-braf: No, I'm not sure how to easily figure that out in the L0 (maybe grepping the logs and looking for the `skipping measurements` clause, but that seems fragile to me)

tgerdesnv (approved):

PR #567, opened 2022-11-14
Update summary report message

nv-braf:
Updated result_summary.pdf statement to look like: 

With a gain: image

When default is the best: image

### DIFF
--- a/model_analyzer/record/record.py
+++ b/model_analyzer/record/record.py
@@ -233,3 +233,32 @@ class Record(metaclass=RecordType):
             return type(self)(value=(self.value() / other))
         else:
             raise TypeError
+
+    @abstractmethod
+    def _positive_is_better(self) -> bool:
+        """
+        Returns a bool indicating if a larger positive value is better
+        for a given record type
+        """
+
+    def calculate_percentage_gain(self, other: "Record") -> float:
+        """
+        Calculates percentage gain between records
+        """
+
+        if self._positive_is_better():
+            return ((self.value() - other.value()) / other.value()) * 100
+        else:
+            return ((other.value() - self.value()) / self.value()) * 100
+
+
+class IncreasingRecord(Record):
+
+    def _positive_is_better(self) -> bool:
+        return True
+
+
+class DecreasingRecord(Record):
+
+    def _positive_is_better(self) -> bool:
+        return False
--- a/model_analyzer/record/record.py
+++ b/model_analyzer/record/record.py
@@ -246,19 +246,32 @@ class Record(metaclass=RecordType):
         Calculates percentage gain between records
         """
 
+        # When increasing values are better gain is based on the orignal value (other):
+        # example: 200 vs. 100 is (200 - 100) / 100 = 100%
+        # example: 100 vs. 200 is (100 - 200) / 200 = -50%
         if self._positive_is_better():
             return ((self.value() - other.value()) / other.value()) * 100
+
+        # When decreasing values are better gain is based on the new value (self):
+        # example: 100 vs. 200 is (200 - 100) / 100 = 100%
+        # example: 200 vs. 100 is (100 - 200) / 200 = -50%
         else:
             return ((other.value() - self.value()) / self.value()) * 100
 
 
 class IncreasingRecord(Record):
+    """
+    Record where an increasing positive value is better
+    """
 
     def _positive_is_better(self) -> bool:
         return True
 
 
 class DecreasingRecord(Record):
+    """
+    Record where an increasing positive value is worse
+    """
 
     def _positive_is_better(self) -> bool:
         return False
--- a/model_analyzer/record/types/cpu_available_ram.py
+++ b/model_analyzer/record/types/cpu_available_ram.py
@@ -14,11 +14,11 @@
 
 from functools import total_ordering
 
-from model_analyzer.record.record import Record
+from model_analyzer.record.record import Record, IncreasingRecord
 
 
 @total_ordering
-class CPUAvailableRAM(Record):
+class CPUAvailableRAM(IncreasingRecord):
     """
     The Available CPU memory
     """
@@ -56,13 +56,6 @@ class CPUAvailableRAM(Record):
         """
 
         return ("Max " if aggregation_tag else "") + "RAM Available (MB)"
-
-    def calculate_percentage_gain(self, other: Record) -> float:
-        """
-        Calculates percentage gain between records
-        """
-
-        return ((self.value() - other.value()) / other.value()) * 100
 
     def __eq__(self, other):
         """
--- a/model_analyzer/record/types/cpu_available_ram.py
+++ b/model_analyzer/record/types/cpu_available_ram.py
@@ -57,6 +57,13 @@ class CPUAvailableRAM(Record):
 
         return ("Max " if aggregation_tag else "") + "RAM Available (MB)"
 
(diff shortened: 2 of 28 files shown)


model_analyzer/result/model_config_measurement.py
```diff
@@ -291,6 +291,43 @@ def _calculate_weighted_score(self, other):
...
+            this ModelConfig measurement is better than the other
+        """
+
+        weighted_pct = 0.0
+        for objective, weight in self._metric_weights.items():
+            self_metric = self.get_metric(tag=objective)
+            other_metric = other.get_metric(tag=objective)
+
+            # Handle the case where objective GPU metric is queried on CPU only
+            if self_metric and other_metric is None:
```
nv-braf, line 318: The percentage gain/loss (100) is arbitrary. Not sure what I should do in this case. Should I throw an error?

model_analyzer/record/types/cpu_available_ram.py
```diff
@@ -57,6 +57,13 @@ def header(aggregation_tag=False):
 
         return ("Max " if aggregation_tag else "") + "RAM Available (MB)"
 
+    def calculate_percentage_gain(self, other: Record) -> float:
```
tgerdesnv, line 60: It still feels like we should be able to define this in Record for the common case, and only overwrite it for cases where lower is better. We shouldn't have to shotgun 25 files. I understand that for add and sub they had to be manually specified because it had to return a new object of the specific type (which I still think could be solved in a better way), but this is just returning a float and is only utilizing value() function which is defined in Record

model_analyzer/reports/report_manager.py
```diff
@@ -472,42 +472,38 @@ def _create_summary_config_phrase(self, best_run_config,
...
-        throughput_gain = round(
-            (best_throughput - default_throughput) / default_throughput *
-            100) if default_throughput else None
-
-        throughput_phrase = "total " if multi_model else ""
-        throughput_phrase = (
-            throughput_phrase +
-            f"throughput: <strong>{best_throughput} infer/sec</strong>.<br><br>"
-        )
+    def _create_summary_objective_phrase(self, report_key,
```
tgerdesnv, line 475: can we type check this?

model_analyzer/result/model_config_measurement.py
tgerdesnv: If both are none, it seems like there is no gain, so 0% would be fine?

tgerdesnv (changes requested):

model_analyzer/result/model_config_measurement.py
nv-braf: Yes, I'm already setting that case to 0%, just wondering what I should be doing for the other two cases (where only one is None)

model_analyzer/reports/report_manager.py
nv-braf: We can! (and it should have been). Type checking added.

model_analyzer/record/types/cpu_available_ram.py
nv-braf: It's a near 50/50 split (11 of 19 need to be inverted). Seems like it would be more work to rip it out at this point and I'm not sure what it buys us. I think it's more readable to have the function in the individual classes and we're less bug prone when adding future records.

tgerdesnv: Any time we are cutting and pasting something over and over, we should find a better way. If we decide to change how this is calculated, now we have to update 25 places and are prone to error. Also, 25 instances of code technically needs to be tested 25 times.

To be clear, this was the basic design I was suggesting:

```
Record:
  _positive_is_better():
    undefined virtual function

  calculate_percentage():
    if _positive_is_better():
      <handle that case>
    else:
      <handle that case>

class IncreasingRecord(Record):
  _positive_is_better(): Return True

class DecreasingRecord(Record):
  _positive_is_better(): Return False

<All records extend either IncreasingRecord or DecreasingRecord. No need to cut and paste over and over>
```

nv-braf: I see what you're saying. I've added this and unit testing to protect it. Please re-review

model_analyzer/record/record.py
```diff
@@ -233,3 +233,32 @@ def __truediv__(self, other):
...
+        Returns a bool indicating if a larger positive value is better
+        for a given record type
+        """
+
+    def calculate_percentage_gain(self, other: "Record") -> float:
+        """
+        Calculates percentage gain between records
+        """
+
+        if self._positive_is_better():
```
tgerdesnv, line 249: Put a comment somewhere in here about why the calculation is different for the two cases.

model_analyzer/record/record.py
```diff
@@ -233,3 +233,32 @@ def __truediv__(self, other):
...
+            return ((self.value() - other.value()) / other.value()) * 100
+        else:
+            return ((other.value() - self.value()) / self.value()) * 100
+
+
+class IncreasingRecord(Record):
```
tgerdesnv, line 255: docstring these classes
"Records where a larger value is 'better'"

model_analyzer/record/types/gpu_free_memory.py
```diff
@@ -13,11 +13,11 @@
 # limitations under the License.
 
 from functools import total_ordering
-from model_analyzer.record.gpu_record import GPURecord
+from model_analyzer.record.gpu_record import GPURecord, IncreasingGPURecord
 
 
 @total_ordering
-class GPUFreeMemory(GPURecord):
```
tgerdesnv, line 20: What was implemented in GPURecord? Are we losing that?

model_analyzer/record/types/gpu_power_usage.py
```diff
@@ -85,7 +85,7 @@ def __lt__(self, other):
         the other
         """
 
-        return self.value() < other.value()
+        return other.value() < self.value()
```
tgerdesnv, line 88: It feels like these should all be in the root class so there is a single point of failure. I won't make you do that this story, but it is something that we should consider in the future

model_analyzer/result/model_config_measurement.py
tgerdesnv: Oh, yeah I misread the code. Is that even possible? Wouldn't the case in the comment hit the None+None case?

I'm fine with 100/-100

tgerdesnv (changes requested):

model_analyzer/record/types/gpu_power_usage.py
nv-braf: I thought about that, but it seemed beyond the scope of this story

model_analyzer/record/types/gpu_free_memory.py
nv-braf: We're not losing any functionality. I've just augmented GPURecord to have increasing and decreasing records.

model_analyzer/record/record.py
nv-braf: Done

model_analyzer/record/record.py
nv-braf: Done

model_analyzer/record/types/gpu_free_memory.py
tgerdesnv: ok, I see now that it is IncreasingGPURecord 👍

tgerdesnv (approved):

PR #568, opened 2022-11-16
Fixing constraints statement for multi-model

nv-braf:
Noticed a bug in the constraints message in our result summary for multi-model runs:
image

Root caused an fixed it:
image

I've also added unit testing to ensure this doesn't break in the future and checked that single model is still printing as intended (just the constraint w/o the test name)

### DIFF
--- a/model_analyzer/reports/report_manager.py
+++ b/model_analyzer/reports/report_manager.py
@@ -439,7 +439,7 @@ class ReportManager:
 
     def _create_multi_model_constraint_string(self, report_key: str,
                                               constraint_strs: str) -> str:
-        constraint_str = None
+        constraint_str = ""
         for model_name in report_key.split(','):
             if model_name in constraint_strs:
                 if constraint_str:
@@ -447,11 +447,8 @@ class ReportManager:
                     for i in range(len("Constraint targets: ")):
                         constraint_str += "&ensp;"
 
-                    constraint_str += "<strong>" + model_name + "</strong>: " + constraint_strs[
-                        model_name]
-                else:
-                    constraint_str = "<strong>" + model_name + "</strong>: " + constraint_strs[
-                        model_name]
+                constraint_str += "<strong>" + model_name + "</strong>: " + constraint_strs[
+                    model_name]
 
         return constraint_str
 
--- a/model_analyzer/reports/report_manager.py
+++ b/model_analyzer/reports/report_manager.py
@@ -437,8 +437,8 @@ class ReportManager:
 
         return constraint_str
 
-    def _create_multi_model_constraint_string(self, report_key: str,
-                                              constraint_strs: str) -> str:
+    def _create_multi_model_constraint_string(
+            self, report_key: str, constraint_strs: Dict[str, str]) -> str:
         constraint_str = ""
         for model_name in report_key.split(','):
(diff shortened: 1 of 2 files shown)


model_analyzer/reports/report_manager.py
```diff
@@ -426,6 +420,28 @@ def _find_best_configs(self, report_key):
...
+            if report_key == TOP_MODELS_REPORT_KEY:
+                constraint_str = constraint_strs['default']
+            elif ',' in report_key:  # indicates multi-model
+                constraint_str = ""
+                for model_name in report_key.split(','):
+                    if model_name in constraint_strs:
+                        constraint_str += "<strong>" + model_name + "</strong>: " + constraint_strs[
+                            model_name] + "<br>&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;"
+
+                constraint_str = constraint_str[:-64]  # remove trailing <br>...
```
tgerdesnv, line 436: what kind of wizardry is this

tests/test_report_manager.py
```diff
@@ -538,6 +538,40 @@ def test_gpu_name_and_memory(self):
...
+            "modelA": "Max p99 latency: 100 ms",
+            "modelB": "Max p99 latency: 200 ms"
+        })
+    def test_constraint_string_multi_model(self, *args):
+        report_manager = ReportManager(mode=MagicMock(),
+                                       config=MagicMock(),
+                                       gpu_info=MagicMock(),
+                                       result_manager=MagicMock())
+        expected_constraint_str = "<strong>modelA</strong>: Max p99 latency: 100 ms"
+        expected_constraint_str += "<br>&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;"
```
tgerdesnv, line 567: What is emsp?

model_analyzer/reports/report_manager.py
matthewkotila: image

tests/test_report_manager.py
nv-braf: It's how you create spaces in the html/pdf file after the newline `<br>`. I believe each `&emsp;` is 4 spaces.

model_analyzer/reports/report_manager.py
nv-braf: Need to remove the trailing newline and spaces from the string. I couldn't think of an easier way to do this, without re-writing the loop to be aware of the terminal model.

tgerdesnv: So you're adding endline and spaces to line up the potential next entry, and you don't know which will be the last entry, so you then delete the last one? 

Why not add the endline/spaces before instead of after, and skip the first one? Similar to how you would print a list of values with commas in between. You wouldn't add a comma after every one and then delete the last one. You would add a comma before every one EXCEPT the first one.

nv-braf: I've refactored and removed the wizardry. I'm now just a muggle. Please re-review.

tgerdesnv (dismissed):

tgerdesnv (approved):

PR #573, opened 2022-12-05
Add PA flag support to quick search's default config

nv-braf:
Fix issue where quick search's default config generation wasn't respecting the PA flags set by the user


tgerdesnv (commented):
Is there a unit test that can be added to protect this?

tgerdesnv (approved):

PR #574, opened 2022-12-05
Add ensemble support to ModelConfig

nv-braf:
Adds new methods to check if the model's platform is ensemble and to change the submodel variant names.

Note: This is merging on to the ensemble branch, not main

### DIFF
--- a/model_analyzer/triton/model/model_config.py
+++ b/model_analyzer/triton/model/model_config.py
@@ -285,7 +285,7 @@ class ModelConfig:
 
         if not self.is_ensemble():
             raise TritonModelAnalyzerException(
-                "Cannot find submodels. Model platfrom is not ensemble.")
+                "Cannot find submodels. Model platform is not ensemble.")
 
         try:
             submodels = [
@@ -294,7 +294,7 @@ class ModelConfig:
             ]
         except:
             raise TritonModelAnalyzerException(
-                "Cannot find submodels. Ensemble Scheduling and/or step is not present in config protobuf"
+                "Cannot find submodels. Ensemble Scheduling and/or step is not present in config protobuf."
             )
 
         return submodels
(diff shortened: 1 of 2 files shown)


model_analyzer/triton/model/model_config.py
```diff
@@ -258,14 +258,56 @@ def set_cpu_only(self, cpu_only):
...
+        -------
+            List[str]: Sub-model names
+        """
+
+        if not self.is_ensemble():
+            return None
+
+        submodels = [
+            model['modelName']
+            for model in self.to_dict()['ensembleScheduling']['step']
```
tgerdesnv, line 291: Should we check the subkeys before indexing in? Seems to assume everything is in a correct state

model_analyzer/triton/model/model_config.py
```diff
@@ -258,14 +258,56 @@ def set_cpu_only(self, cpu_only):
...
+        return submodels
+
+    def set_submodel_variant_name(self, submodel_name: str,
+                                  variant_name: str) -> None:
+        """
+        Replaces the Ensembles submodel's name with the variant name
+        """
+
+        if not self.is_ensemble():
+            return
```
tgerdesnv, line 302: Can we error out here? that would mean bad code and I don't want to silently do that

model_analyzer/triton/model/model_config.py
```diff
@@ -258,14 +258,56 @@ def set_cpu_only(self, cpu_only):
...
+
+        model_config_dict = self.to_dict()
+        for submodel in model_config_dict['ensembleScheduling']['step']:
+            if submodel['modelName'] == submodel_name:
+                submodel['modelName'] = variant_name
+
+        self._model_config = self.from_dict(model_config_dict)._model_config
```
tgerdesnv, line 309: I don't understand this line. Is the only way to update this object to dump it to a dict, modify it, and then import it back from dict?

model_analyzer/triton/model/model_config.py
nv-braf: Changed this to raise an TMA exception.

tgerdesnv: Add a test to protect that and I think this is ready to be merged. (create an ensemble config that is missing 'step', and expect assert to be raised)

nv-braf:
Added tests to protect missing protobuf sections. Please review

PR #575, opened 2022-12-12
Ensemble default config generation

nv-braf:
Logic to create the default config in the Quick RCG for an ensemble model.
This also includes changes to the Model Variant Name Manager.

I have a new unit test in the QRCG and have run this with an ensemble model and verified that we are generating the default configs correctly.

### DIFF
--- a/model_analyzer/config/generate/base_model_config_generator.py
+++ b/model_analyzer/config/generate/base_model_config_generator.py
@@ -208,16 +208,27 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
 
     @staticmethod
     def make_ensemble_model_config(
-            model: ModelProfileSpec, ensemble_submodels: List[ModelConfig],
+            model: ModelProfileSpec,
+            ensemble_submodel_configs: List[ModelConfig],
             model_variant_name_manager: ModelVariantNameManager) -> ModelConfig:
         """
-        Loads the base model config from the model repository, and then mutates
+        Loads the ensemble model spec from the model repository, and then mutates
         the names to match the ensemble submodels
+        
+        Parameters
+        ----------
+        model: ModelProfileSpec
+            The top-level ensemble model spec
+        ensemble_submodel_configs: List of ModelConfigs
+            The list of submodel ModelConfigs 
+        model_variant_name_manager: ModelVariantNameManager
+
         """
         model_name = model.model_name()
         model_config_dict = model.get_default_config()
         ensemble_config_dicts = [
-            submodel.to_dict() for submodel in ensemble_submodels
+            submodel_config.to_dict()
+            for submodel_config in ensemble_submodel_configs
         ]
         ensemble_key = ModelVariantNameManager.make_ensemble_submodel_key(
             ensemble_config_dicts)
@@ -229,8 +240,8 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
         model_config_dict['name'] = variant_name
         model_config = ModelConfig.create_from_dictionary(model_config_dict)
 
-        for submodel in ensemble_submodels:
-            variant_name = submodel.get_field("name")
+        for submodel_config in ensemble_submodel_configs:
+            variant_name = submodel_config.get_field("name")
             submodel_name = BaseModelConfigGenerator.extract_model_name_from_variant_name(
                 variant_name)
 
@@ -251,7 +262,11 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
             self._max_batch_size_warning_printed = True
 
     @staticmethod
-    def extract_model_name_from_variant_name(variant_name):
+    def extract_model_name_from_variant_name(variant_name: str) -> str:
+        """
+        Removes '_config_#/default' from the variant name and returns
+        the model name, eg. model_name_config_10 -> model_name
+        """
         return variant_name[:variant_name.find("_config_")]
 
     @staticmethod
--- a/model_analyzer/config/generate/base_model_config_generator.py
+++ b/model_analyzer/config/generate/base_model_config_generator.py
@@ -210,7 +210,7 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
         model_name = model.model_name()
         model_config_dict = model.get_default_config()
         ensemble_config_dicts = [
-            model.to_dict() for model in ensemble_submodels
+            submodel.to_dict() for submodel in ensemble_submodels
         ]
         ensemble_key = ModelVariantNameManager.make_ensemble_submodel_key(
             ensemble_config_dicts)
@@ -224,7 +224,8 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
 
         for submodel in ensemble_submodels:
             variant_name = submodel.get_field("name")
-            submodel_name = variant_name[:variant_name.find("_config_")]
+            submodel_name = BaseModelConfigGenerator.extract_model_name_from_variant_name(
+                variant_name)
 
             model_config.set_submodel_variant_name(submodel_name=submodel_name,
                                                    variant_name=variant_name)
@@ -241,6 +242,10 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
                 "No longer increasing max_batch_size because throughput has plateaued"
             )
             self._max_batch_size_warning_printed = True
+
+    @staticmethod
+    def extract_model_name_from_variant_name(variant_name):
+        return variant_name[:variant_name.find("_config_")]
 
     @staticmethod
     def _apply_value_to_dict(key: Any, value: Any, dict_in: Dict) -> None:
--- a/model_analyzer/config/generate/base_model_config_generator.py
+++ b/model_analyzer/config/generate/base_model_config_generator.py
@@ -162,6 +162,13 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
         """
         Loads the base model config from the model repository, and then applies the
         parameters in the param_combo on top to create and return a new model config
+        
+        Parameters:
+        -----------
+        param_combo: dict
+            dict of key:value pairs to apply to the model config
+        model: ModelProfileSpec
+        model_variant_name_manager: ModelVariantNameManager
         """
         model_name = model.model_name()
 
--- a/model_analyzer/config/generate/base_model_config_generator.py
+++ b/model_analyzer/config/generate/base_model_config_generator.py
@@ -162,19 +162,6 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
         """
         Loads the base model config from the model repository, and then applies the
         parameters in the param_combo on top to create and return a new model config
-
-        Parameters:
-        -----------
-        param_combo: dict
-            dict of key:value pairs to apply to the model config
-        config: ModelAnalyzerConfig
-        client: TritonClient
-        gpus: List of GPUDevices
-        model: dict
-            dict of model properties
-        model_repository: str
-            path to the model repository on the file system
-        model_variant_name_manager: ModelVariantNameManager
         """
         model_name = model.model_name()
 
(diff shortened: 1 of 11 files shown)


model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -212,6 +199,38 @@ def make_model_config(
...
+            model: ModelProfileSpec, ensemble_submodels: List[ModelConfig],
+            model_variant_name_manager: ModelVariantNameManager) -> ModelConfig:
+        """
+        Loads the base model config from the model repository, and then mutates
+        the names to match the ensemble submodels
+        """
+        model_name = model.model_name()
+        model_config_dict = model.get_default_config()
+        ensemble_config_dicts = [
+            model.to_dict() for model in ensemble_submodels
```
tgerdesnv, line 213: seems bad to reuse 'model' variable name to be the same as one passed in

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -212,6 +199,38 @@ def make_model_config(
...
+        (variant_found, variant_name
+        ) = model_variant_name_manager.get_ensemble_model_variant_name(
+            model_name, ensemble_key)
+
+        model_config_dict['name'] = variant_name
+        model_config = ModelConfig.create_from_dictionary(model_config_dict)
+
+        for submodel in ensemble_submodels:
+            variant_name = submodel.get_field("name")
+            submodel_name = variant_name[:variant_name.find("_config_")]
```
tgerdesnv, line 227: can you create a static helper function for this?  I think result_manager.py line 155 would want to use the same.

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -41,6 +41,17 @@ def from_dict(
...
+    @staticmethod
+    def make_ensemble_submodel_key(
+            ensemble_config_dicts: List[Dict]) -> Dict[str, str]:
+        ensemble_names = [
+            ensemble_config_dict["name"]
+            for ensemble_config_dict in ensemble_config_dicts
+        ]
+        ensemble_key = ','.join(ensemble_names)
+
+        return {"key": ensemble_key}
```
tgerdesnv, line 53: Why is the key a dictionary instead of just a string?

model_analyzer/config/generate/model_variant_name_manager.py
```diff
@@ -65,6 +76,29 @@ def get_model_variant_name(self, model_name: str, model_config_dict: Dict,
 
         return (False, model_variant_name)
 
+    def get_ensemble_model_variant_name(
```
tgerdesnv, line 79: This looks basically identical to the existing non-exsemble version. Can we combine the two?

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -59,8 +60,10 @@ def __init__(self, search_config: SearchConfig,
         config: ConfigCommandProfile
             Profile configuration information
         gpus: List of GPUDevices
-        models: List of ConfigModelProfileSpec
+        models: List of ModelProfileSpec
             List of models to profile
+        ensemble_submodels: Dict of List of ModelProfileSpec
```
tgerdesnv, line 65: It looks like this could in theory support multi-model of any combination of ensemble and non-ensemble models?

model_analyzer/model_manager.py
```diff
@@ -127,6 +131,28 @@ def run_models(self, models):
...
+            config=self._config,
+            client=self._client,
+            gpus=self._gpus,
+            model_repository=self._config.model_repository,
+            model_name=model.model_name())
+
+        model_config = ModelConfig.create_from_dictionary(model_config_dict)
+
+        if not model_config.is_ensemble():
+            return None
```
tgerdesnv, line 147: should this return an empty list instead of None?

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -162,19 +162,6 @@ def make_model_config(
         """
         Loads the base model config from the model repository, and then applies the
         parameters in the param_combo on top to create and return a new model config
-
```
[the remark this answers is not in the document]
nv-braf: Do we still need parameters now that we have type checking? They were incorrect, which is why I removed them

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: Changed to submodel

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: Done

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: Yes, I believe that will work. But, at this point I'm not focusing on testing for those combinations.

model_analyzer/model_manager.py
nv-braf: Actually this function is dead code. I've removed it.

model_analyzer/config/generate/base_model_config_generator.py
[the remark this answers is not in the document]
tgerdesnv: I see they were incorrect. It would be nice if they were correct. The type checking tells us what type is passed in, but doesn't give us all the information. For example, if I pass in a bool, stating it is a bool doesn't tell me what that bool means.

model_analyzer/config/generate/model_variant_name_manager.py
tgerdesnv: I'm still confused by this.

nv-braf: For type compatibility when writing into the MVNM

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -223,6 +250,10 @@ def _print_max_batch_size_plateau_warning(self) -> None:
             )
             self._max_batch_size_warning_printed = True
 
+    @staticmethod
+    def extract_model_name_from_variant_name(variant_name):
```
tgerdesnv, line 254: document

model_analyzer/config/generate/model_variant_name_manager.py
tgerdesnv: ok, I see now where a dict is used.

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -212,6 +206,39 @@ def make_model_config(
 
         return model_config
 
+    @staticmethod
+    def make_ensemble_model_config(
+            model: ModelProfileSpec, ensemble_submodels: List[ModelConfig],
+            model_variant_name_manager: ModelVariantNameManager) -> ModelConfig:
+        """
+        Loads the base model config from the model repository, and then mutates
```
tgerdesnv, line 214: can we doc the args better here? It is unclear what model is (the base ensemble model? An ensemble variant?) and unclear if the ensemble_submodels are the names of the base models, or the newly named variants.

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -337,28 +345,86 @@ def _create_default_run_config(self) -> RunConfig:
...
-        return default_run_config
+        default_model_run_config = self._add_submodels_to_model_run_config(
+            default_model_run_config, default_submodel_configs)
+
+        return default_model_run_config
+
+    def _create_default_submodel_configs(
+            self, model: ModelProfileSpec) -> List[ModelConfig]:
+        default_submodel_configs: List[ModelConfig] = []
+        for submodel in self._ensemble_submodels[model.model_name()]:
```
tgerdesnv, line 381: we should assert somewhere if we are ever ensemble with no submodels

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -337,28 +345,86 @@ def _create_default_run_config(self) -> RunConfig:
...
+            default_submodel_configs.append(
+                BaseModelConfigGenerator.make_model_config(
+                    param_combo={},
+                    model=submodel,
+                    model_variant_name_manager=self._model_variant_name_manager)
+            )
+
+        return default_submodel_configs
+
+    def _add_submodels_to_model_run_config(
```
tgerdesnv, line 391: it appears that adding one submodel is a member function of model_run_config, but adding multiple is not? Can we just hide it all in model_run_config?

model_analyzer/config/generate/run_config_generator_factory.py
```diff
@@ -102,29 +110,43 @@ def _create_brute_run_config_generator(
...
+                                     List[ModelProfileSpec]]) -> SearchConfig:
         dimensions = SearchDimensions()
 
-        #yapf: disable
-        for i, model in enumerate(models):
-            dims = RunConfigGeneratorFactory._get_dimensions_for_model(model.supports_batching())
-            dimensions.add_dimensions(i, dims)
-        #yapf: enable
+        index = 0
+        for model in models:
```
tgerdesnv, line 138: at this point, any ensemble models have been removed, and instead we have submodels (and/or normal models)?

model_analyzer/config/run/model_run_config.py
```diff
@@ -117,6 +128,26 @@ def is_legal_combination(self):
...
+        Returns true if the model_config is an ensemble model
+        """
+        return self._model_config.is_ensemble()
+
+    def get_ensemble_subconfig_names(self) -> Optional[List[str]]:
+        """
+        Returns list of Ensemble Subconfig names
+        """
+        return self._model_config.get_ensemble_submodels(
+        ) if self._model_config.is_ensemble() else None
```
tgerdesnv, line 142: is it safer to return empty list instead of None?

tests/test_quick_run_config_generator.py
```diff
@@ -328,6 +329,91 @@ def test_default_config_generation(self):
...
+                return_value=fake_config):
+            models = [
+                ModelProfileSpec(spec=config.profile_models[0],
+                                 config=config,
+                                 client=MagicMock(),
+                                 gpus=MagicMock())
+            ]
+
+        dims = SearchDimensions()
+        dims.add_dimensions(0, [
```
tgerdesnv, line 382: I would expect more dimensions than this. Is this ignored for this test?  Is there any testing that the SearchConfig dimensions are set up correctly? I saw that you added the code to create it

model_analyzer/config/generate/quick_run_config_generator.py
nv-braf: Maybe, but it doesn't go here. This logic works fine even if no submodels exists, it will just return an empty list. 
How much responsibility should MA have to ensure that the user forms a correct config.pbtxt?

model_analyzer/config/generate/run_config_generator_factory.py
nv-braf: Correct.

model_analyzer/config/run/model_run_config.py
nv-braf: No, None is fine and is equivalent to an empty list when checking in an if statement

tests/test_quick_run_config_generator.py
nv-braf: Yes, good catch. There should be 3 more dimensions, but that wasn't the purpose of this test. I just copy/pasted so I could create a search config.
There is no test for dimensions right now. That will come in the next story when I'm creating non-default ensemble configs.

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: Updated and added parameters. Please let me know if what is still unclear.

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: Done

model_analyzer/config/generate/quick_run_config_generator.py
tgerdesnv: >How much responsibility should MA have to ensure that the user forms a correct config.pbtxt?
None, but we should still protect our code to not take bad paths.  I agree that it shouldn't go there. Just mentioning it should happen somewhere.

model_analyzer/config/run/model_run_config.py
tgerdesnv: if (Y):   -- works for both none and empty list
for X in Y:  -- works for empty list, but not for None.

Is there a case where None is better? I hate having None all over our code and needing to make a bunch of types Optional.

nv-braf: This doesn't change the need for optional. Both None and [] will require it.

nv-braf: But, I've changed this to be []

tgerdesnv (approved):

PR #578, opened 2022-12-15
Fix index-based access of constraints in ConstraintManager

pskiran1:
Changes:
- Class ConstraintManager is modified to hold a dictionary containing ModelConstraints object references for each model
- ConstraintManager class static methods have been converted into class methods, replacing static methods with class methods usage
- Added the ConstraintManager object as a new argument to ModelManager, ReportManager, PlotManager, and ResultManager classes
- Added type-checking for ModelManager, ReportManager, PlotManager, and ResultManager classes


model_analyzer/reports/report_manager.py
```diff
@@ -423,6 +426,7 @@ def _find_best_configs(self, report_key):
     def _create_constraint_string(self, report_key: str) -> str:
         constraint_strs = self._build_constraint_strings()
 
+        constraint_str = "None"
```
pskiran1, line 429: This change was made for fixing the following bug:
```
File "/home/nvex/saikiran/triton_related/initial_rampup/editable_ma/github_branch/model_analyzer/model_analyzer/reports/report_manager.py", line 438, in _create_constraint_string
    return constraint_str
UnboundLocalError: local variable 'constraint_str' referenced before assignment
```
The above error happens when we run profiling if a checkpoint already exists and if two profile models are passed with only one having constraints (no global constraints as well).

model_analyzer/plots/plot_manager.py
```diff
@@ -83,7 +84,7 @@ def _create_summary_plot_for_model(self, model_name, plots_key,
         """
 
         for plot_config in self._config.plots:
-            constraints = self._constraints['default']
+            constraints = self._constraints['__default__']
```
pskiran1, line 87: Modified key to access global constrains as `__default__` to look more unique than model names, as model name will not start with `_`

model_analyzer/model_manager.py
```diff
@@ -30,7 +29,7 @@ class ModelManager:
     """
 
     def __init__(self, config, gpus, client, server, metrics_manager,
-                 result_manager, state_manager):
+                 result_manager, state_manager, constraint_manager):
```
nv-braf, line 32: It would be nice to boyscout in type checking along with these edits

model_analyzer/plots/plot_manager.py
```diff
@@ -30,7 +29,7 @@ class PlotManager:
     of plots generated by model analyzer
     """
 
-    def __init__(self, config, result_manager):
+    def __init__(self, config, result_manager, constraint_manager):
```
nv-braf, line 32: Add type checking

model_analyzer/plots/plot_manager.py
nv-braf: We should make this a define. Can you add this to `constants.py`?

model_analyzer/reports/report_manager.py
nv-braf: I'd like to understand this case better. I don't see how we can have a case where global/default constraints do not exist.

model_analyzer/result/model_config_measurement.py
```diff
@@ -102,6 +102,13 @@ def model_config_name(self):
 
         return self._model_config_name
 
+    def model_name(self):
+        """
+        Return the model name
+        """
+
+        return self._model_config_name.partition("_config_")[0]
```
nv-braf, line 110: Can you add a comment that this should be replaced with the extract_model_name_from_variant_name() static method, once the ensemble code gets merged to main

model_analyzer/result/result_manager.py
```diff
@@ -36,18 +36,22 @@ class ResultManager:
     and sort results
     """
 
-    def __init__(self, config, state_manager):
+    def __init__(self, config, state_manager, constraint_manager):
```
nv-braf, line 39: Add type checking

model_analyzer/plots/plot_manager.py
pskiran1: Defined `GLOBAL_CONSTRAINTS_KEY = "__default__"` in `constants.py`

model_analyzer/model_manager.py
pskiran1: Added type checking for modified methods

model_analyzer/result/model_config_measurement.py
pskiran1: Added the following comment in method description,
```
TODO: This method should be replaced with the extract_model_name_from_variant_name() static method,
        once the ensemble code gets merged to main
```

nv-braf (approved):
All changes past 12/17 look good to me

PR #581, opened 2023-01-05
Fix index-based access of constraints in ConstraintManager

pskiran1:
As part of this PR, only Unit test modules are modified and the library code remains the same as PR #578.


tests/common/test_utils.py
```diff
@@ -408,6 +419,37 @@ def construct_run_config_result(avg_gpu_metric_values,
 
     return run_config_result
 
+class ConfigCommandProfileSubClass(ConfigCommandProfile):
```
tgerdesnv, line 422: These optimizations are nice, but I don't think they are what should have been optimized.
I took a closer look at just one test -- test_sorted_results.py:test_add_result().  That one test creates 210 constraint managers(!). I don't see any reason for any test to ever create more than one. Once you fix that, the latency of the full CLI objects should be totally fine without the need for this optimization.

All of the construct_* functions should likely take an optional (or required?) constraint_manager so that each level doesn't have to create a new one.

tgerdesnv (changes requested):

tests/common/test_utils.py
tgerdesnv: Note that the 210 calls was from your old changes, not your new ones. I see now that you have magic mocked out most of the create calls in the new code.  I had assumed based on our discussion yesterday that you had only optimized the CLI code.

I don't see any reason to fake out the CLI code. As long as no test creates more than one constraint_manager I can't imagine that the full CLI code takes more than 1 second of total test run time.

pskiran1:
Closing this PR since we decided to use https://github.com/triton-inference-server/model_analyzer/pull/590 to review only uint test case changes.

tgerdesnv:
Now that #590 has been reviewed, this side branch can be merged into main.

PR #586, opened 2023-01-10
Add ensemble support to summary result tables

nv-braf:
For ensemble models we will now print the instance groups and max batch sizes for the ensemble subconfigs. In addition, the model config path will now be in the format: "ensemble_config_#: submodel_A_config_#, submodel_B_config_#, ...

Here is what the metrics-model-inference looks like:
image

And, here's the metrics-model-gpu:
image

Golden CSVs and unit testing added


model_analyzer/result/result_table_manager.py
```diff
@@ -35,6 +35,7 @@ class ResultTableManager:
         'batch_size': 'Batch',
         'concurrency': 'Concurrency',
         'model_config_path': 'Model Config Path',
+        'ensemble_subconfig_path': 'Ensemble Subconfig Path',
```
tgerdesnv, line 38: Is this header used? I don't see it in the images

nv-braf: Good catch. This is a remnant from when I was going to make the subconfigs a separate column.

tgerdesnv (approved):

PR #588, opened 2023-01-12
Added ensemble support to detailed report PDF

nv-braf:
Updated the sentence below the table to include details about the submodels that make up the ensemble.

image

(table is unchanged just shown for reference)

### DIFF
--- a/model_analyzer/reports/report_manager.py
+++ b/model_analyzer/reports/report_manager.py
@@ -286,7 +286,7 @@ class ReportManager:
             "are the requested configurable plots showing the relationship between "
             "various metrics measured by the Model Analyzer. The above table contains "
             "detailed data for each of the measurements taken for this model config in "
-            "decreasing order of throughput.",
+            "decreasing order of concurrency.",
             font_size=18)
         return detailed_report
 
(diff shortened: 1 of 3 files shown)


model_analyzer/reports/report_manager.py
```diff
@@ -286,7 +286,7 @@ def _build_detailed_report(self, report_model_config):
             "are the requested configurable plots showing the relationship between "
             "various metrics measured by the Model Analyzer. The above table contains "
             "detailed data for each of the measurements taken for this model config in "
-            "decreasing order of throughput.",
+            "decreasing order of concurrency.",
```
tgerdesnv, line 289: is this confirmed via the code?

model_analyzer/reports/report_manager.py
```diff
@@ -950,12 +950,22 @@ def _build_detailed_info(self, model_config_name):
...
+            sentence = f"This <strong>{model_config_name}</strong> consists of the following sub-model configurations:"
+
+            for ensemble_subconfig in run_config.ensemble_subconfigs():
+                sentence = sentence + '<LI> ' + self._create_summary_config_info(
+                    ensemble_subconfig) + ' </LI>'
+
+            sentence = sentence + f"<br>{len(measurements)} measurement(s) were obtained for the model config on {gpu_cpu_string}."
+        else:
+            sentence = (
+                f"The model config <strong>{model_config_name}</strong> uses {instance_group_string} "
```
tgerdesnv, line 964: old code listed instance counts and max batch size, etc. Should we be doing this for all of the submodels?

model_analyzer/reports/report_manager.py
nv-braf: It's actually decreasing order of throughput or latency (which is what I meant to write!) based on online vs. offline mode. I've made the fix to check the mode and print out the correct sorting type used.

model_analyzer/reports/report_manager.py
tgerdesnv: yes, yes you are. I'm blind

tgerdesnv (approved):

PR #590, opened 2023-01-17
Fix constraint manager

pskiran1:


tests/test_neighborhood.py
```diff
@@ -391,29 +397,31 @@ def test_get_constraints_passing_measurements(self):
...
+                        "perf_latency_p99": {"max": 300}
+                    }
                 }
-            }
+            }}
         )
 
         rcm0 = self._construct_rcm(throughput=100, latency=50)  # pass
-        rcm0.set_constraint_manager(constraint_manager=constraint_manager)
+        rcm0.set_constraint_manager(constraint_manager)
```
tgerdesnv, line 412: Are these calls needed? It looks like the constraint manager is already set in the function the line above

tests/common/test_utils.py
```diff
@@ -430,61 +419,35 @@ def construct_run_config_result(avg_gpu_metric_values,
...
+
+class CLISubclass(CLI):
+    # Override CLI.parse() method
+    def parse(self, yaml_config):
+        config_profile = ConfigCommandProfileSubClass()
+        config_profile.set_config_values(argparse.Namespace(), yaml_config)
+        return config_profile
 
-def construct_constraint_manager(constraints=None, yaml_str=None):
+def construct_constraint_manager(yaml_config: Optional[Dict[str, Dict[str, Any]]]=None):
```
tgerdesnv, line 435: Can yaml_config be a string for simplicity? Then you can utilize the existing evaluate_mock_config() code, and all tests will follow the same path.

tests/test_model_config_generator.py
```diff
@@ -859,14 +861,15 @@ def _run_and_test_model_config_generator(self,
         finally:
             self.mock_model_config.stop()
 
-    def _get_next_fake_results(self):
+    def _get_next_fake_results(self, constraint_manager):
```
tgerdesnv, line 864: Do any of these tests actually require a constraint_manager?

tests/test_sorted_results.py
```diff
@@ -162,15 +173,20 @@ def test_add_results_failing_to_passing(self):
...
+
         # Create 10 failing results
         avg_gpu_metrics = {0: {'gpu_used_memory': 6000, 'gpu_utilization': 60}}
         for i in sample(range(10), 10):
             model_name = str(i)
             model_config_names = [f"{model_name}_config_0"]
-            constraints = {
-                model_name: {'perf_throughput': {'min': 1000}}
-            }
-            constraint_manager = construct_constraint_manager(constraints=constraints)
```
tgerdesnv, line 173: This is the type code I was referring to about only creating one constraint_manager for the test, instead of creating a new one over and over. It looks like this has been fixed.

tests/test_sorted_results.py
```diff
@@ -212,10 +224,15 @@ def test_add_results_failing_to_passing(self):
...
+        constraint_manager = construct_constraint_manager({
+            "profile_models": {
+                str(i): {} for i in range(10)
+            }
+        })
+
         for i in sample(range(10), 10):
             model_name = str(i)
             model_config_names = [f"{model_name}_config_0"]
-            constraint_manager = construct_constraint_manager(constraints={model_name:{}})
```
tgerdesnv, line 218: you fixed this one as well 👍

pskiran1: Yes 👍

tests/test_sorted_results.py
pskiran1: Yes, it's fixed now 👍

tests/test_model_config_generator.py
pskiran1: This additional change was verified and confirmed to not be required. Updated the code.

tests/common/test_utils.py
pskiran1: This has been fixed 👍

tests/test_neighborhood.py
pskiran1: Yes, Tim. As rcm objects are freshly created here, we need to call set_constraint_manager() per rcm object

nv-braf (approved):
We are about 10 seconds slower through the unit test suite on my machine (83 -> 93), but that is acceptable.

tests/test_neighborhood.py
tgerdesnv: I'm pretty sure that line 411 already handles this

tgerdesnv: ok, yes I see now.

PR #595, opened 2023-01-19
Allow min/max setting in quick search

nv-braf:
This change enables the user to specify a min/max concurrency, model batch size or instance count in quick search.

Added unit testing to cover all relevant cases and have tried this out with an ensemble model and confirmed it is working.

### DIFF
--- a/model_analyzer/config/generate/quick_run_config_generator.py
+++ b/model_analyzer/config/generate/quick_run_config_generator.py
@@ -374,7 +374,7 @@ class QuickRunConfigGenerator(ConfigGeneratorInterface):
 
         kind = "KIND_CPU" if model.cpu_only() else "KIND_GPU"
         instance_count = self._calculate_instance_count(
-            dimension_values['instance_count'])
+            int(dimension_values['instance_count']))
 
         param_combo: dict = {
             'instance_group': [{
@@ -385,7 +385,7 @@ class QuickRunConfigGenerator(ConfigGeneratorInterface):
 
         if 'max_batch_size' in dimension_values:
             param_combo['max_batch_size'] = self._calculate_model_batch_size(
-                dimension_values['max_batch_size'])
+                int(dimension_values['max_batch_size']))
 
         if model.supports_dynamic_batching():
             param_combo['dynamic_batching'] = {}
@@ -407,8 +407,9 @@ class QuickRunConfigGenerator(ConfigGeneratorInterface):
         perf_analyzer_config.update_config_from_profile_config(
             model_variant_name, self._config)
 
-        dimension_batch_size = dimension_values.get("max_batch_size", 1)
-        dimension_instance_count = dimension_values.get("instance_count", 1)
+        dimension_batch_size = int(dimension_values.get("max_batch_size", 1))
+        dimension_instance_count = int(dimension_values.get(
+            "instance_count", 1))
 
         model_batch_size = self._calculate_model_batch_size(
             dimension_batch_size)
--- a/model_analyzer/config/generate/quick_run_config_generator.py
+++ b/model_analyzer/config/generate/quick_run_config_generator.py
@@ -407,13 +407,30 @@ class QuickRunConfigGenerator(ConfigGeneratorInterface):
         model_batch_size = dimension_values.get("max_batch_size", 1)
         instance_count = dimension_values.get("instance_count", 1)
 
-        concurrency = 2 * model_batch_size * instance_count
+        concurrency = self._calculate_concurrency(model_batch_size,
+                                                  instance_count)
 
         perf_config_params = {'batch-size': 1, 'concurrency-range': concurrency}
         perf_analyzer_config.update_config(perf_config_params)
 
         perf_analyzer_config.update_config(model.perf_analyzer_flags())
         return perf_analyzer_config
+
+    def _calculate_concurrency(self, model_batch_size: int,
+                               instance_count: int) -> int:
+        concurrency = 2 * model_batch_size * instance_count
+
+        min_concurrency_set = self._config.get_config(
+        )['run_config_search_min_concurrency'].set_by_config()
+        if min_concurrency_set and concurrency < self._config.run_config_search_min_concurrency:
+            return self._config.run_config_search_min_concurrency
+
+        max_concurrency_set = self._config.get_config(
+        )['run_config_search_max_concurrency'].set_by_config()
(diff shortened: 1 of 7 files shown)


model_analyzer/config/input/config_command.py
```diff
@@ -184,7 +185,7 @@ def _check_for_quick_search_incompatability(
             return
 
         self._check_no_search_disable(args, yaml_config)
-        self._check_no_search_values(args, yaml_config)
+        # self._check_no_search_values(args, yaml_config)
```
tgerdesnv, line 188: commented out?

model_analyzer/config/input/config_field.py
```diff
@@ -174,6 +179,8 @@ def set_default_value(self, default_value):
 
         self._default_value = default_value
 
+        self._set_by_user = False
```
tgerdesnv, line 182: is set by user used anywhere?

tests/test_config.py
```diff
@@ -1809,18 +1812,6 @@ def test_quick_search_mode(self):
                                          '--run-config-search-disable',
                                          use_value=False,
                                          use_list=False)
-        self._test_quick_search_with_rcs(args, yaml_content,
```
tgerdesnv, line 1812: are these invalid now?

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -404,17 +407,72 @@ def _get_next_perf_analyzer_config(self, model_variant_name: str,
         perf_analyzer_config.update_config_from_profile_config(
             model_variant_name, self._config)
 
-        model_batch_size = dimension_values.get("max_batch_size", 1)
-        instance_count = dimension_values.get("instance_count", 1)
+        dimension_batch_size = int(dimension_values.get("max_batch_size", 1))
```
tgerdesnv, line 410: can you hide all of the retrieval + calculation in the function? So we don't have half of it done here with a temporary value?

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -404,17 +407,72 @@ def _get_next_perf_analyzer_config(self, model_variant_name: str,
...
         return perf_analyzer_config
 
+    def _calculate_model_batch_size(self, batch_size: int) -> int:
+        min_batch_size_set = self._config.get_config(
+        )['run_config_search_min_model_batch_size'].set_by_config()
+
+        max_batch_size_set = self._config.get_config(
+        )['run_config_search_max_model_batch_size'].set_by_config()
+
+        if min_batch_size_set and batch_size < self._config.run_config_search_min_model_batch_size:
```
tgerdesnv, line 435: My thought was that we would constrain the dimensions. It seems your approach was to make the dimensions above a certain point clamped to the min/max value. Does that have any implication on the algorithm or the ability to rerun?

tests/test_config.py
nv-braf: These are all valid now, so we don't need checks that they are illegal anymore.

model_analyzer/config/input/config_command.py
nv-braf: I've removed this call and the method.

model_analyzer/config/input/config_field.py
nv-braf: Good catch, this should be set_by_config

model_analyzer/config/generate/quick_run_config_generator.py
nv-braf: You bring up a good point, It doesn't have any implication on the algorithm, but it does present a rerun issue. As the checkpoint result is not valid if the user specifies a restriction, we clamp, and then reruns without the restriction. On the subsequent run we will use the clamped result.

I'll need to redo this to clamp the dimension in the algorithm.

nv-braf: It doesn't effect rerun, since we're modifying the model/PA config, which is stored in the checkpoint. So, if the user reruns without clamping we will create a new config.

I'm not convinced this negatively effects the algorithm, as I don't think it cares about the clamping, but we should talk.

nv-braf: I'm good with what we have. I've also added some documentation. Please re-review.

model_analyzer/config/input/config_field.py
```diff
@@ -200,3 +207,9 @@ def required(self):
 
     def set_name(self, name):
         self._field_type.set_name(name)
+
+    def set_by_config(self) -> bool:
```
tgerdesnv, line 211: any function that starts with `set` implies that it is taking an input and setting something. Can you change this to `is_set_by_config` or `is_specified` or something else along those lines?

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -404,17 +406,70 @@ def _get_next_perf_analyzer_config(self, model_variant_name: str,
...
         perf_analyzer_config.update_config(perf_config_params)
 
         perf_analyzer_config.update_config(model.perf_analyzer_flags())
         return perf_analyzer_config
 
+    def _calculate_model_batch_size(
+            self, dimension_values: Dict[str, Union[int, float]]) -> int:
+        batch_size = int(dimension_values.get("max_batch_size", 1))
+
+        min_batch_size_set = self._config.get_config(
```
tgerdesnv, line 421: set in this case is also confusing. Implying it is part of a set or contains a set. I would either do `is_min_batch_size_set` or `is_min_batch_size_specified`

tgerdesnv (changes requested):

model_analyzer/config/generate/quick_run_config_generator.py
nv-braf: I've changed this (and the others) to `min_batch_size_is_set_by_config`

model_analyzer/config/input/config_command.py
```diff
@@ -124,9 +124,10 @@ def _set_field_values(self, args: Namespace,
             config_value = self._get_config_value(key, args, yaml_config)
 
             if config_value:
-                self._fields[key].set_value(config_value)
+                self._fields[key].set_value(config_value, set_by_config=True)
```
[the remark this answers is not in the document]
tgerdesnv: I was confusing config_field with config_value (which has tons of children). I think we're ok now

tgerdesnv (approved):
as long as CI passes this should be ok to merge

PR #596, opened 2023-01-23
Checking for illegal inputs from user for an ensemble model

nv-braf:
This checks for the two illegal inputs the user could provide for an ensemble model:

- More than one model being profiled
- Brute force search mode

### DIFF
--- a/model_analyzer/config/input/config_command.py
+++ b/model_analyzer/config/input/config_command.py
@@ -113,8 +113,9 @@ class ConfigCommand:
     def _check_for_illegal_config_settings(
             self, args: Namespace, yaml_config: Optional[Dict[str,
                                                               List]]) -> None:
-        # Illegal config settings check for ensemble is done in model_manager,
+        # Note: Illegal config settings check for ensemble is done in model_manager,
         # since we don't yet know the type of model being profiled
+
         self._check_for_duplicate_profile_models_option(args, yaml_config)
         self._check_for_multi_model_incompatability(args, yaml_config)
         self._check_for_quick_search_incompatability(args, yaml_config)
--- a/model_analyzer/config/input/config_command.py
+++ b/model_analyzer/config/input/config_command.py
@@ -113,8 +113,8 @@ class ConfigCommand:
     def _check_for_illegal_config_settings(
             self, args: Namespace, yaml_config: Optional[Dict[str,
                                                               List]]) -> None:
-        # Illegal config settings for ensemble is done in model_manager,
-        # since we don't yet know if the model is an ensemble
+        # Illegal config settings check for ensemble is done in model_manager,
+        # since we don't yet know the type of model being profiled
         self._check_for_duplicate_profile_models_option(args, yaml_config)
         self._check_for_multi_model_incompatability(args, yaml_config)
         self._check_for_quick_search_incompatability(args, yaml_config)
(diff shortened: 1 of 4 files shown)


model_analyzer/model_manager.py
```diff
@@ -121,7 +129,7 @@ def run_models(self, models):
         self._server.update_config(params=server_config_copy.server_args())
 
         model_variant_name_manager_dict = self._state_manager.default_encode(
-            rcg._model_variant_name_manager)
+            rcg._model_variant_name_manager)  # type: ignore
```
nv-braf, line 132: My boyscouting of type checking I exposed this error:
```../../model_analyzer/model_manager.py:132: error: "ConfigGeneratorInterface" has no attribute "_model_variant_name_manager"  [attr-defined]```

Don't know why it's showing up or how (or if we) need to fix this. I think it's safe to ignore this, as I don't see any real issue as MVNM is specified everywhere it needs to be.

File: model_analyzer/config/input/config_command.py [hunk 2]
```diff
@@ -113,6 +113,8 @@ def _load_yaml_config(self, args: Namespace) -> Optional[Dict[str, List]]:
     def _check_for_illegal_config_settings(
             self, args: Namespace, yaml_config: Optional[Dict[str,
                                                               List]]) -> None:
+        # Illegal config settings check for ensemble is done in model_manager,
```
tgerdesnv, line 116: This implies it is a comment for the lines below it.
Can you just start it with "Note: Illegal ..." so it is clear it is a standalone note comment, and then add an empty line before the other code?

model_analyzer/model_manager.py
tgerdesnv: well accessing mvnm isn't part of the interface. If every ConfigGenerator is guaranteed to have one, it can/should be added to the interface. If not every one is guaranteed to have one, then this is unsafe code.

It looks like the mvnm is passed in to the object on line 108. Any reason we can't use `self._model_variant_name_manager` instead of `rcg._model_variant_name_manager`?

model_analyzer/model_manager.py
```diff
@@ -136,6 +144,23 @@ def _get_triton_server_flags(self, models):
...
+            if model_config.is_ensemble():
+                if len(models) > 1:
+                    raise TritonModelAnalyzerException(
+                        f'\nProfiling of multiple models is not supported for ensemble models'
+                    )
+
+                if self._config.run_config_search_mode != 'quick':
+                    raise TritonModelAnalyzerException(
+                        f'\nBrute search mode is not supported for ensemble models'
+                        '\nPlease use quick search mode')
```
tgerdesnv, line 162: can you add the exact option they need to add to the error message? If they run into this I don't want them to have to look through docs to figure out how to do it.

model_analyzer/config/input/config_command.py
nv-braf: Done

model_analyzer/model_manager.py
nv-braf: I don't see why not. Changed.

model_analyzer/model_manager.py
nv-braf: Done.

model_analyzer/model_manager.py
tgerdesnv: If this breaks anything, would your unit test catch it?  Can you run another CI before you merge?

tgerdesnv: nevermind this is going into a side branch anyways

tgerdesnv (approved):

model_analyzer/model_manager.py
nv-braf: I'm running a CI now. And, yes I believe unit tests would expose this, but I don't see an issue here as we are passing references since they are classes.

PR #602, opened 2023-02-03
Add Ensemble documentation

nv-braf:
Adding ensemble model support to our documentation. Will file a separate story to add an example ensemble to our documentation when we find a suitable model. I've also removed the `early access` tag from multi-model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Triton Inference Server.
   find the optimal settings when profiling multiple concurrent models, utilizing the Quick Search alogrithm
 
 - [Ensemble Model Search](docs/config_search.md#ensemble-model-search): Model Analyzer can help you find the optimal
-  settings when profiling an ensemble model, utilizing the Quick Search algorithm
+  settings when profiling a non-BLS ensemble model, utilizing the Quick Search algorithm
 
 - [Detailed and summary reports](docs/report.md): Model Analyzer is able to generate
   summarized and detailed reports that can help you better understand the trade-offs
--- a/README.md
+++ b/README.md
@@ -43,11 +43,11 @@ Triton Inference Server.
   to test the model with different concurrency and batch sizes of requests. Using
   [Manual Config Search](docs/config_search.md#manual-brute-search), you can create manual sweeps for every parameter that can be specified in the model configuration.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -43,9 +43,12 @@ Triton Inference Server.
   to test the model with different concurrency and batch sizes of requests. Using
   [Manual Config Search](docs/config_search.md#manual-brute-search), you can create manual sweeps for every parameter that can be specified in the model configuration.
 
-- [Multi-Model Search](docs/config_search.md#multi-model-search-mode): **EARLY ACCESS** - Model Analyzer can help you
+- [Multi-Model Search](docs/config_search.md#multi-model-search-mode): Model Analyzer can help you
   find the optimal settings when profiling multiple concurrent models, utilizing our Quick Search alogrithm
 
+- [Ensemble Model Search](docs/config_search.md#ensemble-model-search): Model Analyzer can help you find the optimal
+  settings when profiling an ensemble model, utilizing our Quick Search algorithm
```
tgerdesnv, line 50: I don't think we want to say "our". Just say "the"? (Same for multi-model line 47)

docs/config_search.md
```diff
@@ -187,9 +184,19 @@ Using the `--run-config-search-<min/max>...` CLI options you have the ability to
 
 Note: That by default quick search runs unbounded and ignores any default values for these settings
 
+## Ensemble Model Search
```
tgerdesnv, line 187: Should we explicitly say we do not support BLS? Or is that considered different enough from ensembles?

docs/config_search.md
```diff
@@ -187,9 +184,19 @@ Using the `--run-config-search-<min/max>...` CLI options you have the ability to
...
+
+_This mode has the following limitations:_
+
+- Can only be run in `quick` search mode
```
tgerdesnv, line 191: Should we put a max submodel number here? At least as a recommendation?

nv-braf: Not sure. I mention below that it should take less than one hour for a 4 submodel run, but if they want to run a larger submodel ensemble I don't see why we should restrict them.

tgerdesnv: I think we should set expectations properly. Let's be upfront that it is only verified or expected to work properly on 4 or 5 and less

tgerdesnv (approved):

Issue #607, opened 2023-02-15
 question about --triton-launch-mode=docker   


BobDLA:
The model requires for several other pip dependency not installed in 
nvcr.io/nvidia/tritonserver:23.01-py3

So, I update the docker image by a docker file like 

from nvcr.io/nvidia/tritonserver:23.01-py3
RUN pip install opencv-contrib-python-headless  onnxruntime-gpu 

And create a tag like triton_onnx:0.1

But when I start the model_analyzer, I got error

07:31:58 [Model Analyzer] Initializing GPUDevice handles
07:31:59 [Model Analyzer] Using GPU 0 NVIDIA GeForce RTX 2060 with UUID GPU-f7572576-2bd0-ee4f-cd17-0a7a8ec6da5b
07:31:59 [Model Analyzer] Starting a Triton Server using docker
07:31:59 [Model Analyzer] Pulling docker image triton_onnx:0.1
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/docker/api/client.py", line 268, in _raise_for_status
    response.raise_for_status()
  File "/usr/local/lib/python3.8/dist-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http+docker://localhost/v1.40/images/create?tag=0.1&fromImage=triton_onnx

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/model-analyzer", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/model_analyzer/entrypoint.py", line 256, in main
    client, server = get_triton_handles(config, gpus)
  File "/usr/local/lib/python3.8/dist-packages/model_analyzer/entrypoint.py", line 130, in get_triton_handles
    server = TritonServerFactory.get_server_handle(config, gpus)
  File "/usr/local/lib/python3.8/dist-packages/model_analyzer/triton/server/server_factory.py", line 129, in get_server_handle
    server = TritonServerFactory._get_docker_server_handle(
  File "/usr/local/lib/python3.8/dist-packages/model_analyzer/triton/server/server_factory.py", line 216, in _get_docker_server_handle
    server = TritonServerFactory.create_server_docker(
  File "/usr/local/lib/python3.8/dist-packages/model_analyzer/triton/server/server_factory.py", line 69, in create_server_docker
[...]

PR #609, opened 2023-02-15
Fix support for custom server docker image

tgerdesnv:
Fix for https://github.com/triton-inference-server/model_analyzer/issues/607

### DIFF
--- a/model_analyzer/triton/server/server_docker.py
+++ b/model_analyzer/triton/server/server_docker.py
@@ -68,8 +68,11 @@ class TritonServerDocker(TritonServer):
         assert self._server_config['model-repository'], \
             "Triton Server requires --model-repository argument to be set."
 
-        logger.info(f"Pulling docker image {self._tritonserver_image}")
-        self._docker_client.images.pull(self._tritonserver_image)
+        try:
+            self._docker_client.images.get(self._tritonserver_image)
+        except:
+            logger.info(f"Pulling docker image {self._tritonserver_image}")
+            self._docker_client.images.pull(self._tritonserver_image)
 
     def start(self, env=None):
         """


model_analyzer/triton/server/server_docker.py
```diff
@@ -68,8 +68,11 @@ def __init__(self, image, config, gpus, log_path, mounts, labels, shm_size):
         assert self._server_config['model-repository'], \
             "Triton Server requires --model-repository argument to be set."
 
-        logger.info(f"Pulling docker image {self._tritonserver_image}")
-        self._docker_client.images.pull(self._tritonserver_image)
+        try:
```
tgerdesnv, line 71: I didn't see any way in the python docker module to say "does this image exist". I could have just put a try around the `pull`, but this seems a little more obvious what the code is doing, and will suppress the `pulling docker image` printout when the image exists locally (either because it was built locally, or it was already pulled)

model_analyzer/triton/server/server_docker.py
```diff
@@ -68,8 +68,11 @@ def __init__(self, image, config, gpus, log_path, mounts, labels, shm_size):
...
+            self._docker_client.images.get(self._tritonserver_image)
+        except:
+            logger.info(f"Pulling docker image {self._tritonserver_image}")
+            self._docker_client.images.pull(self._tritonserver_image)
```
nv-braf, line 75: Does there also need to be a `try` around the pull? What happens if the pull fails?

tgerdesnv: It will fail -- as it should. At that point we don't have any image to use.  
I could put a better failure message there though. I'm not sure if nested `try` blocks is good practice though.

nv-braf (approved):

PR #610, opened 2023-02-15
Add Multi-Model Quick Start

nv-braf:
- Added a quick start for multi-model. 
- Profiling two models: add_sub and resnet50_libtorch
- Have both runs from CLI options and YAML
- Show summary results and then give an example of how to run with constraints


examples/quick-start/resnet50_libtorch/config.pbtxt
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
```
dyastremsky, line 1: These additions and changes look great, fantastic work! Could you double check whether this is the right header? I think we now use "NVIDIA CORPORATION & AFFILIATES" on copyright headers.

nv-braf: Done

dyastremsky (approved):

docs/mm_quick_start.md
```diff
@@ -0,0 +1,181 @@
...
+    --output-model-repository-path <path-to-output-model-repo>/<output_dir>
+    --export-path profile_results
+    --run-config-profile-models-concurrently-enable
+```
+
+**Important:** You must specify an `<output_dir>` subdirectory. You cannot have `--output-model-repository-path` point directly to `<path-to-output-model-repo>`
+
+**Important:** If you already ran this earlier in the container, you can use the `--override-output-model-repository` option to overwrite the earlier results.
+
+**Important**: All models must be in the same directory
```
tgerdesnv, line 94: Is this saying they all need to be in the same model repository? I don't think "in the same directory" is correct

docs/mm_quick_start.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Important:** The checkpoint directory should be removed between consecutive runs of
```
tgerdesnv, line 96: Is this comment necessary? Especially with the one on line 92?

docs/mm_quick_start.md
```diff
@@ -0,0 +1,181 @@
...
+the `model-analyzer profile` command.<br><br>
+
+---
+
+The `--run-config-profile-models-concurrently-enable` option tells MA to load and optimize both models concurrently using the [Quick Search](config_search.md#quick-search-mode) algortihm.
```
tgerdesnv, line 101: algorithm

docs/mm_quick_start.md
```diff
@@ -0,0 +1,181 @@
...
+
+This will profile both models concurrently using the quick search algorithm to find the maximal throughput gain for both models by iterating across instance counts and batch sizes. By default, the algorithm is attempting to find the best balance of gains for each model, not the best combined total throughput.
```
tgerdesnv, line 103: you just said "using quick search algorithm" in the previous sentence

tgerdesnv: wrong tense in "is attempting".  Should be "will attempt"

docs/mm_quick_start.md
tgerdesnv: and I think we should expand all "MA" to be "Model Analyzer"

docs/mm_quick_start.md
```diff
@@ -0,0 +1,181 @@
...
+                              |--- summaries
+                              .        |--- add_sub,resnet50_libtorch
+                              .                |--- result_summary.pdf
+```
+
+## `Optional:` Add constraints
+
+---
+
+In the above results you can see a large (472%) throughput increase, but this is at the expense of a large increase in latency. By creating a YAML config file we can specify per model latency constraints:
```
tgerdesnv, line 148: I love this example and flow!

docs/mm_quick_start.md
```diff
@@ -0,0 +1,181 @@
...
+
+---
+
+Here is an example result summary with the above constraints, run on a TITAN RTX GPU:
+![Result Summary Constraint Top](../examples/mm_result_summary_constraint_top.jpg)
+![Result Summary Constraint Table](../examples/mm_result_summary_constraint_table.jpg)
+
+---
+
+You will observe that the top configuration here has a higher total combined throughput than the unconstrained run, however the total percentage gain is lower due to the large decrease in throuhgput for the `resnet50_libtorch` model.
```
tgerdesnv, line 181: throughput

tgerdesnv: is there a way to run the whole thing through spell check?

docs/mm_quick_start.md
```diff
@@ -0,0 +1,181 @@
...
+---
+
+Here is an example result summary, run on a TITAN RTX GPU:
+
+![Result Summary Top](../examples/mm_result_summary_top.jpg)
+![Result Summary Table](../examples/mm_result_summary_table.jpg)
+
+You will note that the top model configuration has a lower total combined throughput than either the 2nd or 3rd configurations. However, it does a better job at increasing the total percentage throughput gain versus the other configurations.
+
+For example, comparing the top config (42,24) to the third best (37,19), you will see that while the gain in `add_sub` throughput is large (~1600 infer/sec) it is only a 3.5% increase. However, this is offset by a decrease in `resnet50_libtorch` throughput (~75 infer/sec), which represents an 8% decrease in throughput. Therefore, the (42,24) is deemed to be a better configuration by MA.
```
tgerdesnv, line 116: It wasn't clear to me what 42,24 and 37,19 were. I thought they might be throughput values or batch sizes. I see now they are config names. You might want to clarify.

examples/quick-start/add_sub/config.pbtxt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tgerdesnv, line 1: It won't let me comment on the binary file, so commenting here. Do we really want a 100 MB file in our repo? Is there an easy public place it can be downloaded from?
If not, did you add the file normally? Or did you do some special git magic for large files? There is a way to have large files hosted outside of the normal repo. I forget the term or how it is done, but you should follow up with Elias or Misha.

tgerdesnv (changes requested):

tgerdesnv:
@debermudez please review this. I'd like to get eyes from someone who isn't as familiar so we can find places where we are making assumptions about the user's knowledge.

docs/mm_quick_start.md
nv-braf: Removed

docs/mm_quick_start.md
nv-braf: Fixed

docs/mm_quick_start.md
nv-braf: Fixed wording & removed quick search

docs/mm_quick_start.md
nv-braf: Thanks! It took me a while to find a combination of models that would produce this result, but I thought it was important to illustrate this example to the user.

docs/mm_quick_start.md
nv-braf: Yes, I found a spell check extension and fixed all spelling errors

docs/mm_quick_start.md
nv-braf: Good point. I've changed this to be the full config name.

examples/quick-start/add_sub/config.pbtxt
nv-braf: I added the file normally. It is in the example directory - so it is not pulled by when a user downloads our package. Only developers will see this size increase.

nv-braf: Talked w/ Misha and Iman. I'm going to find a smaller model to add as an example. The Triton server example directory has a number of small models and I'm hoping one of them produces similar results.

debermudez: Do we not have a place to store these files?

- NGC
- an S3 bucket
- a digitalocean space

This seems like a solid example and hindering it because of 100MB of storage seems counterintuitive to me

dyastremsky: Would using a Python model be acceptable? They're quite small. For example, there's a Python version of add_sub [here](https://github.com/triton-inference-server/server/tree/main/qa/python_models/add_sub). The parent directory of that folder has a bunch of Python models which are small to store.

I imagine the concern over storage size isn't where to store it as much as that anyone cloning this repo will download the model by default. That'll take more time and use more storage space for this repo. Even if you change the defaults to avoid the MA repo growing in size, if this can be done with a smaller model, it'd probably be preferable for users.

nv-braf: Thanks, David. I've switched to using `resnet50_python` for the example. The model.py is much smaller and I was able to get similar results.

tgerdesnv:
LGTM. Waiting Elias review

tgerdesnv (dismissed):
Wait for Elias's review

dyastremsky (approved):

debermudez (approved):
I think is a very complete example and is good to ship. 
approved on my end.

PR #612, opened 2023-02-16
Rewrite config_search.md

nv-braf:
Overhaul of `config_search.md` to include more examples and be more explicit about our different modes, how they operate, and what options exist in each mode.

### DIFF
--- a/docs/config_search.md
+++ b/docs/config_search.md
@@ -282,6 +282,8 @@ Ensemble models can be optimized using the Quick Search mode's hill climbing alg
 
 After MA has found the best config(s), it will then sweep the top-N configurations found (specified by `--num-configs-per-model`) over the concurrency range before generation of the summary reports.
 
+---
+
 ## Multi-Model Search Mode
 
 _This mode has the following limitations:_
--- a/docs/config_search.md
+++ b/docs/config_search.md
@@ -251,7 +251,7 @@ profile_models:
 
 Using the `--run-config-search-<min/max>...` config options you have the ability to clamp the algorithm's upper or lower bounds for the model's batch size and instance count, as well as the client's request concurrency.
 
-_Note: That by default quick search runs unbounded and ignores any default values for these settings_
+_Note: By default, quick search runs unbounded and ignores any default values for these settings_
 
 ---
 
--- a/docs/config_search.md
+++ b/docs/config_search.md
@@ -14,9 +14,24 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
+# Table of Contents
+
+- [Search Modes](#search-modes)
+- [Default Search Mode](#default-search-mode)
+- [Brute Search Mode](#brute-search-mode)
+  - [Automatic Brute Search](#automatic-brute-search)
+  - [Manual Brute Search](#manual-brute-search)
+- [Quick Search Mode](#quick-search-mode)
+- [Ensemble Model Search](#ensemble-model-search)
+- [Multi-Model Search Mode](#multi-model-search-mode)
+
+<br>
+
 # Model Config Search
 
-Model Analyzer's `profile` subcommand supports multiple modes when searching to find the best model configuration.
+## Search Modes
+
+Model Analyzer's `profile` subcommand supports multiple modes when searching to find the best model configuration:
 
 - [Brute Force Search](config_search.md#brute-search-mode)
   - **Search type:** Brute-force sweep of the cross product of all possible configurations
@@ -30,11 +45,24 @@ Model Analyzer's `profile` subcommand supports multiple modes when searching to 
     - Single ensemble models
     - Multiple models being profiled concurrently
   - **Command:** `--run-config-search-mode quick`
+
+---
+
+## Default Search Mode
+
+Model Analyzer's default search mode depends on the type of model and if you are profiling models concurrently (in the case of multiple models):
+
+- [Sequential (single or multi-model) Search](config_search.md#brute-search-mode)
+  - **Default Search type:** [Brute Force Search](config_search.md#brute-search-mode)
+  - **Command:** N/A
 - [Concurrent / Multi-model Search](config_search.md#multi-model-search-mode)
-  - **Search type:** Heuristic concurrent sweep of all models' search spaces to find an optimal configuration for all models
-  - **Default for:**
-    - None
+  - **Default Search type:** [Quick Search](config_search.md#quick-search-mode)
   - **Command:** `--run-config-profile-models-concurrently-enable`
+- [Ensemble Model Search](config_search.md#ensemble-model-search):
+  - **Default Search type:** [Quick Search](config_search.md#quick-search-mode)
+  - **Command:** N/A
+
+---
 
 ## Brute Search Mode
 
--- a/docs/config_search.md
+++ b/docs/config_search.md
@@ -115,7 +115,7 @@ You can also modify the minimum/maximum values that the automatic search space w
 
 ---
 
-### [Request Concurrency Search Space](<(https://github.com/triton-inference-server/client/blob/main/src/c++/perf_analyzer/README.md#request-concurrency)>)
+### [Request Concurrency Search Space](https://github.com/triton-inference-server/client/blob/main/src/c++/perf_analyzer/README.md#request-concurrency)
 
 - `Default:` 1 to 1024 concurrencies, sweeping over powers of 2 (i.e. 1, 2, 4, 8, ...)
 - `--run-config-search-min-concurrency: <val>`: Changes the request concurrency minimum automatic search space value
--- a/docs/config_search.md
+++ b/docs/config_search.md
@@ -19,26 +19,24 @@ limitations under the License.
 Model Analyzer's `profile` subcommand supports multiple modes when searching to find the best model configuration.
 
 - [Brute Force Search](config_search.md#brute-search-mode)
-  - `Search type:` Brute-force sweep of the cross product of all possible configurations
-  - `Default for:`
+  - **Search type:** Brute-force sweep of the cross product of all possible configurations
+  - **Default for:**
     - Single non-ensemble models
     - Multiple models being profiled sequentially
-  - `Command:` `--run-config-search-mode brute`
+  - **Command:** `--run-config-search-mode brute`
 - [Quick Search](config_search.md#quick-search-mode)
-  - `Search type:` Heuristic sweep using a hill-climbing algorithm to find an optimal configuration
-  - `Default for:`
+  - **Search type:** Heuristic sweep using a hill-climbing algorithm to find an optimal configuration
+  - **Default for:**
     - Single ensemble models
-    - Multiple models beging profiled concurrently
-  - `Command:` `--run-config-search-mode quick`
+    - Multiple models being profiled concurrently
+  - **Command:** `--run-config-search-mode quick`
 - [Concurrent / Multi-model Search](config_search.md#multi-model-search-mode)
-  - `Search type:` Heuristic concurrent sweep of all models' search spaces to find an optimal configuration for all models
-  - `Default for: `
+  - **Search type:** Heuristic concurrent sweep of all models' search spaces to find an optimal configuration for all models
+  - **Default for:**
     - None
-  - `Command:` `--run-config-profile-models-concurrently-enable`
+  - **Command:** `--run-config-profile-models-concurrently-enable`
 
 ## Brute Search Mode
-
----
 
 **Default search mode when profiling non-ensemble models sequentially**
 
@@ -72,8 +70,6 @@ profile_models:
 ```
 
 You can also modify the minimum/maximum values that the automatic search space will iterate through:
-
----
 
 ### [Instance Group Search Space](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#instance-groups)
 
@@ -169,7 +165,7 @@ profile_models:
 
 _This example shows how a value other than instance, batch size or concurrency can be swept, and will sweep through eight configs:<br>_
 
-- \[2 _ `max_batch_size`] \* \[2 _ `max_queue_delay_microseconds`] \* \[2 \* `instance_group`].
+- \[2 \* `max_batch_size`] \* \[2 \* `max_queue_delay_microseconds`] \* \[2 \* `instance_group`].
 
 ```yaml
 model_repository: /path/to/model/repository/
--- a/docs/config_search.md
+++ b/docs/config_search.md
@@ -18,84 +18,132 @@ limitations under the License.
 
 Model Analyzer's `profile` subcommand supports multiple modes when searching to find the best model configuration.
 
-- [Brute](config_search.md#brute-search-mode) is the default when profiling non-ensemble models or when profling mulitple models sequentially, and will do a brute-force sweep of the cross product of all possible configurations
-- [Quick](config_search.md#quick-search-mode) is the default for ensemble models or when profiling multiple models concurrenty, and will use heuristics to try to find the optimal configuration much quicker than brute, and can be enabled via `--run-config-search-mode quick`
-- [Multi-model](config_search.md#multi-model-search-mode) will profile mutliple models to find the optimal configurations for all models while they are running concurrently. This feature is enabled via `--run-config-profile-models-concurrently-enable`
+- [Brute Force Search](config_search.md#brute-search-mode)
+  - `Search type:` Brute-force sweep of the cross product of all possible configurations
+  - `Default for:`
+    - Single non-ensemble models
+    - Multiple models being profiled sequentially
+  - `Command:` `--run-config-search-mode brute`
+- [Quick Search](config_search.md#quick-search-mode)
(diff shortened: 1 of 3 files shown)


docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
...
+  - `Search type:` Brute-force sweep of the cross product of all possible configurations
+  - `Default for:`
+    - Single non-ensemble models
+    - Multiple models being profiled sequentially
+  - `Command:` `--run-config-search-mode brute`
+- [Quick Search](config_search.md#quick-search-mode)
+  - `Search type:` Heuristic sweep using a hill-climbing algorithm to find an optimal configuration
+  - `Default for:`
+    - Single ensemble models
+    - Multiple models beging profiled concurrently
```
dyastremsky, line 31: Nit: "being" (additional letter)

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
...
+  - `Command:` `--run-config-search-mode quick`
+- [Concurrent / Multi-model Search](config_search.md#multi-model-search-mode)
```
dyastremsky, line 33: I think having this section here is a bit confusing. It breaks the paradigm of brute force vs quick search, since concurrent search isn't an alternative to brute or quick (it's in a different category). While I like the layout of this, I think to minimize confusion, it might make sense to move this information, like to the sections you have below.

If you want to do a summary table like this at the top, I think a separate table with single-model non-ensemble, multi-model non-ensemble (sequential), multi-model non-ensemble (concurrent), and single-model ensemble could be useful and clear.

dyastremsky (commented):
Beautiful work, Brian! I really like these changes. Left a couple of small comments.

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
 
 Model Analyzer's `profile` subcommand supports multiple modes when searching to find the best model configuration.
```
tgerdesnv, line 19: Should we have a table of contents at the top here? It feels like there are enough sections and subsections to justify it.

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
...
-In the default mode, automatic config search will try values 1 through 5 for
-[`instance_group`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#instance-groups).
-The maximum value can be changed using the `run_config_search_max_instance_count` key in the Model Analyzer Config.
-For each `instance_group`, Model Analyzer will sweep values 1 through 128 increasing exponentially (i.e. 1, 2, 4, 8, ...) for
-[`max_batch_size`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#maximum-batch-size). The start and end values can be changed using `run_config_search_min_model_batch_size` and `run_config_search_max_model_batch_size`.
-[`Dynamic_batching`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#dynamic-batcher)
-will be enabled for all model configs generated using automatic search.
+You can also modify the minimum/maximum values that the automatic search space will iterate through:
+
+---
```
tgerdesnv, line 76: again, the `through:` implies there will be something after it, but then the thick line implies that what is below is not related to this.

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
...
+
+### [Instance Group Search Space](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#instance-groups)
+
+- `Default:` 1 to 5 instance groups
+- `--run-config-search-min-instance-count: <val>`: Changes the instance count's minimum automatic search space value
+- `--run-config-search-max-instance-count: <val>`: Changes the instance count's maximum automatic search space value
+
+---
+
+### [Model Batch Size Search Space](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#maximum-batch-size)
```
tgerdesnv, line 86: I can't tell if these links are valid. When I click on it, it doesn't jump to the correct subsection. Then again it might just be an artifact of viewing it in the side branch

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
...
+## Manual Brute Search
+
+**Default brute search mode when any model config parameters or parameters are specified**
+
+Using Manual Brute Search, you can create custom sweeps for any parameter that can be specified in the model configuration. There are two ways this mode is enabled when doing a brute search:
+
+- Any `model config parameters` are specified:
+  - You must manually specify the batch sizes and instance counts you want MA to sweep
+  - Request concurrencies will still be automatically swept (as they are not a model config parameter)
+- Any `parameters` are specified:
```
tgerdesnv, line 136: What does parameters mean in this case?

docs/config_search.md
```diff
@@ -108,29 +156,20 @@ profile_models:
...
+      concurrency: 1,2,3,128
+```
 
-You can also specify `concurrency` ranges to sweep through. If unspecified, it will
-automatically sweep concurrency for every model configuration (unless `--run-config-search-disable`
-is set, in which case it will only use the concurrency value of 1)
+_This example shows how a value other than instance, batch size or concurrency can be swept, and will sweep through eight configs:<br>_
 
-An example Model Analyzer Config that performs manual sweeping looks like below:
+- \[2 _ `max_batch_size`] \* \[2 _ `max_queue_delay_microseconds`] \* \[2 \* `instance_group`].
```
tgerdesnv, line 172: underscores should be asterisks?  Or shouldn't exist at all?

tgerdesnv (changes requested):

docs/config_search.md
nv-braf: Fixed

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
...
+  - `Search type:` Heuristic concurrent sweep of all models' search spaces to find an optimal configuration for all models
+  - `Default for: `
+    - None
+  - `Command:` `--run-config-profile-models-concurrently-enable`
 
 ## Brute Search Mode
 
-Model Analyzer's brute search mode will do a brute-force sweep of the cross product of all possible configurations. You can [Manually](config_search.md#manual-brute-search) provide `model_config_parameters` to tell Model Analyzer what to sweep over, or you can
-let it [Automatically](config_search.md#automatic-brute-search) sweep through configurations expected to have the highest impact on performance for Triton models.
+---
```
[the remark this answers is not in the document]
nv-braf: Agreed. These didn't show up in my viewer. I've fixed them so that only the thin line shows up.

docs/config_search.md
nv-braf: removed the line

docs/config_search.md
nv-braf: They link correctly for me.

docs/config_search.md
nv-braf: These are two separate sections in the config YAML: `model config parameters` (for instance and batch size) and `parameters` (where concurrency is set).

docs/config_search.md
nv-braf: Forgot the backslash and vscode converted the asterisks to underscores. Fixed

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
...
 
-Automatic configuration search is the default behavior when running Model
-Analyzer without manually specifying what values to search. The parameters
-that are automatically searched are
-[`max_batch_size`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#maximum-batch-size)
-and
-[`instance_group`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#instance-groups).
-Additionally, [`dynamic_batching`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#dynamic-batcher) will be enabled if it is legal to do so.
+- [Automatic](config_search.md#automatic-brute-search)
+  - **No** model config parameters are specified
```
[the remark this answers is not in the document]
nv-braf: Yes, I believe these statements are accurate. Note there is a difference btw. `model config parameters` (controls instance/batch size) and `parameters` (controls concurrency) which is definitely a source of confusion.

docs/config_search.md
nv-braf: Yeah, I'm not a fan of this either, but wanted to see what you two thought before I made major changes here.
My new plan is to have two sections: one for search types (quick vs. brute) and another for model types (single, sequential, concurrent, ensemble).

docs/config_search.md
```diff
@@ -18,84 +18,132 @@ limitations under the License.
 
 Model Analyzer's `profile` subcommand supports multiple modes when searching to find the best model configuration.
 
-- [Brute](config_search.md#brute-search-mode) is the default when profiling non-ensemble models or when profling mulitple models sequentially, and will do a brute-force sweep of the cross product of all possible configurations
-- [Quick](config_search.md#quick-search-mode) is the default for ensemble models or when profiling multiple models concurrenty, and will use heuristics to try to find the optimal configuration much quicker than brute, and can be enabled via `--run-config-search-mode quick`
-- [Multi-model](config_search.md#multi-model-search-mode) will profile mutliple models to find the optimal configurations for all models while they are running concurrently. This feature is enabled via `--run-config-profile-models-concurrently-enable`
+- [Brute Force Search](config_search.md#brute-search-mode)
+  - `Search type:` Brute-force sweep of the cross product of all possible configurations
```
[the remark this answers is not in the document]
nv-braf: I don't see back ticks. This is how it looks for me:
image

This looks different in the vscode viewer (less busy). I will change some of this to be bold instead.

docs/config_search.md
nv-braf: I've added a ToC

docs/config_search.md
```diff
@@ -147,57 +210,77 @@ profile_models:
...
+  - model_A
+```
+
+---
+
+### **Limiting Batch Size, Instance Count, and Client Concurrency**
+
+Using the `--run-config-search-<min/max>...` config options you have the ability to clamp the algorithm's upper or lower bounds for the model's batch size and instance count, as well as the client's request concurrency.
+
+_Note: That by default quick search runs unbounded and ignores any default values for these settings_
```
dyastremsky, line 254: Could you change this to be a bit more readable? E.g. "Note: By default, quick search runs unbounded and ignores any default values for these settings". I know this verbiage existed before this PR but would be good to fix it going forward.

Otherwise, looks good to me.

dyastremsky (approved):

docs/config_search.md
nv-braf: Done

dyastremsky (dismissed):

docs/config_search.md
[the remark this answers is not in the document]
tgerdesnv: I'm not saying the back ticks render, but they result in highlighting that we generally only use to highlight code or CLI options.  The changes you made look good 👍 

image

docs/config_search.md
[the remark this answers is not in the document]
tgerdesnv: ok, manual vs automatic is just referring to model config parameters? That makes sense to me, because we aren't really "searching" for an optimal concurrency.

docs/config_search.md
```diff
@@ -14,88 +14,160 @@ See the License for the specific language governing permissions and
...
-[`instance_group`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#instance-groups).
-The maximum value can be changed using the `run_config_search_max_instance_count` key in the Model Analyzer Config.
-For each `instance_group`, Model Analyzer will sweep values 1 through 128 increasing exponentially (i.e. 1, 2, 4, 8, ...) for
-[`max_batch_size`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#maximum-batch-size). The start and end values can be changed using `run_config_search_min_model_batch_size` and `run_config_search_max_model_batch_size`.
-[`Dynamic_batching`](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#dynamic-batcher)
-will be enabled for all model configs generated using automatic search.
+You can also modify the minimum/maximum values that the automatic search space will iterate through:
+
+### [Instance Group Search Space](https://github.com/triton-inference-server/server/blob/master/docs/user_guide/model_configuration.md#instance-groups)
```
tgerdesnv, line 102: do we want to be saying instance group here? Or instance count? I think the triton config option is called instance group, but not sure if that is clear to the user in this context. Note that our option name is instance-count. I think we should be consistent and say group or count everywhere.

docs/config_search.md
tgerdesnv: Yeah I just don't think this is clear for the user. Do we have other doc sections that can be linked to give the user context of what those are and how they are specified?

docs/config_search.md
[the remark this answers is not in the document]
tgerdesnv: Wait, i'm still confused. In the Manual Brute Search section it says if any 'parameters' are specified that you must specify batch sizes/instance counts/concurrencies you want to sweep. Does that mean that specifying parameters puts us in manual mode?  Does the user have to specify those options in `model config parameters`?

docs/config_search.md
```diff
@@ -14,88 +14,160 @@ See the License for the specific language governing permissions and
...
+
+Model Analyzer's `profile` subcommand supports multiple modes when searching to find the best model configuration:
+
+- [Brute Force Search](config_search.md#brute-search-mode)
+  - **Search type:** Brute-force sweep of the cross product of all possible configurations
+  - **Default for:**
+    - Single non-ensemble models
+    - Multiple models being profiled sequentially
+  - **Command:** `--run-config-search-mode brute`
+- [Quick Search](config_search.md#quick-search-mode)
```
tgerdesnv, line 42: Is quick our default yet for normal runs? I'm guessing no since the docs you just added say no. Have we considered changing that?

The reason I bring it up is I'm wondering if quick should come before brute in the docs.

docs/config_search.md
```diff
@@ -147,57 +210,77 @@ profile_models:
...
+Ensemble models can be optimized using the Quick Search mode's hill climbing algorithm to search the ensemble sub-model's configuration spaces in parallel, looking for the maximal objective value within the specified constraints. Model Analyzer has observed positive outcomes towards finding the maximum objective value; with runtimes under one hour (compared to the days it would take a brute force run to complete) for ensembles with up to four submodels.
 
-After it has found the best config(s), it will then sweep the top-N configurations found (specified by `--num-configs-per-model`) over the concurrency range before generation of the summary reports.
+After MA has found the best config(s), it will then sweep the top-N configurations found (specified by `--num-configs-per-model`) over the concurrency range before generation of the summary reports.
 
 ## Multi-Model Search Mode
```
tgerdesnv, line 285: The other sections have a thick line before them. Should one be here as well?

docs/config_search.md
nv-braf: We are split in our documentation about 50/50 between using groups and counts. I think counts makes more sense in the context of MA, but Triton calls these groups.

I'm fine either way, but my preference would be to change group -> count.

docs/config_search.md
nv-braf: It is not, as per a previous discussion, we were worried about backwards computability and breaking other teams scripts that rely on brute force MA by default.

docs/config_search.md
nv-braf: I don't think it's very clear either. You really need to be familiar in the art of writing a protobuf to understand the difference. We have extensive examples of `model config parameters` and `parameters` in `config.md`.

I think it's beyond the scope of our document to explain this to the user - my feeling is that they're either familiar with config files and they are going to understand this, or they aren't and will not care about this section as they won't be trying to run this niche case. 

Personally, I hate that we have this schism between automatic and manual, but this is how our logic works and I thought it was important to call it out to the user.

docs/config_search.md
nv-braf: Yes. Fixed

docs/config_search.md
tgerdesnv: Yeah but can't we link the existing docs for those parameter sections? For example, make `parameters` a clickable link to https://github.com/triton-inference-server/model_analyzer/blob/main/docs/config.md#test-configuration-parameter ?

docs/config_search.md
tgerdesnv: I'm fine either way, but lets be consistent. You can go ahead and change group to count

tgerdesnv:
Looks like you'll need to rebase.

docs/config_search.md
nv-braf: I actually don't think we can be consistent here. In the YAML it's called `instance-group`, but in our CLI it's called instance count.

We also refer to this as instance group in our tables. 

I now think the best course of action is to change count -> group and consider a story to change the CLI option to use group instead of count. Thoughts?

docs/config_search.md
nv-braf: Done

tgerdesnv (dismissed):

dyastremsky (dismissed):

tgerdesnv (approved):

PR #618, opened 2023-02-28
Update quick_start.md

knwng:
Add the missing backslash


nv-braf:
Thanks for catching this!

nv-braf (approved):

PR #619, opened 2023-02-28
Multi-model detailed reporting changes

nv-braf:
- Remove message in profile step that instructs user about running detail reporting
- Add error message in detailed report step when multiple different models are detected

Needed to modify **L0_results** test to only check for a single model config

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -69,9 +69,10 @@ class Analyzer:
         state_manager.load_checkpoint(checkpoint_required)
 
         self._constraint_manager = ConstraintManager(self._config)
-        self._result_manager = ResultManager(config=config,
-                                             state_manager=self._state_manager,
-                                             constraint_manager=self._constraint_manager)
+        self._result_manager = ResultManager(
+            config=config,
+            state_manager=self._state_manager,
+            constraint_manager=self._constraint_manager)
 
     def profile(self, client: TritonClient, gpus: List[GPUDevice], mode: str,
                 verbose: bool) -> None:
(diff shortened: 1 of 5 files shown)


qa/L0_results/test.sh
```diff
@@ -23,7 +23,7 @@ MODEL_ANALYZER="`which model-analyzer`"
 REPO_VERSION=${NVIDIA_TRITON_SERVER_VERSION}
 MODEL_REPOSITORY=${MODEL_REPOSITORY:="/mnt/nvdl/datasets/inferenceserver/$REPO_VERSION/libtorch_model_store"}
 CHECKPOINT_REPOSITORY=${CHECKPOINT_REPOSITORY:="/mnt/nvdl/datasets/inferenceserver/model_analyzer_checkpoints/2022_08_02"}
-QA_MODELS="vgg19_libtorch resnet50_libtorch"
+QA_MODELS="resnet50_libtorch"
```
tgerdesnv, line 26: what happens before this change? Does the test fail? Why?
We should double check with Jakub that this change doesn't break Model Navigator. I'm pretty sure they never run the report step, but we should double check.

model_analyzer/analyzer.py
```diff
@@ -123,7 +126,10 @@ def profile(self, client: TritonClient, gpus: List[GPUDevice], mode: str,
         if not self._config.skip_summary_reports:
             self._create_summary_tables(verbose)
             self._create_summary_reports(mode)
-            logger.info(self._get_report_command_help_string())
+
+            # TODO-TMA-650: Detailed reporting not supported for multi-model
```
tgerdesnv, line 130: If someone runs A and B non-concurrently, do we suggest a report command that is now illegal?

nv-braf: This does not (and has never) worked correctly for sequential profiling. Please review the code changes for this. The new output is:
```
16:18:57 [Model Analyzer] To generate detailed reports for the 3 best simple_int8 configurations, run `model-analyzer report --report-model-configs simple_int8_config_7,simple_int8_config_9,simple_int8_config_4 --export-path /swdev/model_analyzer --checkpoint-directory /swdev/profile_checkpoints/`
16:18:58 [Model Analyzer] To generate detailed reports for the 3 best add_sub configurations, run `model-analyzer report --report-model-configs add_sub_config_4,add_sub_config_6,add_sub_config_7 --export-path /swdev/model_analyzer --checkpoint-directory /swdev/profile_checkpoints/`
```

PR #620, opened 2023-02-28
Fix wrong link to detailed report

Alwahsh:
Minor typo but the link of the detailed report was opening the summary.


nv-braf:
Thanks for pointing this out. I've fixed the link.

PR #626, opened 2023-03-03
model_analyzer/config/input/config_defaults: Solving: Not Found manif…

NikeNano:
…est for nvcr.io/nvidia/tritonserver:23.02-py3 not found: manifest unknown: manifest unknown by updating image ref

The current default image: `nvcr.io/nvidia/tritonserver:23.02-py3` does not exist. Update image.


NikeNano:
@nv-braf  could you take a look? Thanks

mc-nv:
The image is in NGC registry: 
image
I also was able to pull the image. 

Suppose we should close this issue.

NikeNano:
Ohh seems like the image was create on the 6 but I make the PR on the 3 of March so might have been created since. No need for this PR then so will close it. Thanks @mc-nv and @nv-braf !

PR #627, opened 2023-03-07
Add caveat to ensemble documentation

nv-braf:
Adds caveat that we don't support submodel `cpu_only` for ensembles
Fixes typos pointed out by external user

### DIFF
--- a/docs/config_search.md
+++ b/docs/config_search.md
@@ -277,6 +277,7 @@ _This mode has the following limitations:_
 
 - Can only be run in `quick` search mode
 - Only supports up to 4 sub-models
+- Does not support `cpu_only` option for submodels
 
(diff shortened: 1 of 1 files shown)


docs/config.md
```diff
@@ -608,10 +608,10 @@ profile_models:
       dynamic_batching:
         max_queue_delay_microseconds: [200, 400, 600]
       instance_group:
-        - - kind: KIND_GPU
-            count: 1
-        - - kind: KIND_CPU
-            count: 1
+        - kind: KIND_GPU
```
tgerdesnv, line 611: Are you sure these changes are correct?

tgerdesnv: I think with single dash it would be [GPU, CPU], which would tell MA to put one instance on a GPU and one instance on a CPU. 
I think with double dash it would be [GPU],[CPU], which would tell MA to try GPU and then try CPU. Double dash might be correct here.

tgerdesnv: We should test to confirm, and make sure it is correct/consistent in this file

tgerdesnv (approved):

PR #628, opened 2023-03-08
Add --bls-composing-models config option

nv-braf:
Limited unit testing for this, but I have a BLS model (pipeline) with two submodels (vae, text encoder) and have verified that the command works as intended.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -89,8 +89,8 @@ model_repository: <string>
 # List of the model names to be profiled
 profile_models: <comma-delimited-string-list>
 
-# List of submodels for BLS models
-bls_submodels: <comma-delimited-string-list>
+# List of composing models for BLS models
+bls_composing_models: <comma-delimited-string-list>
 
 # Full path to directory to which to read and write checkpoints and profile data
 [ checkpoint_directory: <string> | default: './checkpoints' ]
@@ -255,8 +255,8 @@ The following config options are supported **only by the YAML** config file.
 # YAML config section for each model to be profiled
 profile_models: <comma-delimited-string-list|list|profile_model>
 
-# List of submodels for BLS models
-bls_submodels: <comma-delimited-string-list>
+# List of composing models for BLS models
+bls_composing_models: <comma-delimited-string-list>
 
 # List of constraints placed on the config search results
 [ constraints: <constraint> ]
--- a/docs/config.md
+++ b/docs/config.md
@@ -89,6 +89,9 @@ model_repository: <string>
 # List of the model names to be profiled
 profile_models: <comma-delimited-string-list>
 
+# List of submodels for BLS models
+bls_submodels: <comma-delimited-string-list>
+
 # Full path to directory to which to read and write checkpoints and profile data
 [ checkpoint_directory: <string> | default: './checkpoints' ]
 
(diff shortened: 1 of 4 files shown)


tests/test_config.py
```diff
@@ -1918,6 +1918,19 @@ def _test_quick_search_with_rcs(self,
         with self.assertRaises(TritonModelAnalyzerException):
             self._evaluate_config(new_args, yaml_content, subcommand='profile')
 
+    def test_bls_submodels(self):
+        """
+        Test that BLS submodels can be specified only per model in the YAML
```
tgerdesnv, line 1923: This comment doesn't seem to line up with what is happening.

nv-braf: Fixed

tests/test_config.py
```diff
@@ -1918,6 +1918,19 @@ def _test_quick_search_with_rcs(self,
         with self.assertRaises(TritonModelAnalyzerException):
             self._evaluate_config(new_args, yaml_content, subcommand='profile')
 
+    def test_bls_submodels(self):
+        """
+        Test that BLS submodels can be specified
```
tgerdesnv, line 1923: so this just makes sure there is no assert? Should there be a test using yaml since that is legal? Should there be a check that the submodels are supplied correctly in the resulting config?

nv-braf: Yes, makes sure that the new option is legal and understood. 
Maybe, but we haven't done this extensive testing in the past. 
I thought about adding this, but I don't think it's needed here, since we are just reusing the same config primitive string that's already being tested.

nv-braf: I've added a YAML check as well.

docs/config.md
```diff
@@ -252,6 +255,9 @@ The following config options are supported **only by the YAML** config file.
 # YAML config section for each model to be profiled
 profile_models: <comma-delimited-string-list|list|profile_model>
 
+# List of submodels for BLS models
+bls_submodels: <comma-delimited-string-list>
```
Tabrizian, line 259: Submodels is not a terminology that triton uses. Perhaps rename it to `--bls-composing-models`?

nv-braf: Thanks for the insight. I've changed to your suggestion.

Tabrizian (approved):

docs/config.md
tgerdesnv: Is submodel a valid term for ensembles? If not, would composing apply there as well?

nv-braf: I've changed this already

tests/test_config.py
```diff
@@ -1918,6 +1918,30 @@ def _test_quick_search_with_rcs(self,
         with self.assertRaises(TritonModelAnalyzerException):
             self._evaluate_config(new_args, yaml_content, subcommand='profile')
 
+    def test_bls_composing_models(self):
+        """
+        Test that BLS submodels can be specified
```
tgerdesnv, line 1923: "composing"

tgerdesnv (approved):

docs/config.md
Tabrizian: Yes, we use composing for both ensemble and BLS.

PR #629, opened 2023-03-08
Replace submodel references with composing models

nv-braf:


tgerdesnv (approved):
looks like it was in a lot more forward facing places than we thought! Looks good as long as it passed a CI

nv-braf:
Yeah, but it didn't take too long. CI just passed.

PR #637, opened 2023-03-14
Logic to create BLS composing models for default config

nv-braf:
Creates BLS composing models for the default config. Limited unit testing here, but more will be added once the non-default configs are generated.

In addition I have run this with a BLS model and confirmed that we are sending/load configs to PA/tritonserver correctly (but, I needed to hack the BLS composing configs name - which is the follow-on story to this)

### DIFF
--- a/model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
+++ b/model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
@@ -51,6 +51,7 @@ class QuickPlusConcurrencySweepRunConfigGenerator(ConfigGeneratorInterface):
                  config: ConfigCommandProfile, gpus: List[GPUDevice],
                  models: List[ModelProfileSpec],
                  ensemble_composing_models: Dict[str, List[ModelProfileSpec]],
+                 bls_composing_models: List[ModelProfileSpec],
                  client: TritonClient, result_manager: ResultManager,
                  model_variant_name_manager: ModelVariantNameManager):
         """
@@ -65,6 +66,8 @@ class QuickPlusConcurrencySweepRunConfigGenerator(ConfigGeneratorInterface):
             List of models to profile
         ensemble_composing_models: Dict of List of ModelProfileSpec
             Dict indexed by model name of list of composing models to profile
+        bls_composing_models: List of ModelProfileSpec
+            List of BLS composing models to profile
         client: TritonClient
         result_manager: ResultManager
             The object that handles storing and sorting the results from the perf analyzer
@@ -78,6 +81,7 @@ class QuickPlusConcurrencySweepRunConfigGenerator(ConfigGeneratorInterface):
         self._gpus = gpus
         self._models = models
         self._ensemble_composing_models = ensemble_composing_models
+        self._bls_composing_models = bls_composing_models
         self._client = client
         self._result_manager = result_manager
         self._model_variant_name_manager = model_variant_name_manager
@@ -122,6 +126,7 @@ class QuickPlusConcurrencySweepRunConfigGenerator(ConfigGeneratorInterface):
             gpus=self._gpus,
             models=self._models,
             ensemble_composing_models=self._ensemble_composing_models,
+            bls_composing_models=self._bls_composing_models,
             client=self._client,
             model_variant_name_manager=self._model_variant_name_manager)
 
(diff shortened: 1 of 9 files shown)


model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -65,6 +66,8 @@ def __init__(self, search_config: SearchConfig,
             List of models to profile
         ensemble_composing_models: Dict of List of ModelProfileSpec
             Dict indexed by model name of list of composing models to profile
+        bls_composing_models: List of ModelProfileSpec
```
tgerdesnv, line 69: Why is the format for bls vs ensemble so different?

model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -537,6 +550,13 @@ def _create_default_model_run_config(
                                                   default_model_config,
                                                   default_perf_analyzer_config)
 
+        default_composing_model_configs = self._create_default_composing_model_configs(
+            model)
+
+        if default_composing_model_configs:
+            default_model_run_config.add_bls_composing_model_configs(
```
tgerdesnv, line 557: this function is BLS-specific, but the code for the function above implies it can be ensemble or bls?

model_analyzer/config/generate/run_config_generator_factory.py
```diff
@@ -73,12 +73,17 @@ def create_run_config_generator(
...
+            command_config, client, gpus)
+
+        if (command_config.run_config_search_mode == "quick" or
+                bls_composing_models):
             return RunConfigGeneratorFactory._create_quick_plus_concurrency_sweep_run_config_generator(
                 command_config=command_config,
                 gpus=gpus,
                 models=new_models,
                 ensemble_composing_models=ensemble_composing_models,
+                bls_composing_models=bls_composing_models,
```
tgerdesnv, line 86: I'm seeing ensemble and bls passed in separately everywhere. Are they different enough that you have to handle them separately?

model_analyzer/config/generate/run_config_generator_factory.py
```diff
@@ -112,26 +117,28 @@ def _create_quick_plus_concurrency_sweep_run_config_generator(
...
             result_manager=result_manager,
             model_variant_name_manager=model_variant_name_manager)
 
     @staticmethod
     def _create_search_config(
-        models: List[ModelProfileSpec],
-        ensemble_composing_models: Dict[str, List[ModelProfileSpec]]
-    ) -> SearchConfig:
+            models: List[ModelProfileSpec],
+            ensemble_composing_models: Dict[str, List[ModelProfileSpec]],
```
tgerdesnv, line 140: This feels like it is getting less and less clean. We are passing in a bunch of optional dicts/lists, and the contents of them imply whether or not the feature is enabled? It seems that would be really confusing for someone who doesn't have intimate knowledge of the code base.

Are there ways to refactor or clean all of this up? Can there be a thin layer of prep for normal vs multi vs ensemble vs bls, and have all of the underlying code be shared? I think that is sort of what is already happening, but just want to take a step back and think if there are better ways to organize this code. It's hard for me to make those calls just looking at the PR diffs.

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: Because you can have multiple models, one (or more) of which is an ensemble. But, it doesn't work that way for BLS since we need to specify composing models using the config, so as of now we'll only support a single BLS when running MA.

model_analyzer/config/generate/run_config_generator_factory.py
nv-braf: At this point they are - I would like to combine them, but my goal is to get the flow working first and then go back and make improvements if needed. (See the above comment (list vs. dict of list) as to why they are different.

model_analyzer/config/generate/quick_run_config_generator.py
nv-braf: The function above is a generic composing config creator, but this method cannot be reached for an ensemble config (it has a different default config generator method).

tgerdesnv (approved):

PR #639, opened 2023-03-16
Create unchanged BLS composing model output directory

nv-braf:
BLS composing configs cannot mutate their names. This change creates a directory using the original composing model name and changes the config.pbtxt to use this as the model name when loading on the tritonserver.

### DIFF
--- a/model_analyzer/config/generate/base_model_config_generator.py
+++ b/model_analyzer/config/generate/base_model_config_generator.py
@@ -288,7 +288,7 @@ class BaseModelConfigGenerator(ConfigGeneratorInterface):
     def create_original_config_from_variant(
             variant_config: ModelConfig) -> ModelConfig:
         """
-        Removes '_config_$/default' from the variant config and returns
+        Removes 'config_#/default' from the variant config and returns
         a new model config
         """
         original_config = deepcopy(variant_config)
--- a/model_analyzer/config/generate/base_model_config_generator.py
+++ b/model_analyzer/config/generate/base_model_config_generator.py
@@ -22,6 +22,7 @@ from typing import List, Optional, Generator, Dict, Any
 from model_analyzer.constants import LOGGER_NAME
 from model_analyzer.triton.model.model_config import ModelConfig
 from .model_profile_spec import ModelProfileSpec
+from copy import deepcopy
 import abc
 import logging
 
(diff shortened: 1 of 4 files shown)


model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -283,6 +284,21 @@ def extract_model_name_from_variant_name(variant_name: str) -> str:
         """
         return variant_name[:variant_name.find("_config_")]
 
+    @staticmethod
+    def create_original_config_from_variant(
+            variant_config: ModelConfig) -> ModelConfig:
+        """
+        Removes '_config_$/default' from the variant config and returns
```
tgerdesnv, line 291: $?

model_analyzer/record/metrics_manager.py
```diff
@@ -292,6 +292,11 @@ def _create_model_variants(self, run_config):
 
                 self._create_model_variant(original_name, bls_composing_config)
 
+                bls_original_composing_config = BaseModelConfigGenerator.create_original_config_from_variant(
```
tgerdesnv, line 295: Is this needed? Or do you have access to the original anywhere here in this function? It seems like we are creating variants from original and then original from variant, but the function names may just be misleading

model_analyzer/triton/model/model_config.py
```diff
@@ -347,6 +347,11 @@ def set_composing_model_variant_name(self, composing_model_name: str,
 
         self._model_config = self.from_dict(model_config_dict)._model_config
 
+    def set_model_name(self, model_name: str) -> None:
```
tgerdesnv, line 350: is there no easier way than a full from/to dict?

model_analyzer/record/metrics_manager.py
nv-braf: The originally named config is just the default, which is not what we want. Need to create an originally (unmodified) named model config, that has been modified with whatever the RCG has chosen. I'm open to a better name suggestion for this method.

model_analyzer/triton/model/model_config.py
nv-braf: There is, but it would require string parsing with search and replacement which seems unsafe/prone to error.

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: Typo. Fixed.

model_analyzer/record/metrics_manager.py
tgerdesnv: ah, so 'original' in one case is the base config, and 'original' here is the variant config with the base name. Confusing. Yes, one of the two mentions of 'original' should be updated.

nv-braf: No, in both cases we are taking in the variant config.

tgerdesnv (approved):

PR #641, opened 2023-03-21
Improve speed of neighborhood creation

tgerdesnv:
Speed up neighborhood creation by building it up instead of starting with a huge square/cube and reducing it

### DIFF
--- a/qa/L0_no_config_file/test.sh
+++ b/qa/L0_no_config_file/test.sh
@@ -29,6 +29,8 @@
 
 source ../common/util.sh
 
+REPO_VERSION=${NVIDIA_TRITON_SERVER_VERSION}
+
 rm -f *.log
 OUTPUT_MODEL_REPOSITORY=${OUTPUT_MODEL_REPOSITORY:=`get_output_directory`}
 rm -rf $OUTPUT_MODEL_REPOSITORY
@@ -42,7 +44,6 @@ rm $MODEL_REPOSITORY/*/config.pbtxt
 
 # Set test parameters
 MODEL_ANALYZER="`which model-analyzer`"
-REPO_VERSION=${NVIDIA_TRITON_SERVER_VERSION}
 TRITON_DOCKER_IMAGE=${TRITONSERVER_BASE_IMAGE_NAME}
 TRITON_LAUNCH_MODES="local docker"
 CLIENT_PROTOCOLS="http grpc"


model_analyzer/config/generate/coordinate.py
```diff
@@ -74,6 +74,12 @@ def __eq__(self, other: Any) -> bool:
                 return False
         return True
 
+    def __lt__(self, other: Any) -> bool:
+        for i, v in enumerate(self._values):
+            if v != other[i]:
```
nv-braf, line 79: I don't understand what the _lt__ is trying to evaluate? I would have thought it would be that each entry in other is less than it's counterpart in self.  So, why do you need to check that the values are not equal first? 

Should this just be:
```
for i, v in enumerate(self._values):
    if v > other[i]
        return False
return True

File: model_analyzer/config/generate/coordinate.py [hunk 2]
```diff
@@ -74,6 +74,12 @@ def __eq__(self, other: Any) -> bool:
                 return False
         return True
 
+    def __lt__(self, other: Any) -> bool:
```
nv-braf, line 77: Doesn't other have to be a List[int] here (and same for the above dunder methods)?

model_analyzer/config/generate/neighborhood.py
```diff
@@ -214,48 +214,82 @@ def get_nearest_neighbor(self, coordinate_in: Coordinate) -> Coordinate:
...
     def _create_neighborhood(self) -> List[Coordinate]:
+        """
+        Create and return a neighborhood of all Coordinates within 
+        range <_radius> that are also within all bounds
+        """
 
-        neighborhood = []
-        potential_neighborhood = self._get_potential_neighborhood(
-            self._home_coordinate, self._radius)
+        def _is_in_bounds(potential_coordinate: List[int]) -> bool:
```
nv-braf, line 222: What's the benefit of embedding this method inside of _create_neighborhood()?

model_analyzer/config/generate/neighborhood.py
```diff
@@ -214,48 +214,82 @@ def get_nearest_neighbor(self, coordinate_in: Coordinate) -> Coordinate:
...
+        Create and return a list of all possible step vectors that are 
+        within <_radius> distance
+        """
 
-        return neighborhood
+        result_list = []
+        v = [0] * num_coordinates
+        base = [0] * num_coordinates
+
+        def _append_combinations_to_results(curr_val: List[int],
```
nv-braf, line 251: In my opinion this is less clean/readable. Took me a couple of minutes to figure out that this method is really just 247-249, and 291-292

model_analyzer/config/generate/coordinate.py
tgerdesnv: I just need a way to compare coordinates so they can be sorted in a deterministic way. My method treats them like a number. 0,9,9,9 is less than 1,0,0,0 (999 is less than 1000). Your proposal is not deterministic. [1,0] compared with [0,1] would return false regardless of the order I put them in.

That all being said, this shouldn't be added to the real class just to be used in a test. Let me try to move the logic into a test file.

tgerdesnv: this code is gone (and simpler code is now in the test file)

model_analyzer/config/generate/coordinate.py
tgerdesnv: this code is gone

model_analyzer/config/generate/neighborhood.py
tgerdesnv: it also gives access to all of the variables in the parent function so that they don't all have to be explicitly passed into all of the child functions

PR #643, opened 2023-03-21
Logic for non-default BLS profiling

nv-braf:
Includes unit testing and I've run this on FaceDetect and confirmed it is working.

### DIFF
--- a/model_analyzer/config/generate/quick_run_config_generator.py
+++ b/model_analyzer/config/generate/quick_run_config_generator.py
@@ -321,6 +321,17 @@ class QuickRunConfigGenerator(ConfigGeneratorInterface):
         model_run_config = ModelRunConfig(model_name, model_config,
                                           perf_analyzer_config)
 
+        bls_composing_model_configs = []
+        for bls_composing_model in self._bls_composing_models:
+            bls_composing_model_config = self._get_next_model_config(
+                bls_composing_model, model_index)
+            model_index += 1
+            bls_composing_model_configs.append(bls_composing_model_config)
+
+        if self._bls_composing_models:
+            model_run_config.add_bls_composing_model_configs(
+                bls_composing_model_configs)
+
         if model.model_name() in self._ensemble_composing_models:
             model_run_config.add_ensemble_composing_model_configs(
                 ensemble_composing_configs)
--- a/tests/test_quick_run_config_generator.py
+++ b/tests/test_quick_run_config_generator.py
@@ -29,7 +29,7 @@ from model_analyzer.config.generate.model_profile_spec import ModelProfileSpec
 from tests.common.test_utils import evaluate_mock_config
 
 
-def mock_ensemble_configs(*args, **kwargs):
+def mock_configs(*args, **kwargs):
     fake_config = {
         "name": "my-model",
         "platform": "ensemble",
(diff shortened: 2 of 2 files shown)


model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -321,6 +321,17 @@ def _get_next_model_run_config(
         model_run_config = ModelRunConfig(model_name, model_config,
                                           perf_analyzer_config)
 
+        bls_composing_model_configs = []
+        for bls_composing_model in self._bls_composing_models:
+            bls_composing_model_config = self._get_next_model_config(
+                bls_composing_model, model_index)
+            model_index += 1
```
tgerdesnv, line 328: This model index code seems very fragile. A lot of baked in knowledge to know the implicit locations of the models. Let's make sure that is part of the refactor.

tests/test_quick_run_config_generator.py
```diff
@@ -911,6 +930,138 @@ def _get_next_run_config_ensemble(self,
 
         self.assertEqual(perf_config['batch-size'], 1)
 
+    def _get_next_run_config_bls(self, max_concurrency=0, min_concurrency=0):
+        """
+        Test that get_next_run_config() creates a proper RunConfig for BLS
+
+        Sets up a case where the coordinate is [1,2,3,4,5,6], which corresponds to
```
tgerdesnv, line 937: Is this comment correct? Seems like way more than 6 values below

tests/test_quick_run_config_generator.py
```diff
@@ -911,6 +930,138 @@ def _get_next_run_config_ensemble(self,
...
+          - composing model 1 concurrency = 8*5*2 = 80
+          - composing model 2 max_batch_size = 32
+          - composing model 2 instance_count = 7
+          - composing model 2 concurrency = 32*7*2 = 448
+          - model concurrency = 12 (minimum value of [12, 80, 448])
+
+        Also,
+        - sequence batching should be on for model 1
+        - dynamic batching should be on for model 2
+        - existing values from the base model config should persist if they aren't overwritten
```
tgerdesnv, line 952: Are these two statements tested below?

tests/test_quick_run_config_generator.py
```diff
@@ -911,6 +930,138 @@ def _get_next_run_config_ensemble(self,
 
         self.assertEqual(perf_config['batch-size'], 1)
 
+    def _get_next_run_config_bls(self, max_concurrency=0, min_concurrency=0):
+        """
+        Test that get_next_run_config() creates a proper RunConfig for BLS
```
tgerdesnv, line 935: I love this test if it does everything it says it does! Very thorough

tests/test_quick_run_config_generator.py
nv-braf: Yes, these are the lines that check it:
``` 
self.assertEqual(composing_model_config0.to_dict(),
                         expected_model_config0)
self.assertEqual(composing_model_config1.to_dict(),
                         expected_model_config1)
```

tests/test_quick_run_config_generator.py
nv-braf: It does! And, thanks.

tests/test_quick_run_config_generator.py
nv-braf: Reordered

model_analyzer/config/generate/quick_run_config_generator.py
tgerdesnv: ok, we can revisit it when refactoring

tests/test_quick_run_config_generator.py
nv-braf: Good catch. Base model config check was missing. I've added it. Relevant PA config settings are checked (concurrency/batch size) here:
```
if max_concurrency:
            self.assertEqual(perf_config['concurrency-range'], max_concurrency)
        elif min_concurrency:
            self.assertEqual(perf_config['concurrency-range'], min_concurrency)
        else:
            self.assertEqual(perf_config['concurrency-range'], 12)

        self.assertEqual(perf_config['batch-size'], 1)
```

tgerdesnv (approved):

PR #654, opened 2023-03-30
Remove unused imports

tgerdesnv:
Resolve all "unused import" warnings

### DIFF
--- a/model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
+++ b/model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
@@ -17,7 +17,6 @@ from typing import List, Optional, Generator, Dict
 from .config_generator_interface import ConfigGeneratorInterface
 
 from model_analyzer.config.generate.search_config import SearchConfig
-from model_analyzer.config.generate.brute_run_config_generator import BruteRunConfigGenerator
 from model_analyzer.config.generate.quick_run_config_generator import QuickRunConfigGenerator
 from model_analyzer.config.generate.model_variant_name_manager import ModelVariantNameManager
 from model_analyzer.config.generate.perf_analyzer_config_generator import PerfAnalyzerConfigGenerator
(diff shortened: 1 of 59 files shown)


model_analyzer/result/run_config_result_comparator.py
```diff
@@ -14,10 +14,6 @@
 
 from typing import List, Dict
 
-from model_analyzer.record.record import Record
-
-from model_analyzer.constants import COMPARISON_SCORE_THRESHOLD
```
tgerdesnv, line 19: @nv-braf This went away because we used to only say A was less than B if it was at least 5% less? Now it is strictly less than?

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
```diff
@@ -28,10 +27,8 @@
 from model_analyzer.config.generate.model_profile_spec import ModelProfileSpec
 from model_analyzer.result.result_manager import ResultManager
 from model_analyzer.result.run_config_measurement import RunConfigMeasurement
-from model_analyzer.result.run_config_result import RunConfigResult
 
 from model_analyzer.constants import LOGGER_NAME
-from model_analyzer.config.input.config_defaults import DEFAULT_RUN_CONFIG_MIN_CONCURRENCY, DEFAULT_RUN_CONFIG_MAX_CONCURRENCY
```
tgerdesnv, line 34: @nv-braf This one is sketchy. It appears we hardcode this? Are we correctly honoring the values passed in?

model_analyzer/result/run_config_result_comparator.py
nv-braf: Correct.

model_analyzer/config/generate/quick_plus_concurrency_sweep_run_config_generator.py
nv-braf: We get max concurrency directly from the config (see line 144). These are fine to be removed

tgerdesnv: we don't need to honor a minimum?

tgerdesnv: something like this?
https://github.com/triton-inference-server/model_analyzer/pull/654/commits/3612efa03d194f02abb9ac1492ed716abfc9d425

nv-braf: We do honor the minimum in the quick search.  And, maybe we should for the concurrency sweep? But, I don't see the harm in providing these values to flesh out the detailed report for the user.

nv-braf (approved):

PR #666, opened 2023-04-13
Brute Force Search for Request Rate Range

nv-braf:
I've reworked the concurrency sweep to be a parameter sweep that works for either RRR or concurrency. Added unit testing and have tried it live.


model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -154,87 +152,103 @@ def set_last_results(
...
-            self._concurrency_results.extend(measurement)
+            self._parameter_results.extend(measurement)
 
-    def _create_concurrency_list(self, cli_config: ConfigCommandProfile,
-                                 model_parameters: dict) -> List[int]:
+    def _create_parameter_list(self, model_parameters: dict) -> List[int]:
         if model_parameters['concurrency']:
             return sorted(model_parameters['concurrency'])
-        elif cli_config.run_config_search_disable:
+        elif self._cli_config.request_rate_range:
```
tgerdesnv, line 160: Why do we look at model_parameters for concurrency but cli for rrr? Where does model_parameters come from?

model_analyzer/config/input/config_defaults.py
```diff
@@ -41,8 +41,8 @@
 DEFAULT_CLIENT_PROTOCOL = 'grpc'
 DEFAULT_RUN_CONFIG_MAX_CONCURRENCY = 1024
 DEFAULT_RUN_CONFIG_MIN_CONCURRENCY = 1
-DEFAULT_RUN_CONFIG_MAX_REQUEST_RATE_RANGE = 1024
-DEFAULT_RUN_CONFIG_MIN_REQUEST_RATE_RANGE = 1
+DEFAULT_RUN_CONFIG_MAX_REQUEST_RATE_RANGE = 8192
+DEFAULT_RUN_CONFIG_MIN_REQUEST_RATE_RANGE = 1024
```
tgerdesnv, line 45: why such a high default? Aren't there some slow models that can't hit anywhere near this?

tests/test_cli.py
```diff
@@ -79,8 +79,8 @@ def get_test_options():
         OptionStruct("int", "profile", "--perf-analyzer-max-auto-adjusts", None, "100", "10"),
         OptionStruct("int", "profile", "--run-config-search-min-concurrency", None, "2", "1"),
         OptionStruct("int", "profile", "--run-config-search-max-concurrency", None, "100", "1024"),
-        OptionStruct("int", "profile", "--run-config-search-min-request-rate-range", None, "2", "1"),
-        OptionStruct("int", "profile", "--run-config-search-max-request-rate-range", None, "100", "1024"),
+        OptionStruct("int", "profile", "--run-config-search-min-request-rate-range", None, "2", "1024"),
```
tgerdesnv, line 82: I don't think this should say 'range' (here and many other places).
PA has options for concurrency-range and request-rate-range, but that allows them to set multiple concurrencies or multiple request rates to sweep over.  From our point of view it is just 'concurrency' and 'request-rate'.

tests/test_perf_analyzer_config_generator.py
```diff
@@ -347,6 +347,129 @@ def test_min_concurrency(self):
...
+            """)
+        # yapf: enable
+
+        request_rate_ranges = [1, 2, 3, 4]
+        expected_configs = [
+            construct_perf_analyzer_config(request_rate_range=rrr)
+            for rrr in request_rate_ranges
+        ]
+
+        pa_cli_args = ['-rrr', '1,2,3,4']
```
tgerdesnv, line 374: is `-rrr` an PA thing, and MA thing, or just something made up for testing?

tgerdesnv (changes requested):

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: I've added request rate to model parameters

model_analyzer/config/input/config_defaults.py
nv-braf: Lowered the default

tests/test_cli.py
nv-braf: Removed range

tests/test_perf_analyzer_config_generator.py
nv-braf: Removed `-rrr`

model_analyzer/config/input/config_command_profile.py
```diff
@@ -1134,20 +1136,31 @@ def _autofill_values(self):
...
-                new_model['parameters'] = {
-                    'batch_sizes': self.batch_sizes,
-                    'concurrency': model.parameters()['concurrency']
-                }
-            elif 'concurrency' not in model.parameters():
-                new_model['parameters'] = {
-                    'batch_sizes': model.parameters()['batch_sizes'],
-                    'concurrency': self.concurrency
+                    'concurrency': self.concurrency,
+                    'request_rate': self.request_rate
```
tgerdesnv, line 1140: Do we want both concurrency and request_rate to be specified in this dict? That seems confusing, and prone to error downstream however this is used (unless it is guaranteed that at least one of them is always empty list?)

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -154,87 +152,108 @@ def set_last_results(
             measurement = [max(valid_measurements)]
 
             self._last_results = measurement
-            self._concurrency_results.extend(measurement)
+            self._parameter_results.extend(measurement)
 
-    def _create_concurrency_list(self, cli_config: ConfigCommandProfile,
-                                 model_parameters: dict) -> List[int]:
+    def _create_parameter_list(self, model_parameters: dict) -> List[int]:
```
tgerdesnv, line 157: I'm concerned about this code. This if/else chain and the decision of rr vs conc doesn't seem to be captured in a function or variable, so other parts of the code could make different decisions.  It seems possible that we are in request-rate mode and would return a list of concurrencies, or vice versa?

For example, what if request_rate is specified in the yaml? It seems we would hit the case on line 163 here, but wants_request_rate would still return false? So line 203 would use those values but tag them as concurrency instead? Am I misreading it? Is there unit testing for the yaml specification case?

nv-braf: My `_config_specifies_request_rate()` method handles both CLI and yaml. I also have unit testing that checks for request_rate as a model parameter.

model_analyzer/config/input/config_command_profile.py
tgerdesnv: please add an error check

nv-braf: Done

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: I've refactored this

tgerdesnv: ty that is what i had envisioned 👍

tgerdesnv (approved):

PR #669, opened 2023-04-25
Request Rate in Summary Tables

nv-braf:
Changed summary tables to use request rate, instead of concurrency if we profiled using request rate.
Added unit testing to protect and ran a model (vgg19_libtorch) through profile to ensure it's working correctly.
You can look at `golden-metrics-model-gpu/inference.csv` to see the change.

Note: No changes are needed in the actual summary report PDF as it doesn't use concurrency


model_analyzer/config/input/config_command_profile.py
```diff
@@ -1197,3 +1203,16 @@ def _autofill_values(self):
 
             new_profile_models[model.model_name()] = new_model
         self._fields['profile_models'].set_value(new_profile_models)
+
+    def _using_request_rate(self) -> bool:
```
tgerdesnv, line 1207: didn't we already have similar code in a different story? Any way to reuse (or store the result somewhere)?

nv-braf: Yes, we do. The logic is very similar, but I don't see how it can be reused. It's in `perf_analyzer_config_generator.py:167` in the branch `add-support-for-request-rate-range`

tgerdesnv: this function is in ConfigCommandProfile . A ConfigCommandProfile is passed into PerfAnalyzerConfigGenerator, right? Can PACG leverage this new code in CCP instead of figuring it out on its own?

Also, am I reading it correctly on line 1216 that if ANY model has request rate enabled that this will return true? Won't that break if there is a mix?  Can this function take a model name to fix that?

nv-braf: I've added a code and unit testing to check for this case. Please re-review.

PR #671, opened 2023-04-26
Update README and docs post-23.04 release

mc-nv:


Dockerfile
```diff
@@ -12,11 +12,11 @@
...
 # limitations under the License.
 
-ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:23.03-py3
-ARG TRITONSDK_BASE_IMAGE=nvcr.io/nvidia/tritonserver:23.03-py3-sdk
+ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:23.04-py3
+ARG TRITONSDK_BASE_IMAGE=nvcr.io/nvidia/tritonserver:23.04-py3-sdk
 
-ARG MODEL_ANALYZER_VERSION=1.27.0dev
-ARG MODEL_ANALYZER_CONTAINER_VERSION=23.04dev
+ARG MODEL_ANALYZER_VERSION=1.28.0dev
```
tgerdesnv, line 18: I don't think these are supposed to be updated now? Those are updated after freeze.

VERSION
```diff
@@ -1 +1 @@
-1.27.0dev
+1.28.0dev
```
tgerdesnv, line 1: not supposed to be updated?

tgerdesnv (changes requested):

Dockerfile
tgerdesnv: I see. We missed this last time so updating it now. This is fine

VERSION
tgerdesnv: We missed this last time so updating it now. This is fine

tgerdesnv (changes requested):
Please restore the README changes that were incorrectly removed [here](https://github.com/triton-inference-server/model_analyzer/commit/fb1bf9f521dd126d9e63561c81675c40b087cc2b)

tgerdesnv (approved):

PR #672, opened 2023-04-27
Adding request rate doc to config search

nv-braf:
Most documentation changes went in with the new config/CLI options.
Not sure where else this needs to be documented, other than in our release notes.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -513,10 +513,11 @@ cannot be specified globally.
 
 Options available under this parameter are described in table below:
 
-| Option Name   | Description                                             | Supporting Types                                   |
-| :------------ | :------------------------------------------------------ | :------------------------------------------------- |
-| `concurrency` | Request concurrency used for generating the input load. | `<range>`, `<comma-delimited-list>`, or a `<list>` |
-| `batch_sizes` | Static batch size used for generating requests.         | `<range>`, `<comma-delimited-list>`, or a `<list>` |
+| Option Name    | Description                                             | Supporting Types                                   |
+| :------------- | :------------------------------------------------------ | :------------------------------------------------- |
+| `concurrency`  | Request concurrency used for generating the input load. | `<range>`, `<comma-delimited-list>`, or a `<list>` |
+| `request_rate` | Request rate used for generating the input load.        | `<range>`, `<comma-delimited-list>`, or a `<list>` |
+| `batch_sizes`  | Static batch size used for generating requests.         | `<range>`, `<comma-delimited-list>`, or a `<list>` |
 
 An example `<parameter>` looks like below:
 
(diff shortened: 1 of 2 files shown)


docs/config.md
```diff
@@ -513,10 +513,11 @@ cannot be specified globally.
 
 Options available under this parameter are described in table below:
 
-| Option Name   | Description                                             | Supporting Types                                   |
-| :------------ | :------------------------------------------------------ | :------------------------------------------------- |
-| `concurrency` | Request concurrency used for generating the input load. | `<range>`, `<comma-delimited-list>`, or a `<list>` |
-| `batch_sizes` | Static batch size used for generating requests.         | `<range>`, `<comma-delimited-list>`, or a `<list>` |
+| Option Name    | Description                                             | Supporting Types                                   |
```
tgerdesnv, line 516: Do we need to mention anywhere that only one of RR or concurrency can be specified?

docs/config_search.md
```diff
@@ -255,9 +255,9 @@ profile_models:
 
 ---
 
-### **Limiting Batch Size, Instance Group, and Client Concurrency**
+### **Limiting Batch Size, Instance Group, and Client Concurrency or Request Rate**
```
tgerdesnv, line 258: Isn't there a line further up the file to list run-config-search-min/max-request-rate?

docs/config.md
nv-braf: No, I don't think so. The user should already understand this, if they are going to be deliberately using request rate; and, if they don't they'll get a nice error message explaining that you can't use both.

docs/config_search.md
nv-braf: Yes, but this is in relation to quick search. Which I realized I shouldn't have changed, since we don't support request rate for quick search. I've removed this edit.

docs/config_search.md
```diff
@@ -117,12 +117,18 @@ You can also modify the minimum/maximum values that the automatic search space w
 
 ---
 
-### [Request Concurrency Search Space](https://github.com/triton-inference-server/client/blob/main/src/c++/perf_analyzer/README.md#request-concurrency)
+### [Request Concurrency Search Space](https://github.com/triton-inference-server/client/blob/main/src/c++/perf_analyzer/docs/inference_load_modes.md#concurrency-mode))
```
tgerdesnv, line 120: Was this link broken? Shouldn't our CI be catching this?

tgerdesnv (approved):

docs/config_search.md
nv-braf: Link wasn't broken. It just wasn't as accurate as it could have been. Now we go directly to the page with the info about concurrency.

PR #676, opened 2023-05-03
Adding config option of max BCS steps

nv-braf:


docs/config.md
```diff
@@ -212,6 +212,9 @@ bls_composing_models: <comma-delimited-string-list>
 # Maximum request rate used for the automatic/quick config search
 [ run_config_search_max_request_rate: <int> | default: 8092 ]
 
+# Maximum number of steps taken during binary concurrency search
```
tgerdesnv, line 215: Is there any chance we add support for binary request rate? Should this be a more generic 'binary' option so we don't end up with `max_brs_steps` as well?  `run_config_search_max_binary_search_steps` (?)

tgerdesnv: Do we also want something (a magic -1 value?) to mean "Do a full binary search until you find the exact boundary"?

model_analyzer/config/input/config_command_profile.py
```diff
@@ -596,6 +596,15 @@ def _add_run_search_configs(self):
                 description=
                 "Value for the model's max_batch_size that run config search will start from."
             ))
+        self._add_config(
+            ConfigField(
+                'run_config_search_max_bcs_steps',
```
tgerdesnv, line 601: Is another story going to tie these two stories together? Or are you just waiting for the first one to merge before you start using the new option?

nv-braf: Just planned to make this change at the branch level

docs/config.md
nv-braf: I don't think that's necessary

tgerdesnv: So the decision is to keep concurrency (or 'c') in the option name?

nv-braf: To the first question, that's a good point. I'll make the name and description more generic

tgerdesnv (approved):

PR #677, opened 2023-05-03
Integrating concurrency search into quick search

nv-braf:
There's no unit testing, so I've run this on a live model (add_sub). Here are the results:
image
image
image


tgerdesnv:
Almost none of the data points in the summary above are to the right of the latency cutoff. Was this rerun using your latest changes?

PR #678, opened 2023-05-03
Add binary search to L0_quick_search

nv-braf:
Adding  `--latency-budget 10` to **L0_quick_search** will enable a binary search.

This exposed some bugs in the code around cases where PA doesn't return a result and we add `None` to the ConcurrencySearch's RCM list.

I've updated the code to check and deal with None.

### DIFF
--- a/model_analyzer/result/concurrency_search.py
+++ b/model_analyzer/result/concurrency_search.py
@@ -50,13 +50,14 @@ class ConcurrencySearch():
         self._max_concurrency_index = int(
             log2(config.run_config_search_max_concurrency))
 
-        self._run_config_measurements: List[RunConfigMeasurement] = []
+        self._run_config_measurements: List[Optional[RunConfigMeasurement]] = []
         self._concurrencies: List[int] = []
         self._last_failing_concurrency = 0
         self._last_passing_concurrency = 0
 
     def add_run_config_measurement(
-            self, run_config_measurement: RunConfigMeasurement) -> None:
+            self,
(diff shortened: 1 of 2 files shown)


model_analyzer/result/concurrency_search.py
```diff
@@ -111,31 +113,56 @@ def _has_objective_gain_saturated(self) -> bool:
...
         else:
             return False
 
     def _at_constraint_failure_boundary(self, index: int) -> bool:
-        at_failure_boundary = not self._run_config_measurements[
+        if not self._run_config_measurements[
+                index] or not self._run_config_measurements[index - 1]:
+            return False
+
+        at_failure_boundary = not self._run_config_measurements[  # type: ignore
```
tgerdesnv, line 162: are 2 type ignores needed? Makes this hard to read

tgerdesnv (commented):
I'm glad to hear testing caught issues!

model_analyzer/result/concurrency_search.py
nv-braf: Grrr...yes, it's really annoying, but I can't figure out any other way to do this.

tgerdesnv: optional measurements have always caused us problems. These type hints point to part of our design that we should try to clean up

nv-braf: Well, tell PA to stop returning no measurement and this wouldn't need to be optional!

PR #683, opened 2023-05-09
Add `cpu_only` option for ensemble composing models

nv-braf:
Adds a new config option `--cpu-only-composing-models` which asks for a list of composing models that will be profiled as `cpu_only`.

Added checks and unit testing to ensure this can only be specified if the model is an ensemble.
When the model profile is created will now set `cpu_only` if the composing model name is in the new list.
Added unit testing and ran a live test to confirm this is working as intended.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -276,7 +276,7 @@ profile_models: <comma-delimited-string-list|list|profile_model>
 # List of composing models for BLS models
 bls_composing_models: <comma-delimited-string-list>
 
-# List of an ensemble's composing models that should be profiled using CPU instances only
+# List of composing models that should be profiled using CPU instances only
 cpu_only_composing_models: <comma-delimited-string-list>
 
 # List of constraints placed on the config search results
--- a/docs/config.md
+++ b/docs/config.md
@@ -92,7 +92,7 @@ profile_models: <comma-delimited-string-list>
 # List of composing models for BLS models
 bls_composing_models: <comma-delimited-string-list>
 
-# List of an ensemble's composing models that should be profiled using CPU instances only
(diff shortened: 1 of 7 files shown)


docs/config.md
```diff
@@ -273,6 +276,9 @@ profile_models: <comma-delimited-string-list|list|profile_model>
 # List of composing models for BLS models
 bls_composing_models: <comma-delimited-string-list>
 
+# List of an ensemble's composing models that should be profiled using CPU instances only
```
tgerdesnv, line 279: BLS as well, right? Not just ensemble? Or do we not have control of that?

tests/test_quick_run_config_generator.py
```diff
@@ -910,11 +910,12 @@ def _get_next_run_config_ensemble(self,
         Also,
         - sequence batching should be on for model 1
         - dynamic batching should be on for model 2
+        - cpu_only should be set for model 2
```
tgerdesnv, line 913: This comment (model 2) conflicts with the variable name on line 943 (model_config1)

docs/config.md
nv-braf: Yes, this will work for BLS as well. I've changed the assertion to allow this feature for BLS models and updated the documentation.

tests/test_quick_run_config_generator.py
nv-braf: I've change the names to be model A/B to clear this up.

docs/config.md
tgerdesnv: ok but this doc statement still says ensemble

tgerdesnv (approved):

PR #685, opened 2023-05-10
Update 'main' to track development for 1.29.0 / r23.05

mc-nv:


debermudez:
Do you need to update the top level readme to show that we are tracking a new version?

mc-nv:
> Do you need to update the top level readme to show that we are tracking a new version?

@debermudez  I suppose you are referring to : https://github.com/triton-inference-server/model_analyzer/blob/main/README.md?plain=1#L21-L28 this part of the README.md file . 

We don't need to update it for this step, we will do it later.

debermudez:
thanks for checking me on that.

debermudez (approved):

PR #688, opened 2023-05-18
Detect when tritonserver fails to launch

nv-braf:
Will now scan the triton_output log looking for `Unexpected argument:` and will terminate the `wait_for_server_ready()` method early (saving about 50 retries and around 15 seconds of user time).

Will now print out the failure message rather than the generic client error.

Have tested this live and added a unit test to check this case.

Note: We will now delete an existing log file (rather than append) to ensure that any error messages seen are for the current run.

### DIFF
--- a/model_analyzer/triton/client/client.py
+++ b/model_analyzer/triton/client/client.py
@@ -65,6 +65,9 @@ class TritonClient:
                     log_file.seek(0)
                     log_output = log_file.read()
 
+                    if not type(log_output) == str:
+                        log_output = log_output.decode('utf-8')
+
                     if log_output:
                         if "Unexpected argument:" in log_output:
                             error_start = log_output.find(
--- a/model_analyzer/triton/client/client.py
+++ b/model_analyzer/triton/client/client.py
@@ -61,21 +61,7 @@ class TritonClient:
                     time.sleep(sleep_time)
                     retries -= 1
             except Exception as e:
-                if log_file:
-                    log_file.seek(0)
-                    log_output = log_file.read()
-
-                    if not type(log_output) == str:
-                        log_output = log_output.decode('utf-8')
-
-                    if log_output:
-                        if "Unexpected argument:" in log_output:
-                            error_start = log_output.find(
-                                "Unexpected argument:")
-                            raise TritonModelAnalyzerException(
-                                f'Error: TritonServer did not launch successfully\n\n{log_output[error_start:]}'
-                            )
-
+                self._check_for_triton_log_errors(log_file)
                 time.sleep(sleep_time)
                 retries -= 1
                 if retries == 0:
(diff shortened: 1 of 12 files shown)


model_analyzer/triton/client/client.py
```diff
@@ -28,7 +28,12 @@ class TritonClient:
     TritonClientFactory
     """
 
-    def wait_for_server_ready(self, num_retries, sleep_time=1):
+    def wait_for_server_ready(
+        self,
+        num_retries,
+        sleep_time=1,
+        log_file=None,
```
tgerdesnv, line 35: can you update the docstring below?

model_analyzer/triton/client/client.py
```diff
@@ -53,6 +58,7 @@ def wait_for_server_ready(self, num_retries, sleep_time=1):
                     time.sleep(sleep_time)
                     retries -= 1
             except Exception as e:
+                self._check_for_triton_log_errors(log_file)
```
tgerdesnv, line 61: am i reading line 55 correctly that even after the server is ready we still wait one more time?

tgerdesnv: also, what is it that throws into this except block? The self._client.is_server_ready()? Because your code is only in the except block, so if nothing is throwing an exception it will never run

model_analyzer/triton/client/client.py
```diff
@@ -174,3 +180,20 @@ def is_server_ready(self):
...
+
+        if not type(log_output) == str:
+            log_output = log_output.decode('utf-8')
+
+        if log_output:
+            if "Unexpected argument:" in log_output:
+                error_start = log_output.find("Unexpected argument:")
+                raise TritonModelAnalyzerException(
+                    f'Error: TritonServer did not launch successfully\n\n{log_output[error_start:]}'
+                )
```
tgerdesnv, line 199: would still be nice to have a catchall for a generic triton error. Is there a standard error string they print?

model_analyzer/triton/client/client.py
```diff
@@ -174,3 +180,20 @@ def is_server_ready(self):
         Returns true if the server is ready. Else False
         """
         return self._client.is_server_ready()
+
+    def _check_for_triton_log_errors(self, log_file):
```
tgerdesnv, line 184: wondering if this responsibility should be on the server class and not the client class to parse and understand the server output. Weird that the client is reaching in to the server log and making decisions on it. Then again not sure how to make that happen without passing the server object into the client.

model_analyzer/triton/client/client.py
nv-braf: Good point. I've added to the block above as well. 

I believe that what throws us into the exception block is the fact that the server has an error (due to the bad argument), but I've also seen us go into this block for non-errors, which is why I think the code is written the way it is: counting down retry in both cases and not aborting right away.

model_analyzer/triton/client/client.py
nv-braf: Yes, it would. I'll file a follow on story to do this.

model_analyzer/triton/client/client.py
nv-braf: Done

model_analyzer/triton/client/client.py
nv-braf: I thought it was cleaner this way, else you have to pass the entire sever class into the client and then call a method there that checks the output.

tgerdesnv (approved):

PR #689, opened 2023-05-18
Capture PA errors in a log file

nv-braf:
Current behavior only prints PA errors to stdout. 

The new behavior will capture PA errors that occur during profiling to a log file, inform the user that errors occurred, and point them to the new log file.

The log file contains (per PA error) the configuration name, PA command line. and PA error.

There is no unit testing for this (as neither metrics manager or analyzer has a unit test env), but I've run this on live tests that have known PA errors and observed the log file being properly created.

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -137,7 +137,6 @@ class Analyzer:
             if self._metrics_manager.encountered_perf_analyzer_error():
                 logger.warning(f"Perf Analyzer encountered an error when profiling one or more configurations. " \
                       f"See {PA_ERROR_LOG_FILENAME} for further details.\n")
-                print(f"")
 
     def report(self, mode: str) -> None:
         """
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -135,8 +135,9 @@ class Analyzer:
                             model.model_name()))
 
             if self._metrics_manager.encountered_perf_analyzer_error():
+                output_path = self._config.perf_output_path if self._config.perf_output_path else self._config.export_path
                 logger.warning(f"Perf Analyzer encountered an error when profiling one or more configurations. " \
-                      f"See {self._config.export_path}/{PA_ERROR_LOG_FILENAME} for further details.\n")
+                      f"See {output_path}/{PA_ERROR_LOG_FILENAME} for further details.\n")
 
     def report(self, mode: str) -> None:
         """
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -135,9 +135,8 @@ class Analyzer:
                             model.model_name()))
 
             if self._metrics_manager.encountered_perf_analyzer_error():
-                output_path = self._config.perf_output_path if self._config.perf_output_path else self._config.export_path
                 logger.warning(f"Perf Analyzer encountered an error when profiling one or more configurations. " \
-                      f"See {output_path}/{PA_ERROR_LOG_FILENAME} for further details.\n")
+                      f"See {self._config.export_path}/{PA_ERROR_LOG_FILENAME} for further details.\n")
 
     def report(self, mode: str) -> None:
         """
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -14,7 +14,7 @@
 
 from typing import List, Union, Optional
 import sys
-from model_analyzer.constants import LOGGER_NAME
+from model_analyzer.constants import LOGGER_NAME, PA_ERROR_LOG_FILENAME
 from .model_manager import ModelManager
 from .result.result_manager import ResultManager
 from .result.result_table_manager import ResultTableManager
@@ -133,6 +133,11 @@ class Analyzer:
                     logger.info(
                         self._get_report_command_help_string(
                             model.model_name()))
+
+            if self._metrics_manager.encountered_perf_analyzer_error():
+                logger.warning(f"Perf Analyzer encountered an error when profiling one or more configurations. " \
+                      f"See {PA_ERROR_LOG_FILENAME} for further details.\n")
+                print(f"")
 
     def report(self, mode: str) -> None:
         """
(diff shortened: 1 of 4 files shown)


model_analyzer/constants.py
```diff
@@ -57,5 +57,8 @@
 # Logging
 LOGGER_NAME = "model_analyzer_logger"
 
+# PA Error Log Filename
+PA_ERROR_LOG_FILENAME = "perf_analyzer_error.log"
```
tgerdesnv, line 61: should this be configurable? Is there any chance the CWD isn't writable?
What if the file already exists from a previous run? Is this deleted?  Was any thought given to using a temp file?

model_analyzer/record/metrics_manager.py
```diff
@@ -474,17 +485,10 @@ def _run_perf_analyzer(self, run_config, perf_output_writer):
...
 
-        if perf_output_writer:
-            perf_output_writer.write(
-                '============== Perf Analyzer Launched ==============\n'
-                f'Command: {perf_analyzer.get_cmd()}\n\n',
-                append=True)
-            if perf_analyzer.output():
-                perf_output_writer.write(perf_analyzer.output() + '\n',
-                                         append=True)
+        self._write_perf_analyzer_output(perf_output_writer, perf_analyzer)
```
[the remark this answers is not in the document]
nv-braf: It's a file the user can specify (see `--perf_output/--perf_output_path`). And, yes, if the user specifies `--perf_output`, the PA output will also be written to the error file.

model_analyzer/constants.py
nv-braf: By default we write to CWD, but the user could change this by specifying an export path. I'll change this so that the error log file also goes to the export path.

model_analyzer/analyzer.py
```diff
@@ -134,6 +134,11 @@ def profile(self, client: TritonClient, gpus: List[GPUDevice], mode: str,
                         self._get_report_command_help_string(
                             model.model_name()))
 
+            if self._metrics_manager.encountered_perf_analyzer_error():
+                logger.warning(f"Perf Analyzer encountered an error when profiling one or more configurations. " \
+                      f"See {PA_ERROR_LOG_FILENAME} for further details.\n")
+                print(f"")
```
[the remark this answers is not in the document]
nv-braf: Good catch. That's not supposed to be there anymore. Removed

model_analyzer/analyzer.py
```diff
@@ -134,6 +134,11 @@ def profile(self, client: TritonClient, gpus: List[GPUDevice], mode: str,
                         self._get_report_command_help_string(
                             model.model_name()))
 
+            if self._metrics_manager.encountered_perf_analyzer_error():
+                logger.warning(f"Perf Analyzer encountered an error when profiling one or more configurations. " \
+                      f"See {PA_ERROR_LOG_FILENAME} for further details.\n")
```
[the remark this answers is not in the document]
nv-braf: It is, but I don't think it's a good idea. They can be quite long and I think it's too much info for the summary output and might confuse the user about what's going on.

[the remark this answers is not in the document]
tgerdesnv: In the future it would be good to parse/bucket the failures into common cases and communicate it to stdout so the users don't have to read and parse a huge log

model_analyzer/constants.py
tgerdesnv: export_path instead of perf_output_path?

model_analyzer/analyzer.py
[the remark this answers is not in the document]
tgerdesnv: and, the output should point to the full file, not just the file name. `{self._config.export_path}/{PA_ERROR_LOG_FILENAME}` (or perf_output_path if you update it to be there instead)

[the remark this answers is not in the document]
nv-braf: Hopefully PA errors are rare enough that this doesn't matter!
I've pointed the output to full file path.

model_analyzer/constants.py
nv-braf: Yes, there will not be a perf output path specified in most cases (since perf logging is off by default).

nv-braf: And, it seemed like a lot of unneeded code to figure out when to switch btw. using export vs. perf_output path.

tgerdesnv: no switching between. It would always be perf_output_path. If unspecified, it defaults to CWD. It doesn't make sense that perf_output_path controls half of the potential PA output, and that the other goes into a place where all of our data goes.

tgerdesnv: ```If perf_output_path is None, output will be written to stdout.```
Got it, ok that doesn't default like I had expected.  export path it is 👍 

Are there any doc changes needed about this new output file and what controls where it goes?

nv-braf: We append it to the end of a previous log file, but will only show the warning message if new errors show up.

tgerdesnv: Yeah I think we need to delete it. It would be too confusing otherwise.
Imagine they have some shape errors. They look at the error log and go fix it. Now they rerun and there is a new PA failure. They open the file and see (no longer valid) shape errors.

nv-braf: Done

model_analyzer/record/metrics_manager.py
```diff
@@ -497,6 +501,42 @@ def _run_perf_analyzer(self, run_config, perf_output_writer):
...
+                '============== Perf Analyzer Launched ==============\n'
+                f'Command: {perf_analyzer.get_cmd()}\n\n',
+                append=True)
+            if perf_analyzer.output():
+                perf_output_writer.write(perf_analyzer.output() + '\n',
+                                         append=True)
+
+    def _handle_unsuccessful_perf_analyzer_run(
+            self, perf_analyzer: PerfAnalyzer) -> None:
+        output_path = self._config.perf_output_path if self._config.perf_output_path else self._config.export_path
```
tgerdesnv, line 518: Sorry for the churn. I don't think it should be variable like this, especially because now two different objects need to understand how to re-calculate this path. Too hard to document and too confusing. I misunderstood what perf_output_path was for and lead you down the wrong path.  Let's just stick with the export path

tgerdesnv (approved):

PR #697, opened 2023-05-31
Stop MA if failures are detected on initial profiles

nv-braf:
We will now stop MA and report an error to the user, rather than printing out a blank table and report, if MA detects that the first two measurements do not get valid results from PA.


model_analyzer/constants.py
```diff
@@ -62,3 +62,6 @@
 
 # Constraints
 GLOBAL_CONSTRAINTS_KEY = "__default__"
+
+# Measurement threshold
```
tgerdesnv, line 66: This comment means nothing. Either expand it or remove it.

model_analyzer/model_manager.py
```diff
@@ -188,9 +195,26 @@ def _check_for_ensemble_model_incompatibility(
...
+        else:
+            self._measurement_attempts += 1
+
+    def _stop_ma_if_no_valid_measurement_threshold_reached(self) -> None:
+        if self._received_measurement_values_from_pa:
+            return
+
+        if self._measurement_attempts >= INVALID_MEASUREMENT_THRESHOLD:
+            raise TritonModelAnalyzerException(
+                            f'The first {INVALID_MEASUREMENT_THRESHOLD} attempts to acquire measurements have failed. ' \
```
tgerdesnv, line 218: Did this auto-format like this? Why is it so indented? It seems to be too wide, even for my ultra wide screen monitor.

qa/L0_perf_analyzer/test.sh
```diff
@@ -63,8 +63,6 @@ for CONFIG_FILE in ${LIST_OF_CONFIG_FILES[@]}; do
         TEST_NAME="count_window"
     elif [[ "$CONFIG_FILE" == "config-additive-args-count-no-adjust.yml" ]]; then
         TEST_NAME="count_window"
-    elif [[ "$CONFIG_FILE" == "config-perf-output-timeout.yml" ]]; then
```
tgerdesnv, line 66: Is this no longer valid?

model_analyzer/model_manager.py
```diff
@@ -188,9 +195,26 @@ def _check_for_ensemble_model_incompatibility(
...
         self._state_manager.set_state_variable(
             'ModelManager.model_variant_name_manager',
             self._state_manager.default_encode(ModelVariantNameManager()))
+
+    def _check_for_valid_measurement(
+            self, measurement: Optional[RunConfigMeasurement]) -> None:
+        if measurement:
+            self._received_measurement_values_from_pa = True
+        else:
+            self._measurement_attempts += 1
```
tgerdesnv, line 210: Based on the name, should this always increment?  I know you are only using it in a specific way, but we might as well do the right thing so it doesn't burn us later.  Otherwise rename to failed_measurement_attempts or something.

model_analyzer/constants.py
nv-braf: Fixed

model_analyzer/model_manager.py
nv-braf: It did. I've fixed the formatting by breaking it into 3 lines. Auto-formatting is "funny" with fstrings, if I don't manually breakup the line it will let me continue way past the character limit.

qa/L0_perf_analyzer/test.sh
nv-braf: Correct. This checked that the summary table was empty. Now it will fail because we have two consecutive PA timeouts. I've made this a unit test now.

model_analyzer/model_manager.py
nv-braf: Good point. I've changed variable name to `failed_measurement_attempts`.

tgerdesnv (approved):

PR #700, opened 2023-06-05
Create export_path if it doesn't exist

nv-braf:
Instead of erroring, we will now create an export path for the user if it doesn't exist, but it's parent directory does. 
We will still error in the case where the export path is a file or if multiple levels of directories need to be created.

### DIFF
--- a/model_analyzer/config/input/config_command_profile.py
+++ b/model_analyzer/config/input/config_command_profile.py
@@ -822,7 +822,7 @@ class ConfigCommandProfile(ConfigCommand):
                         flags=['-e', '--export-path'],
                         default_value=DEFAULT_EXPORT_PATH,
                         field_type=ConfigPrimitive(
-                            str, validator=file_path_validator),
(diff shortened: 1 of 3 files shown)


model_analyzer/config/input/config_command_profile.py
```diff
@@ -1055,9 +1055,12 @@ def _preprocess_and_verify_arguments(self):
             logger.warning(
                 f"--export-path not specified. Using {self._fields['export_path'].default_value()}"
             )
-        elif self.export_path and not os.path.isdir(self.export_path):
+        elif os.path.exists(self.export_path) and os.path.isfile(
```
tgerdesnv, line 1058: Why would this change from "not isdir" to "isfile" when the error message is saying "this is not a dir".

nv-braf: Fixed

tgerdesnv (approved):

PR #702, opened 2023-06-07
Add Support for Custom Docker Args

nv-braf:


model_analyzer/config/input/config_command_profile.py
```diff
@@ -294,8 +294,77 @@ def _add_profile_models_configs(self):
                             type_=str))
                 for k in PerfAnalyzerConfig.allowed_keys()
             })
+
         triton_server_environment_scheme = ConfigObject(
             schema={'*': ConfigPrimitive(str)})
+
+        # This comes from /usr/local/lib/python3.8/dist-packages/docker/models/containers.py
```
tgerdesnv, line 301: /usr/local/bin is a local path on your machine, right? Is there a better way to point to the source?

model_analyzer/config/input/config_command_profile.py
```diff
@@ -1226,6 +1310,12 @@ def _autofill_values(self):
                     'triton_server_environment'] = model.triton_server_environment(
                     )
 
+            # triton docker args
+            if not model.triton_docker_args():
+                new_model['triton_docker_args'] = self.triton_docker_args
+            else:
+                new_model['triton_docker_args'] = model.triton_docker_args()
```
tgerdesnv, line 1317: This is global vs model-specific args?  But it is all or nothing, right?  If globally has 3 options set and then one is specified at the model level, only that one option is used?

model_analyzer/triton/server/server_docker.py
```diff
@@ -142,7 +146,8 @@ def start(self, env=None):
                 tty=False,
                 stdin_open=False,
                 detach=True,
-                shm_size=self._shm_size)
+                shm_size=self._shm_size,
+                **self._args)
```
tgerdesnv, line 150: Is there anything that makes sure the same arg isn't passed in twice? What if they pass in shm_size through the new option instead of the old one?

nv-braf: Nothing explicit that I've done since this will cause an exception that says: `run() got multiple values for keyword argument 'shm_size'

Felt like that was good enough and a user should be able to figure out what they've done wrong.

docs/config.md
```diff
@@ -304,6 +304,10 @@ cpu_only_composing_models: <comma-delimited-string-list>
 # Dict of name=value pairs containing metadata for the tritonserver docker container
 # launched in docker launch mode
 [ triton_docker_labels: <dict> ]
+
+# Dict of name=value pairs containing arguments that will be applied to the tritonserver docker
+# container launched in docker launch mode
```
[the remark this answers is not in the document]
nv-braf: Fixed

model_analyzer/config/input/config_command_profile.py
nv-braf: Correct. It's an installed package. I've updated the message and removed reference to my local path

tgerdesnv (approved):

PR #710, opened 2023-06-09
Periodic checkpointing in MA

pskiran1:


qa/L0_state_management/check_results.py
```diff
@@ -85,8 +85,8 @@ def check_interrupt_handling(self):
         if log_contents.find(token) == -1:
             return False
 
-        # check that 2nd model is profiled once
-        token = f"Profiling {self._profile_models[1]}"
+        # check that 1st model is profiled twice
```
nv-braf, line 88: I don't understand this change/comment. In the test.sh it states that it is going to throw an interrupt when it detects the creation of the `1.ckpt` file.
With your change my understanding is that we would not be creating such a file (the 0.ckpt would contain all 3 models). Can you confirm what the new behavior is now?

qa/L0_state_management/check_results.py
```diff
@@ -117,7 +117,7 @@ def check_early_exit(self):
         return True
 
     def check_continue_after_checkpoint(self,
-                                        expected_resnet_count=3,
+                                        expected_resnet_count=4,
```
nv-braf, line 120: This makes the comment of line 137 incorrect? Can you explain how you now guarantee that no measurements are taken before the interrupt? I would have expected changes in the `test.sh` for this test were needed

nv-braf (changes requested):
How many times have you run the L0_state_management test? I was getting inconsistent results from the CI

qa/L0_state_management/check_results.py
pskiran1: In the `test.sh` script, at line 109, script will wait until the first checkpoint is created.

Previously, the MA would create a checkpoint after completing the model profiling. So when the `0.ckpt` checkpoint was created(indicating completion of profiling two configurations for the first model), the script would break the loop and generate an interrupt. Also meanwhile, the first configuration of the second model would also get profiled.

Now, with the updated approach, a checkpoint is written for every configuration (multiple times for the same model). So MA will generate a checkpoint after the first configuration of the first model is executed. As a result, script would break the loop and only the second configuration of the first model will be profiled thereafter, and the first configuration of the second model will never be started.

So I modified the logic to check only for the first model, to make sure 2 configurations are profiled.

qa/L0_state_management/check_results.py
pskiran1: Yes, as mentioned in the previous comment, the profiling of the second model will not be started.

After the `interrupt_handling` test case is finished, only the first model will be profiled with its two configurations. Then, in the test case `check_continue_after_checkpoint`, the remaining two models will be profiled, with two configurations each. Therefore, we need to verify a total of four entries for the first to models.

This makes the comment on line 137 incorrect. I need to modify the comment, sorry for missing doing so.

I think to fix this script failure, we can make modifications to either `test.sh` or `check_results.py`. I have chosen to modify `check_results.py`. Please let me know if you would like `test.sh` to be modified instead.

qa/L0_state_management/check_results.py
nv-braf: Is this working consistently for you? I'm worried it's possible to potentially sneak a second profile within the 5 second sleep + kill. 

Please change the comment in `test.sh` to indicate that a single configuration of the first model will be profiled before the interrupt occurs

qa/L0_state_management/check_results.py
nv-braf: Please update the comment on line 137

nv-braf (approved):

pskiran1:
> How many times have you run the L0_state_management test? I was getting inconsistent results from the CI

The pipeline was successfully executed in GitLab CI. I executed the job three times consecutively and triggered three pipelines with a few hours between each run. In both cases, there were no failures observed.

PR #711, opened 2023-06-13
Detailed reporting after profile

nv-braf:
The big change from the previous version is to `_run_report_command()` which now creates a new config and substitutes it for the profile config before calling the report method.

The other change is in `CLI()` which will now accept an optional arg list (instead of always grabbing from the command line).


model_analyzer/cli/cli.py
```diff
@@ -128,12 +139,16 @@ def parse(self):
         args : Namespace
             Object that contains the parse CLI commands
             Used for the global options
+        input_args: List
```
tgerdesnv, line 142: Can you clean up this docstring? This is under 'returns', which isn't true.

tgerdesnv (commented):
Please clean up the codeql issues as well

model_analyzer/cli/cli.py
nv-braf: Fixed

tgerdesnv (approved):

PR #715, opened 2023-06-20
Detect and handle invalid metrics url

nv-braf:
Will now raise an error in local or docker mode if no GPU records are returned from the tritonserver before the first measurement is taken.

### DIFF
--- a/model_analyzer/monitor/remote_monitor.py
+++ b/model_analyzer/monitor/remote_monitor.py
@@ -52,10 +52,18 @@ class RemoteMonitor(Monitor):
                 f"GPU monitoring does not currently support the following metrics: {unsupported_metrics}]"
             )
 
+    def is_monitoring_connected(self) -> bool:
+        try:
+            status_code = requests.get(self._metrics_url).status_code
+        except:
+            return False
+
+        return status_code == requests.code["okay"]
+
     def _monitoring_iteration(self):
         """
         When this function runs, it requests all the metrics
-        that triton has collected aand organizes them into
+        that triton has collected and organizes them into
         the dict. This function should run as fast
         as possible
         """
@@ -66,7 +74,7 @@ class RemoteMonitor(Monitor):
     def _collect_records(self):
         """
         This function will organize the metrics responses
-        and creat Records out of them
+        and create Records out of them
         """
 
         records = []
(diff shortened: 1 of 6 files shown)


model_analyzer/record/metrics_manager.py
```diff
@@ -559,9 +559,20 @@ def _get_gpu_inference_metrics(self):
         # Stop and destroy DCGM monitor
         gpu_records = self._gpu_monitor.stop_recording_metrics()
 
+        if not gpu_records and self._expected_gpu_records():
```
tgerdesnv, line 562: It still feels like the gpu monitor should have a heartbeat check, instead of passing this responsibility up to the metrics manager.

model_analyzer/record/metrics_manager.py
```diff
@@ -559,9 +559,20 @@ def _get_gpu_inference_metrics(self):
...
+            raise TritonModelAnalyzerException(
+                f'No GPU metrics returned. Please check that the `triton_metrics_url` value is set correctly.'
+            )
+
         gpu_metrics = self._aggregate_gpu_records(gpu_records)
         return gpu_metrics
 
+    def _expected_gpu_records(self) -> bool:
```
tgerdesnv, line 570: I think this is poorly named, or missing extra conditions. Based on its name, it should also look at 'cpu_only'. I know that flag is evaluated earlier (and that this function will only be called if cpu_only is false), but this could be confusing to a developer in the future.

Also, as a follow up story, remote should work fine. Not sure about c_api yet.

Also, as a follow up story (related to remote), we should consider changing the default metrics URL to be based off of the triton server url.  You can look at the recent PA change I made around this [here](https://github.com/triton-inference-server/client/pull/338/files).  For normal PA runs, MA wouldn't need to change anything. However, to get the server-only metrics it would need a similar change.

model_analyzer/monitor/remote_monitor.py
```diff
@@ -52,10 +52,18 @@ def __init__(self, metrics_url, frequency, metrics):
                 f"GPU monitoring does not currently support the following metrics: {unsupported_metrics}]"
             )
 
+    def is_monitoring_connected(self) -> bool:
+        try:
+            status_code = requests.get(self._metrics_url).status_code
+        except:
+            return False
+
+        return status_code == requests.codes["okay"]
```
tgerdesnv, line 61: This is the equivalent of code 200?

nv-braf: Yes

tgerdesnv (approved):

PR #717, opened 2023-06-21
Fix Coordinate comparison error

nv-braf:
Added back in the __lt__ method and unit tests to Coordinate class and added slow mode to one of the unit tests so we have coverage of the method and failure point observed by the user.


tests/test_neighborhood.py
```diff
@@ -757,6 +757,7 @@ def test_all_same_throughputs(self):
         step_vector = n._get_step_vector()
         self.assertEqual(step_vector, Coordinate([0, 0, 0]))
 
+        n.force_slow_mode()
```
tgerdesnv, line 760: do we lose fast-mode coverage with this? Should it just be a new test (or the same test called twice with a bool for slow/fast mode)?

tgerdesnv (approved):

PR #718, opened 2023-06-22
Auto-formatting

dyastremsky:
With the [Clang upgrade in our formatter](https://github.com/triton-inference-server/common/pull/94), we are running the formatter on all Triton repos. If you would like to use the formatting tool in this repo, feel free to approve this PR.

### DIFF
--- a/model_analyzer/plots/simple_plot.py
+++ b/model_analyzer/plots/simple_plot.py
@@ -143,9 +143,8 @@ class SimplePlot:
 
         for model_config_name, data in self._data.items():
             # Sort the data by x-axis
-            x_data, y_data = (
-                list(t)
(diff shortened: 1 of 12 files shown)


nv-braf:
Thanks, these files should have the correct formatting. And, when I save them locally it matches the changes I'm seeing in this PR. Looks like they were all touched last by Sai, I'll talk with him to and ensure that his environment is setup correctly.

nv-braf (approved):

dyastremsky:
Thanks for the quick review, @nv-braf! We just upgraded Clang from 6.0 to 15 due to the containers now using Ubuntu 22.04, so some of these changes might be due to that upgrade.

Issue #725, opened 2023-07-11
AttributeError: 'TritonServerDocker' object has no attribute '_log_file'. Did you mean: 'log_file'?


levipereira:
Model analyzer return error on triton-server-23.06.

  ```
  docker run -it --gpus all  \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v /storage/triton-server/devel/triton-server-23.06/models_01:/storage/triton-server/devel/triton-server-23.06/models_01 \
  -v /storage/triton-server/devel/triton-server-23.06/model_analyzer:/storage/triton-server/devel/triton-server-23.06/model_analyzer \
  -v /storage/triton-server/devel/triton-server-23.06/model_analyzer/analysis_results:/storage/triton-server/devel/triton-server-23.06/model_analyzer/analysis_results  \
  --net=host  nvcr.io/nvidia/tritonserver:23.06-py3-sdk  
```

```
model-analyzer profile     --model-repository /storage/triton-server/devel/triton-server-23.06/models_01    \
 --profile-models  yolov7 \
--triton-launch-mode=docker     \
--output-model-repository-path /storage/triton-server/devel/triton-server-23.06/model_analyzer/output_model    \
--export-path /storage/triton-server/devel/triton-server-23.06/model_analyzer/analysis_results  \
--override-output-model-repository

[Model Analyzer] Initializing GPUDevice handles
[Model Analyzer] Using GPU 0 NVIDIA GeForce RTX 2070 SUPER with UUID GPU-3fed825f-252b-32ea-e3d7-266c45b62ce7
[Model Analyzer] WARNING: Overriding the output model repo path "/storage/triton-server/devel/triton-server-23.06/model_analyzer/output_model"
[Model Analyzer] Starting a Triton Server using docker
[Model Analyzer] No checkpoint file found, starting a fresh run.
[Model Analyzer] Profiling server only metrics...
Traceback (most recent call last):
  File "/usr/local/bin/model-analyzer", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/model_analyzer/entrypoint.py", line 267, in main
    analyzer.profile(client=client,
  File "/usr/local/lib/python3.10/dist-packages/model_analyzer/analyzer.py", line 112, in profile
    self._get_server_only_metrics(client, gpus)
  File "/usr/local/lib/python3.10/dist-packages/model_analyzer/analyzer.py", line 214, in _get_server_only_metrics
[...]

PR #727, opened 2023-07-12
Fix error in docker mode when triton output path is not specified

nv-braf:
See [https://github.com/triton-inference-server/model_analyzer/issues/725](https://github.com/triton-inference-server/model_analyzer/issues/725) for details on the failure.

The solution is to implement a temp log file (to match the code in `server_local.py`)

### DIFF
--- a/model_analyzer/triton/server/server_docker.py
+++ b/model_analyzer/triton/server/server_docker.py
@@ -15,8 +15,10 @@
 # limitations under the License.
 
 import logging
+import tempfile
 from io import TextIOWrapper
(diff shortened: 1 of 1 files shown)


nv-hwoo (approved):
Quick question: Is it ok if we just silently create a random file for the output without letting user know about it? Other than that, LGTM!

nv-braf:
> Quick question: Is it ok if we just silently create a random file for the output without letting user know about it? Other than that, LGTM!

Yes, It's a temp file that automatically gets removed when the program terminates. The user will never even know it existed.

PR #730, opened 2023-07-19
Quick search preserves model config parameters

nv-braf:
In brute search mode, we toggle between automatic and manual search based on the presence of `model_config_parameters`.
In quick search mode, we currently ignore any model config parameters and silently overwrite them when generating configurations.

This is remedied in two ways:

1. Added a config check which checks that only a single configuration of parameters can be generated (else it tells the user to run in brute search mode)
2. Quick search parameters are now additive and will preserve the original model config parameters set by the user

Unit testing was added to check for these cases and documentation is updated to reflect this restriction in quick search.

### DIFF
--- a/model_analyzer/config/generate/quick_run_config_generator.py
+++ b/model_analyzer/config/generate/quick_run_config_generator.py
@@ -431,7 +431,10 @@ class QuickRunConfigGenerator(ConfigGeneratorInterface):
             model_config_params.pop("max_batch_size", None)
 
             # This is guaranteed to only generate one combination (check is in config_command)
-            param_combo = GeneratorUtils.generate_combinations(model_config_params)[0]
+            param_combos = GeneratorUtils.generate_combinations(model_config_params)
+            assert len(param_combos) == 1
+
+            param_combo = param_combos[0]
         else:
             param_combo = {}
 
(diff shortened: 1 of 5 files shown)


model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -424,17 +426,24 @@ def _get_next_model_config(
             self._coordinate_to_measure, dimension_index
         )
 
+        model_config_params = deepcopy(model.model_config_parameters())
+        if model_config_params:
+            model_config_params.pop("max_batch_size", None)
+
+            # This is guaranteed to only generate one combination (check is in config_command)
+            param_combo = GeneratorUtils.generate_combinations(model_config_params)[0]
```
tgerdesnv, line 434: can we recheck here for sanity? As an assert? I don't like this coupling, having one section of the code make an unchecked assumption about other code.
param_combos = blah
if (len != 1) raise error
param_combo = param_combos[0]

nv-braf: Assertion added

tgerdesnv (approved):

nv-hwoo (approved):

PR #734, opened 2023-07-24
New config option: --always-report-gpu-metrics

nv-braf:
Added a new option to profile and report: `--always-report-gpu-metrics`

When this is set MA will always report GPU metrics for both the summary and detailed reports, even if the model is `cpu_only`.

Added unit tests to check this new option and also performed a live run.


tests/test_report_manager.py
```diff
@@ -295,10 +307,10 @@ def subtest_build_summary_table(self, mode, cpu_only):
             )
 
     def test_build_detailed_info(self):
-        for cpu_only in [True, False]:
-            self._subtest_build_detailed_info(cpu_only)
+        for report_gpu_metrics in [True, False]:
+            self._subtest_build_detailed_info(report_gpu_metrics)
```
nv-hwoo, line 311: Is there a reason why we don't loop over `cpu_only` options here?

model_analyzer/reports/report_manager.py
```diff
@@ -508,20 +505,14 @@ def _build_summary_table(
...
-        summary_table = (
-            self._construct_summary_result_table_cpu_only(
-                sorted_measurements, multi_model, has_composing_models
-            )
-            if cpu_only
-            else self._construct_summary_result_table(
-                sorted_measurements, multi_model, has_composing_models
-            )
+        summary_table = self._construct_summary_result_table(
+            sorted_measurements, multi_model, has_composing_models, report_gpu_metrics
```
nv-hwoo, line 515: So much cleaner 🤩

model_analyzer/config/input/config_command_profile.py
```diff
@@ -18,7 +18,7 @@
 import logging
 import os
 
-import numba
+import numba.cuda
```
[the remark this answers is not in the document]
nv-braf: Yes. Without this change I was getting an error when running in debug mode from vscode.

model_analyzer/record/metrics_manager.py
```diff
@@ -244,7 +244,7 @@ def profile_models(self, run_config: RunConfig) -> Optional[RunConfigMeasurement
             if not self._config.perf_output
             else FileWriter(self._config.perf_output_path)
         )
-        cpu_only = run_config.cpu_only()
+        cpu_only = run_config.cpu_only() and not self._config.always_report_gpu_metrics
```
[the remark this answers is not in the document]
nv-braf: Good catch. I should have updated this like I did for report manager. I've changed this to be `capture_gpu_metrics`. Please re-review the changes.

tests/test_report_manager.py
nv-braf: Yes. Because I forgot to add it! This is fixed now.

nv-hwoo (approved):

docs/config.md
```diff
@@ -227,6 +227,9 @@ cpu_only_composing_models: <comma-delimited-string-list>
 # Enables the searching of request rate (instead of concurrency)
 [ request_rate_search_enable: <bool> | default: false]
 
+# Always report GPU metrics
```
tgerdesnv, line 230: This comment in a .md file doesn't add any value. Can you mention the case where it would be valuable?

docs/config.md
```diff
@@ -227,6 +227,9 @@ cpu_only_composing_models: <comma-delimited-string-list>
...
+[ always_report_gpu_metrics: <bool> | default: false]
```
tgerdesnv, line 231: I don't love the long name. Looking around, I see "collect-cpu-metrics". Can we parallel that with "collect-gpu-metrics"? And state that the default depends on cpu-only of the model(s)?

I guess that might be messy though because then you need to know if it was specified by the user or not?

nv-braf: I don't mind the name (obviously!). I much prefer it to be slightly longer and descriptive over the alternative.

docs/config.md
nv-braf: Done

tgerdesnv (approved):

PR #739, opened 2023-08-03
Add L0 test to protect end-to-end ensemble model

pskiran1:


qa/L0_ensemble_model/check_results.py
```diff
@@ -0,0 +1,104 @@
...
+        Check that each model was profiled the number of times
+        corresponding with batch size and concurrency combinations
+
+        (No model config parameter combos expected here!)
+        """
+
+        with open(self._analyzer_log, "r") as f:
+            log_contents = f.read()
+
+        expected_min_num_measurements = 20
```
nv-braf, line 52: When you ran this on average how many measurements are taken?

qa/common/util.sh
```diff
@@ -22,7 +22,7 @@ ANALYZER_LOG=${ANALYZER_LOG:="$LOGS_DIR/test.log"}
 
 mkdir -p $LOGS_DIR
 
-create_logs_dir() {
```
nv-braf, line 25: Why was this change made?

pskiran1: It seems that I forgot to include the "function" keyword when I implemented the function. I have now added it.

qa/L0_ensemble_model/check_results.py
pskiran1: I observed measurements ranging from 30 to 45 during the test runs. I considered the minimum and maximum counts mentioned above from the [L0_quick_search](https://github.com/triton-inference-server/model_analyzer/tree/main/qa/L0_quick_search) test.

nv-braf (approved):

PR #743, opened 2023-08-09
ModelConfigVariant dataclass

nv-braf:
Created a new dataclass, ModelConfigVariant, that wraps around the current ModelConfig class. This also us to break the coupling that the ModelConfig name field must hold the variant name; which will is necessary before switching to use the Tritonserver API in local/docker/remote mode.

Note that for now I have added backward checkpoint compatibility (checking for either ModelConfig or ModelConfigVariant), since I want to wait on regenerating the L0 checkpoints until after the work to add ensemble/BLS tests is complete.


model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -203,34 +206,33 @@ def make_model_config(
         model_config_dict["name"] = variant_name
```
tgerdesnv, line 206: I can't say I fully understand the changes. Should this still be here?

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -203,34 +206,33 @@ def make_model_config(
...
-        ensemble_composing_model_configs: List[ModelConfig],
+        ensemble_composing_model_config_variants: List[ModelConfigVariant],
         model_variant_name_manager: ModelVariantNameManager,
         param_combo: Dict = {},
-    ) -> ModelConfig:
+    ) -> ModelConfigVariant:
         """
-        Loads the ensemble model spec from the model repository, and then mutates
-        the names to match the ensemble composing models
+        Loads the ensemble model spec from the model repository
```
tgerdesnv, line 229: Is this comment true? Does it load the one from the repo? Or does it create a variant of it?

model_analyzer/config/generate/base_model_config_generator.py
```diff
@@ -257,9 +259,17 @@ def make_ensemble_model_config(
         )
 
         model_config_dict["name"] = variant_name
+        logger.info("")
+        if variant_found:
+            logger.info(f"Found existing ensemble model config: {variant_name}")
+        else:
+            logger.info(f"Creating ensemble model config: {variant_name}")
+        for str in logger_str:
```
tgerdesnv, line 267: I don't see other changes related to this. Where did logger_str come from?

model_analyzer/triton/model/model_config_variant.py
```diff
@@ -0,0 +1,29 @@
...
+
+@dataclass
+class ModelConfigVariant:
+    """
+    A dataclass that holds the ModelConfig as well as the variant name
+    for the model
+    """
+
+    model_config: ModelConfig
+    variant_name: str
```
tgerdesnv, line 29: Is there a reason we didn't just add variant_name into ModelConfig?

model_analyzer/config/run/model_run_config.py
```diff
@@ -77,18 +78,35 @@ def model_variant_name(self) -> str:
...
+        Returns
+        -------
+        ModelConfigVariant
+            The ModelConfigVariant corresponding to this run
+        """
+
+        return self._model_config_variant
 
-    def model_config(self) -> ModelConfig:
+    def model_config(self) -> Optional[ModelConfig]:
```
tgerdesnv, line 97: What changed to make this optional?

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: It's passed in to the method. I added this to make ensemble's printout the top-level creation step.

model_analyzer/triton/model/model_config_variant.py
nv-braf: Yes, this makes the code cleaner as a ModelConfig should not contain things (like the variant name) that are not part of the Triton model's metadata.

model_analyzer/config/generate/base_model_config_generator.py
nv-braf: Yes, until we make the switch to use the API we are still using the variant name in both the output repo and as the model name in the ModelConfig.

model_analyzer/triton/model/model_config_variant.py
tgerdesnv: Now everyone that wants anything from the model config has to do model_variant.model_config instead of just using the config. That seems less clean, that users need to know about this hierarchy.

pskiran1 (approved):

model_analyzer/reports/report_manager.py
```diff
@@ -613,7 +613,7 @@ def _create_summary_sentence(
     def _create_ensemble_summary_sentence(self, run_config: RunConfig) -> str:
         summary_sentence = "<BR><BR>"
         best_config_name = (
-            run_config.model_run_configs()[0].model_config().get_field("name")
+            run_config.model_run_configs()[0].model_config_variant().variant_name
```
nv-hwoo, line 616: I think we can call `model_variant_name()` to avoid accessing `model_config_variant` inside ModelRunConfig.

model_analyzer/record/metrics_manager.py
```diff
@@ -354,11 +359,11 @@ def _create_model_variant(
                 self._first_config_variant.setdefault(original_name, None)
 
                 if ignore_first_config_variant:
-                    variant_config.write_config_to_file(
+                    variant_config.model_config.write_config_to_file(
```
nv-hwoo, line 362: I feel like users using `ModelConfigVariant` shouldn't need to know to access internal `model_config` object to call this method. It would be better if `ModelConfigVariant` support `write_config_to_file` functionality directly.

model_analyzer/triton/model/model_config_variant.py
nv-hwoo: >Yes, this makes the code cleaner as a ModelConfig should not contain things (like the variant name) that are not part of the Triton model's metadata.

In this case, are we also considering to just make `ModelRunConfig` hold the extra information? It does seem like this change could introduce extra layer of calls/access.

nv-braf: A good thought (and how I was originally coding it), but that's ugly in different way, as now a bunch of methods would have to pass back tuples of ModelConfig, variant_name, which is what I'm trying to avoid with this new dataclass

model_analyzer/record/metrics_manager.py
nv-braf: It's a slippery slope bringing up methods from the child class. Obviously, we do bring up some methods as high as the ModelRunConfig class, but I think it's better to keep the MCV class as a pure dataclass.

model_analyzer/reports/report_manager.py
nv-braf: Done

PR #744, opened 2023-08-11
Local & Docker mode loading via config using Triton API

nv-braf:
In local and docker mode we will now load configs from a (ModelConfig) dictionary instead of pointing the TritonServer at a subdirectory of the output model repository.

Models are now loaded from the model repository with a different dictionary for each config. This removes the requirement that the model name match the subdirectory. Therefore, we will no longer mutate the model names in the output model repo.


model_analyzer/config/generate/quick_run_config_generator.py
```diff
@@ -483,7 +487,7 @@ def _create_next_model_run_config(
         composing_model_config_variants: List[ModelConfigVariant],
     ) -> ModelRunConfig:
         perf_analyzer_config = self._get_next_perf_analyzer_config(
-            model_config_variant.variant_name, model, model_index
+            model.model_name(), model, model_index
```
nv-hwoo, line 490: Was this an intended change? Just wanted to make sure since everything seemed like it was changing to use `model_config_variant.variant_name`.

model_analyzer/triton/client/client.py
```diff
@@ -74,7 +74,7 @@ def wait_for_server_ready(
             "Could not determine server readiness. " "Number of retries exceeded."
         )
 
-    def load_model(self, model_name):
+    def load_model(self, model_name, variant_name, config_str=None):
```
nv-hwoo, line 77: What is the purpose of `variant_name` here? Do we need this if it is only used for logging? 

I think we can at least make it optional so that we avoid a call that looks like 
```python
load_model("model_name", "model_name")
```

tests/test_triton_client.py
```diff
@@ -166,12 +166,12 @@ def test_load_unload_model(self):
 
         # Try to load a dummy model and expect error
         self.tritonclient_mock.raise_exception_on_load()
-        self.assertTrue(client.load_model("dummy"), -1)
+        self.assertTrue(client.load_model("dummy", "dummy"), -1)
```
nv-hwoo, line 169: Related to above comment

tests/test_triton_client.py
```diff
@@ -166,12 +166,12 @@ def test_load_unload_model(self):
...
 
         self.tritonclient_mock.reset()
 
         # Load the test model
-        client.load_model(TEST_MODEL_NAME)
+        client.load_model(TEST_MODEL_NAME, TEST_MODEL_NAME)
```
nv-hwoo, line 174: Related to above comment

model_analyzer/config/run/model_run_config.py
```diff
@@ -145,7 +145,8 @@ def representation(self) -> str:
         Returns a representation string for the ModelRunConfig that can be used
         as a key to uniquely identify it
         """
-        repr = self.perf_config().representation()
+        repr = self.model_variant_name()
+        repr += " " + self.perf_config().representation()
```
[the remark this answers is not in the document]
nv-hwoo: Makes sense 👍  Thanks for clarification!

model_analyzer/triton/client/client.py
nv-braf: Yes. Good point. I've made it optional.

nv-hwoo (approved):
I think `add_sub-results.csv` somehow slipped into the change. But othewise LGTM!  👍

PR #754, opened 2023-08-28
Exception if remote server is not in explicit mode

nv-braf:
The current behavior is that we print an info message to the user when the model fails to load, but in this specific case we should immediately terminate MA and tell the user exactly what is wrong.

There is no API call to determine what mode the server is in, so I'm string matching on the relevant part of the exception message being returned.

### DIFF
--- a/model_analyzer/triton/client/client.py
+++ b/model_analyzer/triton/client/client.py
@@ -105,6 +105,10 @@ class TritonClient:
             return None
         except Exception as e:
             logger.info(f"Model {variant_name} load failed: {e}")
+            if hasattr(e, "_msg") and "polling is enabled" in e._msg:
+                raise TritonModelAnalyzerException(
+                    "The remote Tritonserver needs to be launched in EXPLICIT mode"
+                )
             return -1
 
     def unload_model(self, model_name):


model_analyzer/triton/client/client.py
```diff
@@ -105,6 +105,10 @@ def load_model(self, model_name, variant_name="", config_str=None):
             return None
         except Exception as e:
             logger.info(f"Model {variant_name} load failed: {e}")
+            if hasattr(e, "_msg") and "polling is enabled" in e._msg:
```
nv-hwoo, line 108: Instead of accessing `self._msg`, I think we can use `message` method from `InferenceServerException` class:

```python
if "polling is enabled" in e.message():
    ...
```

https://github.com/triton-inference-server/client/blob/d257c0e5c3de6e15d6ef289ff2b96cecd0a69b5f/src/python/library/tritonclient/utils/__init__.py#L71

nv-braf: Done

model_analyzer/triton/client/client.py
```diff
@@ -105,6 +105,11 @@ def load_model(self, model_name, variant_name="", config_str=None):
             return None
         except Exception as e:
             logger.info(f"Model {variant_name} load failed: {e}")
+            if "polling is enabled" in e.message():
+                # if hasattr(e, "_msg") and "polling is enabled" in e._msg:
```
nv-hwoo, line 109: nit: you forgot to remove the comment

nv-braf: Removed!

nv-hwoo (approved):
LGTM!

Tabrizian (approved):

PR #756, opened 2023-08-28
Check if other models on server

nv-braf:
Check and print a warning message if there are models other than the one being profiled on the remote server.

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -116,6 +116,9 @@ class Analyzer:
         self._create_metrics_manager(client, gpus)
         self._create_model_manager(client, gpus)
 
+        if self._config.triton_launch_mode == "remote":
+            self._warn_if_other_models_loaded_on_remote_server(client)
+
         if self._config.model_repository:
             self._get_server_only_metrics(client, gpus)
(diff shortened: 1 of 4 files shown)


model_analyzer/analyzer.py
```diff
@@ -395,3 +398,15 @@ def _create_report_config(self, args: list) -> ConfigCommandReport:
         cli.add_subcommand(cmd="report", help="", config=config)
         cli.parse(args)
         return config
+
+    def _warn_if_other_models_loaded_on_remote_server(self, client):
```
nv-hwoo, line 402: From the meeting, I think I heard that this will no longer be a problem once out-proc (?) feature is implemented on the triton server side. Should we include a short docstring that says this warning can be removed once that is done?

nv-braf: No, that's not what was being said. It's the case where the loading/unloading of the same model will be fixed.
There will still be an issue if other models are loaded (and potentially running) on the server while a model is being profiled.

nv-hwoo: Thanks for clarification 👍

nv-hwoo (approved):

PR #757, opened 2023-08-31
Truncate model configuration name in plots

pskiran1:


model_analyzer/plots/simple_plot.py
```diff
@@ -184,6 +186,35 @@ def plot_data_and_constraints(self, constraints):
         self._ax.legend()
         self._ax.grid()
 
+    def truncate_model_config_name(self, model_config_name):
```
nv-braf, line 189: I can see this being used outside of simple plots. Can you make a new file called `report_utils.py` and add this?

nv-braf: We also need unit testing here, so I would add a test_report_utils.py and add testing.

nv-braf (changes requested):

model_analyzer/plots/simple_plot.py
pskiran1: Done the changes.

nv-braf (approved):
Looks great!

PR #759, opened 2023-09-06
Information in logs if Manual or Automatic config search is used

pskiran1:


nv-braf (commented):
We should also get rid of the generic "starting brute..." (in `brute_plus_parameter_search.py`)

nv-braf (approved):

PR #762, opened 2023-09-08
Move cpu_only to MCV class

nv-braf:
- Move `cpu_only` to the ModelConfigVariant class
- Remove backwards compatibility in checkpointing
- Updated all unit test checkpoints and switched the ensemble and BLS tests to use our new quick start models (which are easier to regenerate)
- Updated the L0 checkpoint and changed to use the `add_sub` model
- Updated the quick search L0's to use add_sub, which should (in theory) result in more consistency btw. runs on the CI


model_analyzer/plots/detailed_plot.py
```diff
@@ -168,7 +168,6 @@ def plot_data(self):
 
         # Sort the data by request rate or concurrency
         if "request_rate" in self._data and self._data["request_rate"][0]:
-            print(f"\n\nFound request rate: {self._data['request_rate']}\n\n")
```
nv-hwoo, line 171: Trying to understand the change. Is this not needed anymore?

nv-hwoo (commented):
One small comment but LGTM overall!

model_analyzer/plots/detailed_plot.py
nv-braf: That was a debug comment that somehow found got checked-in previously. Found it by inspection when making these changes.

nv-hwoo (approved):

PR #763, opened 2023-09-12
Remove requirement that remote mode have a model repository path

nv-braf:
Currently, in all modes we require a path to the model repository. This is no longer necessary for remote mode, as we are now loading the models via API with the tritonserver in explicit mode, therefore the tritonserver already needs to be pointed to the model directory (on the remote machine) and MA doesn't need access (only the model name).

This story also fixed/refactoring the error messages around creating the default configuration.

### DIFF
--- a/model_analyzer/triton/model/model_config.py
+++ b/model_analyzer/triton/model/model_config.py
@@ -88,35 +88,27 @@ class ModelConfig:
         try:
             config = ModelConfig._create_from_file(model_path).get_config()
         except Exception:
-            if (
-                config.triton_launch_mode == "docker"
-                or config.triton_launch_mode == "local"
-            ):
+            if ModelConfig._can_launch_mode_get_default_config_from_server(config):
                 config = ModelConfig._get_default_config_from_server(
-                    config, client, gpus, model_name, model_path
+                    config, client, gpus, model_name
                 )
             else:
-                if not os.path.exists(model_path):
-                    raise TritonModelAnalyzerException(
-                        f'Model path "{model_path}" specified does not exist.'
-                    )
-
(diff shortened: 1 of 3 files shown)


model_analyzer/triton/model/model_config.py
```diff
@@ -88,35 +88,27 @@ def create_model_config_dict(config, client, gpus, model_repository, model_name)
...
-                    f" Triton does not support default config creation for {config.triton_launch_mode} mode."
-                )
+                ModelConfig._check_default_config_exceptions(config, model_path)
 
         ModelConfig._default_config_dict[model_name] = config
         return deepcopy(config)
 
     @staticmethod
-    def _get_default_config_from_server(config, client, gpus, model_name, model_path):
+    def _can_launch_mode_get_default_config_from_server(config):
```
matthewkotila, line 102: Just out of curiosity, why write these as static methods rather than instance methods, especially when the instance is being passed in?

nv-braf: Since the parent method is static all methods the parent calls (like these refactoring methods) must also be static

matthewkotila: I see. Does the parent method also take the instance as a parameter? I think it'd be cleaner to have a pattern going forward of using instance methods if the static method takes an instance as a parameter.

Doesn't need to do anything for this PR but something to keep in mind.

matthewkotila (approved):

model_analyzer/triton/model/model_config.py
nv-braf: No, it doesn't, which is why all the refactoring methods need to be static.

PR #768, opened 2023-09-29
Adding new options for LLM

nv-braf:
Added the following new options to support LLM:

- --llm-search-enable
- --run-config-search-min/max-prompt-length
- --run-config-search-min/max-token-count
- --prompt-length
- --token-count


model_analyzer/config/input/config_command_profile.py
```diff
@@ -36,62 +37,6 @@
 from model_analyzer.triton.server.server_config import TritonServerConfig
 
 from .config_command import ConfigCommand
-from .config_defaults import (
```
nv-braf, line 39: It was getting annoying (and ugly) to keep adding more defaults. So, I switched this to import all the defaults

tests/test_cli.py
```diff
@@ -71,23 +72,27 @@ def get_test_options():
         # The following options can be None:
         #   short_option
         #   expected_default_value
-        OptionStruct("int", "profile", "--client-max-retries", "-r", "125", "50"),
```
nv-braf, line 74: Boyscoutted the removal of magic numbers

Dockerfile
```diff
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:23.08-py3
```
nv-braf, line 15: I don't know why these changes are showing up after I rebased, but these and the docs changes are safe to ignore.

model_analyzer/config/input/config_defaults.py
```diff
@@ -51,12 +51,17 @@
 DEFAULT_RUN_CONFIG_MIN_MODEL_BATCH_SIZE = 1
 DEFAULT_RUN_CONFIG_MAX_MODEL_BATCH_SIZE = 128
 DEFAULT_RUN_CONFIG_MAX_BINARY_SEARCH_STEPS = 5
+DEFAULT_RUN_CONFIG_MIN_PROMPT_LENGTH = 1
+DEFAULT_RUN_CONFIG_MAX_PROMPT_LENGTH = 1000
+DEFAULT_RUN_CONFIG_MIN_TOKEN_COUNT = 1
+DEFAULT_RUN_CONFIG_MAX_TOKEN_COUNT = 256
```
nv-hwoo, line 57: General question: What is MA trying to do by searching through different prompt sizes and token counts (aka. max_tokens)? Is it trying to find a good prompt size and token count?

Dockerfile
nv-hwoo: Maybe you can try this to remove the unwanted commit from PR: https://stackoverflow.com/a/51400593

nv-braf: I'm not too concerned. They match what is on the `add-llm-mode`  branch and will go away when this subbranch is pushed. Just wanted to call it out so you wouldn't review it.

model_analyzer/config/input/config_defaults.py
nv-braf: Yes, MA will sort the results by avg. first token-to-token latency in a table. But, the real benefit will be the graph that shows these values on the x-axis and avg. first token-to-token latency on the y-axis.

nv-hwoo (approved):

PR #774, opened 2023-10-10
Capture LLM metrics from PA

nv-braf:
These are the changes needed in the metric manager to capture the timestamps coming back from PA and then convert the timestamps into the first/avg-token-to-token latency metrics.

Changes also made to perf_analyzer to add in the new argument for passing back the timestamp file to MA.

Unit tests were updated/added to ensure correctness of these changes.


model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -457,21 +480,86 @@ def _parse_outputs(self, metrics):
...
+        )
+
+        return avg_first_token_to_token_latency
+
+    def _calculate_avg_token_to_token_latency(self, llm_output: Dict) -> float:
+        total_token_latency = 0.0
+        for request in llm_output["experiments"][0]["requests"]:
+            total_response_latency = 0
+            for response_timestamp in request["response_timestamps"]:
+                total_response_latency += response_timestamp - request["timestamp"]
```
nv-hwoo, line 552: The T2T latency calculation should be the difference between the two consecutive responses, not response - request timestamp. We can use `itertools.pairwise`:

```suggestion
            for prev_response, response in pairwise(request["response_timestamps"]):
                total_response_latency += response - prev_response
```

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -216,6 +226,19 @@ def get_gpu_records(self):
...
+        """
+        Returns
+        -------
+        The LLM records from the last perf_analyzer run
+        """
+
+        if self._llm_records:
+            return self._llm_records
+        raise TritonModelAnalyzerException(
+            "Attempted to get perf_analyzer results" "without calling run first."
```
nv-hwoo, line 239:
```suggestion
            "Attempted to get perf_analyzer results without calling run first."
```

model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -457,21 +480,86 @@ def _parse_outputs(self, metrics):
...
+        return avg_first_token_to_token_latency
+
+    def _calculate_avg_token_to_token_latency(self, llm_output: Dict) -> float:
+        total_token_latency = 0.0
+        for request in llm_output["experiments"][0]["requests"]:
+            total_response_latency = 0
+            for response_timestamp in request["response_timestamps"]:
+                total_response_latency += response_timestamp - request["timestamp"]
+
+            total_token_latency += total_response_latency / len(
```
nv-hwoo, line 554: I don't think `total_token_latency` is accurate name here since it is "average" of single request responses/tokens. But seems like `avg_token_to_token_latency` is also taken.

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: Understand and I've made the fix. However, itertools.pairwise isn't available, so I've had to code this the old-fashioned way.

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: Changed this to be `total_reponse_to_response...` and `total_token_to_token...`

tests/common/test_utils.py
```diff
@@ -253,6 +254,8 @@ def construct_perf_analyzer_config(
         The name of the model
     output_file_name: str
         The name of the output file
+    output_llm_file_name: str
+        The name of the LLM output file
```
nv-hwoo, line 258: I would rather call it export file than llm file since export file is not restricted to llm models.
```suggestion
    export_file_name: str
        The name of the profile export file
```

nv-hwoo (approved):

PR #777, opened 2023-10-17
Correct how periodic concurrency works in PACG

nv-braf:
I've changed the PA config generator to correctly reflect how periodic concurrency is supposed to work (start:stop:range). 

This incorporates all the new CLI/config fields and updates unit testing to check that the correct number (and value) of configs are generated when periodic concurrency is searched.

I've also updated the default search values to keep the number of searches under 100. The algorithm will search by powers of 2 (just like how the other searches work in brute mode).


model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -247,16 +248,52 @@ def _create_concurrency_list(self) -> List[int]:
...
+                    ):
+                        continue
+
+                    periodic_concurrencies.append(
+                        f"{min_periodic_concurrency}:{max_periodic_concurrency}:{step}"
+                    )
+        return periodic_concurrencies
+
+    def _is_illegal_periodic_concurrency_combination(
+        self, min: int, max: int, step: int
```
nv-hwoo, line 287:
```suggestion
        self, start: int, end: int, step: int
```

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -247,16 +248,52 @@ def _create_concurrency_list(self) -> List[int]:
...
+            self._cli_config.run_config_search_max_periodic_concurrency,
+        )
+
+        step_doubled_list = utils.generate_doubled_list(
+            self._cli_config.run_config_search_min_periodic_concurrency_step,
+            self._cli_config.run_config_search_max_periodic_concurrency_step,
+        )
+
+        for min_periodic_concurrency in periodic_concurrency_doubled_list:
+            for max_periodic_concurrency in periodic_concurrency_doubled_list:
```
nv-hwoo, line 274: Since we are looking for a valid `<start:end:step>` combination, I think we it would be more accurate to call them `start` and `end` instead of min and max.

```suggestion
        for start in periodic_concurrency_doubled_list:
            for end in periodic_concurrency_doubled_list:
```

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -247,16 +248,52 @@ def _create_concurrency_list(self) -> List[int]:
                 self._cli_config.run_config_search_max_concurrency,
             )
 
-    def _create_periodic_concurrency_list(self) -> List[int]:
+    def _create_periodic_concurrency_list(self) -> List[str]:
         if self._model_parameters["periodic_concurrency"]:
             return sorted(self._model_parameters["periodic_concurrency"])
```
nv-hwoo, line 253: Does this also return `List[str]`?

model_analyzer/config/generate/perf_analyzer_config_generator.py
```diff
@@ -247,16 +248,52 @@ def _create_concurrency_list(self) -> List[int]:
...
+    def _is_illegal_periodic_concurrency_combination(
+        self, min: int, max: int, step: int
+    ) -> bool:
+        if min > max:
+            return True
+        elif max == min and step != 1:
+            return True
+        elif (max - min) % step:
+            return True
+
```
nv-hwoo, line 295: Remove whitespace

model_analyzer/config/input/config_list_numeric.py
```diff
@@ -103,7 +103,14 @@ def set_value(self, value):
         try:
             if self._is_string(value):
                 self._value = []
-                value = value.split(",")
+                if "," in value:
+                    value = value.split(",")
+                elif ":" in value:
```
nv-hwoo, line 108: Is there reason why we can't use comma as a separator for periodic concurrency? Seems like it was already using comma for other inference loads

nv-braf: This is how it's documented in PA:
```--periodic-concurrency-range <start:end:step>```

Just wanted to keep us consistent.

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: Yes `self._model_parameters["periodic_concurrency"]` returns a List[str]

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: Fixed

model_analyzer/config/generate/perf_analyzer_config_generator.py
nv-braf: Agreed. I've made the change

model_analyzer/config/input/config_list_numeric.py
nv-hwoo: Is the `ConfigListNumeric` class still used for periodic concurrency? I see that periodic concurrency is using `ConfigListString`

nv-braf: No, it's using `ConfigListString`. The change in `ConfigListNumeric` is for non-periodic concurrency. So, the user can specify <start:end:step> without needing to create a dictionary: `--concurrency 5:50:5`

For periodic concurrency they specify a string like: `"5:50:5", "10:100:10"` which can contain multiple sets of <start:stop:end> (vs. ConfigListNumeric which only allows a single set).

PR #781, opened 2023-10-31
Adding illegal LLM checks w/ unit testing + some minor cleanup

nv-braf:
Add in checks for illegal combination of options when running an LLM model through MA.

### DIFF
--- a/model_analyzer/config/input/config_command_report.py
+++ b/model_analyzer/config/input/config_command_report.py
@@ -209,6 +209,10 @@ class ConfigCommandReport(ConfigCommand):
 
         super().set_config_values(args)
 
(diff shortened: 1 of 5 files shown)


model_analyzer/config/input/config_command_report.py
```diff
@@ -209,6 +209,10 @@ def set_config_values(self, args):
 
         super().set_config_values(args)
 
+    # TODO: Update this when adding support for detailed reporting
```
debermudez, line 212: Do we no longer tag these with ticket numbers?

nv-braf: I've added the ticket number.

debermudez (approved):

PR #784, opened 2023-11-02
Output config.pbtxt when in remote mode

nv-braf:
Currently we are not outputting anything about the configuration when we are in remote mode (as we don't have access to the model). This change is to output the `config.pbtxt` variant for each configuration MA creates.

### DIFF
--- a/model_analyzer/record/metrics_manager.py
+++ b/model_analyzer/record/metrics_manager.py
@@ -347,35 +347,74 @@ class MetricsManager:
         repository and fills directory with config
         """
 
-        variant_name = variant_config.variant_name
         if self._config.triton_launch_mode != "remote":
-            model_repository = self._config.model_repository
+            self._create_non_remote_mode_model_variant(
+                original_name, variant_config, ignore_first_config_variant
+            )
+        else:
+            self._create_remote_mode_model_variant(original_name, variant_config)
 
-            original_model_dir = os.path.join(model_repository, original_name)
-            new_model_dir = os.path.join(self._output_model_repo_path, variant_name)
-            try:
-                # Create the directory for the new model
-                os.makedirs(new_model_dir, exist_ok=True)
-                self._first_config_variant.setdefault(original_name, None)
+    def _create_non_remote_mode_model_variant(
+        self,
+        original_name: str,
+        variant_config: ModelConfigVariant,
+        ignore_first_config_variant: bool = False,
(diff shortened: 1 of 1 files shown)


model_analyzer/record/metrics_manager.py
```diff
@@ -347,35 +347,74 @@ def _create_model_variant(
...
+        original_model_dir = os.path.join(model_repository, original_name)
+        new_model_dir = os.path.join(self._output_model_repo_path, variant_name)
+        try:
+            # Create the directory for the new model
+            os.makedirs(new_model_dir, exist_ok=True)
+            self._first_config_variant.setdefault(original_name, None)
+
+            if ignore_first_config_variant:
+                variant_config.model_config.write_config_to_file(
+                    new_model_dir, original_model_dir, None
```
tgerdesnv, line 379: I'm obviously very rusty, so bear with me. Is this "ignore" still creating directories and writing out an empty file?

model_analyzer/record/metrics_manager.py
```diff
@@ -347,35 +347,74 @@ def _create_model_variant(
         repository and fills directory with config
         """
 
-        variant_name = variant_config.variant_name
         if self._config.triton_launch_mode != "remote":
-            model_repository = self._config.model_repository
+            self._create_non_remote_mode_model_variant(
+                original_name, variant_config, ignore_first_config_variant
```
tgerdesnv, line 352: What does ignore_first_config_variant do?

what is the difference in the code between remote vs non-remote? As far as I can tell the only difference is the use of ignore_first_config_variant. Is that true?

model_analyzer/record/metrics_manager.py
nv-braf: It gets passed through multiple methods and eventually has an effect in `model_config.py::write_config_to_file`, but I haven't changed the behavior around this logic.

model_analyzer/record/metrics_manager.py
nv-braf: See my above comment. It's not relevant for remote mode because there is no actual model to write into the directory. The difference btw. remote and non-remote is that non-remote will have the `config.pbtxt` along with the model; remote mode only has the `config.pbtxt` (since the model is on a remote machine and MA doesn't have access to it).

tgerdesnv (approved):

Issue #800, opened 2023-12-20
Client batch size for quick search method


TheConstant3:
Is it possible to set specific client batch size for all or one profiling models?
While quick search client batch size always starts from 1, but my models will process batches more then 32. How can i test my models with batch size > 32 only?

tgerdesnv:
Unfortunately, at this time setting the client batch size with quick search is not supported.

One option is to use [Brute Search](https://github.com/triton-inference-server/model_analyzer/blob/main/docs/config_search.md#brute-search-mode) instead of quick search. This will take more time but will test all desired combinations.

Another option that may work is using quick search with setting `--run_config_search_min_model_batch_size=32`. That would still send client requests one at a time, but the testing would be done with high enough concurrency such that Triton should always be able to batch 32 or more together on the server side.

Another option would be to wait for us to add support for this. Adding support for a single static client batch size doesn't look like it would be too hard

lzcchl:
Yes, I also have this requirement. When evaluating the ensemble model using model analyzer, the maximum throughput obtained is 120. However, when I use perf analyzer to execute the model, I can get a maximum throughput of 220. After comparison, I found that the parameter - b of perf analyzer is in effect. 

However, when I set batch sizes in model analyzer, the log I see still shows client batch size=1, which makes me very confused, Because I am using version R23.12, does the 24.01 version of the model analyzer support modifying the batch size for quick search.

tgerdesnv:
As of 23.12 (and will be still true for 24.01):
- Quick search still only supports a fixed batch size of 1. I have a draft PR up to support a different (but still fixed) batch size. I'm waiting on a team lead to come back from a medical procedure to review it.
- Brute search only supports a single model. The search time exponentially grows to days and weeks for brute multi-model, so there is no plan to add support for brute multi-model
- There are some known issues around non-1 batch sizes in the reports

There is an effort in place to significantly improve quick search to be able to support searching any args (including batch size). This is probably still a couple months away.

PR #802, opened 2023-12-20
Proposed changes to support a single client batch size for quick search

tgerdesnv:
Potential fix for https://github.com/triton-inference-server/model_analyzer/issues/800


the-david-oy:
Closing, as this effort is no longer being developed for now. Optuna search is now supported to provide similar functionality.

Issue #791, opened 2023-11-24
Illegal configuration ensemble model, when a config.pbtxt of a submodel is auto generated


SDJustus:
Hi folks,

trying out the model analyzer I noticed, that illegal configurations occur, when autogenerating a config for an ONNX-Model. 
This happens due to the max_batch_size, which is set to 1 form my ensemble model and is set to 4 (along with the prefered_batch_size) for the ONNX Model by Triton, since this is the default for this backend...

Is it expected behaviour for the configs generated by the model analyzer to be illegal in this case, since this configuration is deployable on Triton and works?

SDJustus:
I think this is due to the model_analyzer trying to set the max_batch_size of the submodel to 1 while there is an dynamic_batching entry autogenerated.
```
dynamic_batching {
  preferred_batch_size: 4
}
```

tgerdesnv:
What you are describing makes sense and sounds like a bug. There is code that tries to avoid illegal combinations, but something around autogeneration of configs and/or interaction with ensembles may be misbehaving. 

It would be helpful to have a simple repro of the issue. I don't need a model, but any configs or the CLI options you sent into MA would help me to reproduce and fix the issue.

SDJustus:
I'll get a minimal example tomorrow. :)

SDJustus:
[ensemble-model.zip](https://github.com/triton-inference-server/model_analyzer/files/13526078/ensemble-model.zip)

SDJustus:
Thanks for digging further into it. Really appreciate the effort.

PR #810, opened 2024-01-11
Fix loading of composing models for ensembles

tgerdesnv:
Before this change, there was no way for Model Analyzer to load ensemble composing models via the API. This meant that they all defaulted to auto-complete configs. Now (with a required change to Triton) they can be loaded and won't be overwritten when the main ensemble config is loaded.

Fixes https://github.com/triton-inference-server/model_analyzer/issues/791
Needs this change to work: https://github.com/triton-inference-server/core/pull/304

### DIFF
--- a/model_analyzer/record/metrics_manager.py
+++ b/model_analyzer/record/metrics_manager.py
@@ -430,19 +430,15 @@ class MetricsManager:
         """
         Loads all model variants in the client
         """
-        # TODO TMA-1487: Make BLS and ensemble both load all composing model variants first
         for mrc in run_config.model_run_configs():
+            # Load all composing model variants first, and then the parent model
+            for composing_config_variant in mrc.composing_config_variants():
+                if not self._load_model_variant(
+                    variant_config=composing_config_variant
+                ):
+                    return False
             if not self._load_model_variant(variant_config=mrc.model_config_variant()):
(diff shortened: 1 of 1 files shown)


model_analyzer/record/metrics_manager.py
```diff
@@ -430,19 +430,15 @@ def _load_model_variants(self, run_config: RunConfig) -> bool:
         """
         Loads all model variants in the client
         """
-        # TODO TMA-1487: Make BLS and ensemble both load all composing model variants first
         for mrc in run_config.model_run_configs():
+            # Load all composing model variants first, and then the parent model
+            for composing_config_variant in mrc.composing_config_variants():
```
debermudez, line 435: Why are we no longer checking if mrc is a bls model?

debermudez (approved):
Discussed offline. This is good to go

model_analyzer/record/metrics_manager.py
tgerdesnv: We need to always load composing models now, including for ensembles, since we go through the API and don't want default configs to be loaded.

debermudez: Conclusion: all models need to load sub models first so the bls check is pointless now.

PR #834, opened 2024-03-11
 Update README and versions for 1.38.0 / 24.03

mc-nv:


nv-kmcgill53 (approved):
Thanks for covering me

PR #839, opened 2024-03-18
New Records for LLM metrics

nv-braf:
I've added Records to MA for the new LLM metrics:

- Inter Token Latency: min, max, avg, p75, p90, p99
- Time to First Token Latency: min, max, avg, p75, p90, p99
- Output Token Throughput


model_analyzer/record/types/inter_token_latency_avg.py
```diff
@@ -0,0 +1,96 @@
...
+from model_analyzer.record.record import DecreasingRecord
+
+
+@total_ordering
+class InterTokenLatencyAvg(DecreasingRecord):
+    """
+    A record for perf_analyzer Inter token latency metric
+    """
+
+    tag = "inter_token_latency_avg"
```
tgerdesnv, line 28: Is there not a way to code these up such that we only need one base class and 6 derived classes for InterTokenLatency?

nv-braf: There probably is and that would save us a little bit of code since the __lt__, __eq__, __add__, __sub__ wouldn't need to be replicated, but everything else in the class is still required or am I missing something?

tgerdesnv: I think in my mind this was also going to collapse the number of files and comments, and that it would end up parameterized where the tag and header were passed in to a template or something. I'm fine with how this is though.

tgerdesnv (approved):

PR #840, opened 2024-03-18
Move MA concurrency and request-rate check to autofill

### LINKED ISSUE #838: Encountered an error: "Cannot specify both concurrency and request rate as model parameters." while not specifying `concurrency` at all.`
I've encountered an error: "Cannot specify both concurrency and request rate as model parameters." while not specifying `concurrency` at all.

Here is my config for model analyzer:

```yaml
run_config_search_disable: true
run_config_profile_models_concurrently_enable: false
request_rate_search_enable: true

triton_launch_mode: docker

profile_models:
  yolo-10class:
    parameters:
      request_rate: [240, 264, 300, 360]
      batch_sizes: [1]
    model_config_parameters:
      max_batch_size: [4, 14, 16, 20, 32]
      dynamic_batching:
        max_queue_delay_microseconds: [0, 15000, 20000]
      instance_group:
      - kind: KIND_GPU
        count: [1, 2]

objectives:
  - perf_latency_p99

num_configs_per_model: 20

perf_analyzer_flags:
  shared-memory: cuda

triton_server_flags:
  log_verbose: true

```
I'm running it with a command: 

```bash
model-analyzer --verbose profile --gpus=all \
      --config-file /config/model-analyzer/yolo.yaml \
      --model-repository ${PWD}/.triton-repository \
      --output-model-repository-path /analyzer-logs/iteration \
      --export-path /profile_results
```

I've been able to track that error raises here https://github.com/triton-inference-server/model_analyzer/blob/16aadb922daaea8f2cdab3b4f9d645744a72b744/model_analyzer/config/input/config_command_profile.py#L1434-L1440

print-debug gives me
```python
>>> new_model["parameters"]["concurrency"]
[0]
```

Which is not what is expected to be if you specify only `request_rate`, right?

dyastremsky:
During preprocessing, Model Analyzer checks whether request rate and concurrency are both missing if "run config search" is disabled. However, self.request_rate_ and self.concurrency_ are not set in the code, so the code always sets a concurrency value of 1. This results in an error if request-rate is provided with "run config search" disabled.

Move this check later to when the provided parameters are being checked and default values are set. Use the user-provided values and default values to determine whether concurrency and request-rate are set before using a default value for concurrency.

Fixes #838.

### DIFF
--- a/model_analyzer/config/input/config_command_profile.py
+++ b/model_analyzer/config/input/config_command_profile.py
@@ -1424,7 +1424,7 @@ class ConfigCommandProfile(ConfigCommand):
                 elif not "request_rate" in model.parameters():
                     new_model["parameters"].update({"concurrency": self.concurrency})
                 else:
-                    new_model["parameters"].update({"concurrency": None})
+                    new_model["parameters"].update({"concurrency": []})
 
                 if "request_rate" in model.parameters():
                     new_model["parameters"].update(
--- a/model_analyzer/config/input/config_command_profile.py
+++ b/model_analyzer/config/input/config_command_profile.py
@@ -1421,8 +1421,10 @@ class ConfigCommandProfile(ConfigCommand):
                     new_model["parameters"].update(
                         {"concurrency": model.parameters()["concurrency"]}
                     )
-                else:
+                elif not "request_rate" in model.parameters():
                     new_model["parameters"].update({"concurrency": self.concurrency})
+                else:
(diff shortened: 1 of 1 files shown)


dyastremsky:
@nv-braf, this is now ready for review. When run locally with [this config file](https://github.com/triton-inference-server/model_analyzer/issues/838) with a toy model, it runs fine. CI testing has passed as well.

nv-braf (approved):
Looks good!

PR #842, opened 2024-03-19
Add --llm-mode option

nv-braf:
Adds new option `--llm-mode` which will be used by code in MA to get LLM specific metrics from PA and create LLM style reporting


docs/config.md
```diff
@@ -236,6 +236,9 @@ cpu_only_composing_models: <comma-delimited-string-list>
 # Skips the generation of detailed reports and tables
 [ skip_detailed_reports: <bool> | default: false]
 
+# Informs MA that we are profiling a LLM
+[ llm_mode: <bool> | default: false]
```
tgerdesnv, line 240: Shouldn't all of this be going into a side branch? Otherwise there will be a new arg in main that does nothing?

model_analyzer/config/input/config_command_profile.py
```diff
@@ -297,6 +298,16 @@ def _fill_config(self):
                 description="Skips the generation of detailed summary reports and tables.",
             )
         )
+        self._add_config(
+            ConfigField(
+                "llm_mode",
+                flags=["--llm-mode"],
```
tgerdesnv, line 304: I worry about adding a bunch of args in the future -- for other modes like embedding mode or video mode or something. I know that genai-perf was designed to be generic -- it is currently LLMs but the plan is to eventually expand it. I don't think they added an arg to specify the mode yet in genai-perf, but I feel like we should have a flag that takes in the mode instead of a bunch of boolean modes where only one can be set.

docs/config.md
nv-braf: Yes. I was not merging to the correct base.

model_analyzer/config/input/config_command_profile.py
nv-braf: I'm fine with that. I've changed the option to be `--model-type` with `generic` or `LLM` as options.

tgerdesnv (approved):

PR #849, opened 2024-03-26
Profiling model using genai-perf

nv-braf:
Successfully profiled and checkpointed a a vLLM run through MA using genai-perf. Here is the resulting table output:

```
Models (Inference):
Model       Batch   Concurrency   Model Config Path          Instance Group   Max Batch Size   Satisfies Constraints   Throughput (infer/sec)   p99 Latency (ms)   p99 Inter Token Latency (ms)   p99 Time To First Token (ms)   Output Token Throughput (infer/sec)  
gpt2_vllm   1       1             gpt2_vllm_config_default   1:MODEL          0                Yes                     8.9                      117.0              11.3                           47704.6                        76192.3                              

Models (GPU Metrics):
Model       GPU UUID                                   Batch   Concurrency   Model Config Path          Instance Group   Satisfies Constraints   GPU Memory Usage (MB)   GPU Utilization (%)   GPU Power Usage (W)  
gpt2_vllm   GPU-8557549f-9c89-4384-8bd6-1fd823c342e0   1       1             gpt2_vllm_config_default   1:MODEL          Yes                     12768.5                 20.4                  65.2                 

Server Only:
Model           GPU UUID                                   GPU Memory Usage (MB)   GPU Utilization (%)   GPU Power Usage (W)  
triton-server   GPU-8557549f-9c89-4384-8bd6-1fd823c342e0   12768.0                 0.5                   42.1                 
```

### DIFF
--- a/model_analyzer/perf_analyzer/perf_analyzer.py
+++ b/model_analyzer/perf_analyzer/perf_analyzer.py
@@ -115,15 +115,15 @@ class PerfAnalyzer:
     ]
 
     llm_metric_table = [
-        ["time_to_first_token_avg", "Time to First Token (ns) avg",          TimeToFirstTokenAvg, "1000"],
-        ["time_to_first_token_min", "Time to First Token (ns) min",          TimeToFirstTokenMin, "1000"],
-        ["time_to_first_token_max", "Time to First Token (ns) max",          TimeToFirstTokenMax, "1000"],
-        ["time_to_first_token_p99", "Time to First Token (ns) p99",          TimeToFirstTokenP99, "1000"],
-        ["time_to_first_token_p95", "Time to First Token (ns) p95",          TimeToFirstTokenP95, "1000"],
-        ["time_to_first_token_p90", "Time to First Token (ns) p90",          TimeToFirstTokenP90, "1000"],
-        ["time_to_first_token_p75", "Time to First Token (ns) p75",          TimeToFirstTokenP75, "1000"],
-        ["time_to_first_token_p50", "Time to First Token (ns) p50",          TimeToFirstTokenP50, "1000"],
-        ["time_to_first_token_p25", "Time to First Token (ns) p25",          TimeToFirstTokenP25, "1000"],
+        ["time_to_first_token_avg", "Time To First Token (ns) avg",          TimeToFirstTokenAvg, "1000"],
+        ["time_to_first_token_min", "Time To First Token (ns) min",          TimeToFirstTokenMin, "1000"],
+        ["time_to_first_token_max", "Time To First Token (ns) max",          TimeToFirstTokenMax, "1000"],
+        ["time_to_first_token_p99", "Time To First Token (ns) p99",          TimeToFirstTokenP99, "1000"],
+        ["time_to_first_token_p95", "Time To First Token (ns) p95",          TimeToFirstTokenP95, "1000"],
+        ["time_to_first_token_p90", "Time To First Token (ns) p90",          TimeToFirstTokenP90, "1000"],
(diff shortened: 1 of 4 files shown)


model_analyzer/perf_analyzer/perf_analyzer.py
```diff
@@ -323,14 +323,33 @@ def _get_cmd(self):
...
-        cmd = [self.bin_path]
-        if self._is_multi_model():
-            cmd += ["--enable-mpi"]
-        cmd += self._get_pa_cli_command(index).replace("=", " ").split()
+        if self._model_type == "LLM":
+            cmd = [
+                "genai-perf",
+                "-m",
+                self._config.models_name(),
+                "--streaming",
```
tgerdesnv, line 331: not sure that we want to hardcode streaming. Should at least create a story to hook this up to something, either the online vs offline mode or some new LLM flag

model_analyzer/perf_analyzer/perf_config.py
```diff
@@ -340,19 +342,22 @@ def to_cli_string(self):
...
 
-    def _parse_short_options(self):
+    def _parse_short_options(self, exclude_model_name: bool = False) -> List:
         """
         Parse the perf analyzer single dash options
         """
         temp_args = []
         for key, value in self._options.items():
             if value:
+                if exclude_model_name and key == "-m":
```
tgerdesnv, line 358: I see you checking short options. Could model-name be in the long options?

model_analyzer/record/metrics_manager.py
```diff
@@ -589,9 +611,10 @@ def _run_perf_analyzer(
             max_retries=self._config.perf_analyzer_max_auto_adjusts,
             timeout=self._config.perf_analyzer_timeout,
             max_cpu_util=self._config.perf_analyzer_cpu_util,
+            model_type=self._config.model_type,
         )
 
-        metrics_to_gather = self._perf_metrics + self._gpu_metrics
+        metrics_to_gather = self._perf_metrics + self._llm_metrics + self._gpu_metrics
```
tgerdesnv, line 617: will this always try to collect llm_metrics even when the model type isn't LLM?

model_analyzer/record/metrics_manager.py
```diff
@@ -601,6 +624,12 @@ def _run_perf_analyzer(
             return (None, None)
 
         perf_records = perf_analyzer.get_perf_records()
+
+        if self._config.model_type == "LLM":
```
tgerdesnv, line 628: I would think all of this would be handled inside of get_perf_records(). That it would look to see if LLM or not and extend what it returns.

tgerdesnv (commented):
Great job! Awesome to see it able to run and print out results.

model_analyzer/perf_analyzer/perf_config.py
nv-braf: No, model name (`-m`) is something MA adds in when it creates the PA config.

model_analyzer/perf_analyzer/perf_analyzer.py
nv-braf: I have a story to track this change as part of adding all the genai-perf CLI options.

model_analyzer/record/metrics_manager.py
nv-braf: In this line, it will always try to add them, but if the model type is not LLM then `self._llm_metrics` will be blank.

model_analyzer/record/metrics_manager.py
nv-braf: I thought about doing that. It's ugly no matter where this step happens. This will go away once we switch from using CSVs to an API.

tgerdesnv (approved):

PR #868, opened 2024-04-30
Class to hold info about parameters

nv-braf:
As part of the scaffolding for the new search we need to more cleanly define the parameters the user wants to search. A new dataclass (ConfigParameter) and the accessor class (ConfigParameters) have been created to cleanly handle user defined parameters.

Note: Usage of this in MA will be in a follow-on story

### DIFF
--- a/model_analyzer/config/generate/config_parameter.py
+++ b/model_analyzer/config/generate/config_parameter.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from enum import Enum, auto
+from typing import Any, List, Optional
+
+
+class ParameterType(Enum):
+    MODEL = auto()
+    RUNTIME = auto()
+    BUILD = auto()
+
+
+class ParameterCategory(Enum):
+    INTEGER = auto()
+    EXPONENTIAL = auto()
+    LIST = auto()
+
+
+@dataclass
+class ConfigParameter:
+    """
+    A dataclass that holds information about a configuration parameter
+    """
+
+    ptype: ParameterType
+    category: ParameterCategory
+
+    # These are only applicable to INTEGER and EXPONENTIAL categories
+    min_range: Optional[int] = None
+    max_range: Optional[int] = None
+
+    # This is only applicable to LIST category
+    enumerated_list: List[Any] = []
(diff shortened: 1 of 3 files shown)


model_analyzer/config/generate/config_parameters.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def _check_for_illegal_list_input(
+        self,
+        min_range: Optional[int],
+        max_range: Optional[int],
+        enumerated_list: List[Any],
+    ) -> None:
+        if not enumerated_list:
+            raise TritonModelAnalyzerException(
+                f"enumerated_list must be specified for a LIST"
```
nv-hwoo, line 91: Is this typo?
```suggestion
                f"enumerated_list must be specified for a list"
```

model_analyzer/config/generate/config_parameters.py
```diff
@@ -0,0 +1,100 @@
...
+        return self._parameters[name].ptype
+
+    def get_category(self, name: str) -> ParameterCategory:
+        return self._parameters[name].category
+
+    def get_range(self, name: str) -> Tuple[int, int]:
+        return (self._parameters[name].min_range, self._parameters[name].max_range)
+
+    def get_list(self, name: str) -> List[Any]:
+        return self._parameters[name].enumerated_list
```
nv-hwoo, line 60: Should these be handled by ConfigParameters object? I feel like we just need to provide a single accessor to the underlying parameter by name and the users can use ConfigParameter object to get whatever they need with that parameter (since it's a dataclass).

model_analyzer/config/generate/config_parameters.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def add_parameter(
+        self,
+        name: str,
+        ptype: ParameterType,
+        category: ParameterCategory,
+        min_range: Optional[int] = None,
+        max_range: Optional[int] = None,
+        enumerated_list: List[Any] = [],
+    ):
```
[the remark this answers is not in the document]
nv-braf: Fixed

model_analyzer/config/generate/config_parameters.py
nv-braf: Fixed

model_analyzer/config/generate/config_parameters.py
nv-braf: I'm not sure. Maybe this is overkill and never gets used, but I figured it couldn't hurt to add these.

nv-hwoo: IMO I don't think it will be late to add these later, but I'll leave it up to you to make the final call.

nv-braf: There in and tested at this point, so it'd be more work to rip them out. Going to leave it as is and merge.

model_analyzer/config/generate/config_parameter.py
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from enum import Enum, auto
+from typing import Any, List, Optional
+
+
+class ParameterType(Enum):
```
dyastremsky, line 22: SearchUsageType

model_analyzer/config/generate/config_parameter.py
```diff
@@ -0,0 +1,48 @@
...
+    MODEL = auto()
+    RUNTIME = auto()
+    BUILD = auto()
+
+
+class ParameterCategory(Enum):
```
dyastremsky, line 28: ParameterType

PR #869, opened 2024-05-03
Connect up parameter description class

nv-braf:
The objective is to capture all user search requirements into a class (called ConfigParameters) to standardize this before going to whatever search algorithm we end up implementing.

Incorporates the new ConfigParameters class into the CLI.
The the unit testing for the new mehtods is done in `test_config` as it's easier to test there, than to bring all the code into a new testbench to handle paring CLI/YAML config files.

**RENAMED class to SearchParameters**


model_analyzer/config/input/config_command.py
```diff
@@ -289,10 +289,10 @@ def _check_quick_search_model_config_parameters_combinations(self) -> None:
         if not "profile_models" in config:
             return
 
-        if config["run_config_search_mode"] != "quick":
```
nv-braf, line 292: These are bugs I found via inspection

model_analyzer/config/generate/config_parameters.py
```diff
@@ -53,12 +54,146 @@ def get_type(self, name: str) -> ParameterType:
...
+            self._add_parameter(
+                name=parameter_name,
+                ptype=ptype,
+                category=ParameterCategory.LIST,
+                enumerated_list=parameter_list,
+            )
+        else:
+            category = self._determine_parameter_category(parameter_name)
+
+            if category == ParameterCategory.EXPONENTIAL:
```
dyastremsky, line 110: This logic appears to be very trusting that things are done right elsewhere. Do we want to add safeguards?

If a parameter list is passed, do the RCS values get silently ignored? I thought all four parameters (including parameter_list) are passed elsewhere, which means the RCS values would be ignored. If a parameter list is not passed, it feels like the error is a bit misleading, since what if it's not passed but the parameter isn't supposed to be exponential/linear. I think some refactoring needs to be done here. I think this can exist as one function, but with increased documentation to understand what is happening and safeguarding to prevent logic errors.

model_analyzer/config/generate/config_parameters.py
```diff
@@ -53,12 +54,146 @@ def get_type(self, name: str) -> ParameterType:
...
+            parameter_name="concurrency",
+            rcs_min_value=config.run_config_search_min_concurrency,
+            rcs_max_value=config.run_config_search_max_concurrency,
+            parameter_list=parameters["concurrency"],
+        )
+
+    def _populate_parameter(
+        self,
+        parameter_name: str,
+        rcs_min_value: Optional[int] = None,
```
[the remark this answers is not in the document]
dyastremsky: Also, a nit, but I would love for the params to go parameter_name, parameter_list, rcs_min_value, rcs_max_value. It feels like parameters are a bit more logically grouped there, but this is very subjective.

model_analyzer/config/input/config_command_profile.py
```diff
@@ -1529,19 +1529,34 @@ def _autofill_values(self):
...
-                    "request_rate": self.request_rate,
-                }
+                if self.run_config_search_mode != "optuna":
+                    new_model["parameters"] = {
+                        "batch_sizes": self.batch_sizes,
+                        "concurrency": self.concurrency,
+                        "request_rate": self.request_rate,
+                    }
+                else:
+                    if self._fields["batch_sizes"].is_set_by_user():
```
dyastremsky, line 1539: Why does this work differently than concurrency and request_rate? Can we remove the branching to speed up the code? E.g. set self.batch_sizes to an empty list by default then update them if the user sets batch sizes?

dyastremsky: Actually, it looks like Python may not optimize for branchless code, so removing the branching should only be done if it makes the code cleaner without making it do more work (I think it still might).

StackOverflow conversation on why Python does not optimize branchless code [here](stackoverflow.com/questions/72118249). Note that I'm seeing different things online about whether it optimizes for branchless programming or not.

tests/test_config.py
```diff
@@ -18,14 +18,26 @@
...
 
+from model_analyzer.analyzer import Analyzer
 from model_analyzer.cli.cli import CLI
+from model_analyzer.config.generate.config_parameters import (
+    ParameterCategory,
+    ParameterType,
+)
 from model_analyzer.config.input.config_command_profile import ConfigCommandProfile
 from model_analyzer.config.input.config_command_report import ConfigCommandReport
 from model_analyzer.config.input.config_defaults import (
```
dyastremsky, line 33: Would it maybe make sense to import than config_defaults, then use all of the below constants? That would make this section more concise. It seems like you'd be using all/most of them anyway.

tests/test_config.py
```diff
@@ -2388,6 +2400,172 @@ def test_model_type_llm(self):
...
+
+        config = self._evaluate_config(args, yaml_content)
+
+        analyzer = Analyzer(config, MagicMock(), MagicMock(), MagicMock())
+        analyzer._populate_search_parameters()
+
+        batch_sizes = analyzer._search_parameters["add_sub"].get_parameter(
+            "batch_sizes"
+        )
+        self.assertEqual(ParameterType.MODEL, batch_sizes.ptype)
```
dyastremsky, line 2431: Could you add comment headers to visually organize the test (helps with skimming for understanding and finding the parts you need)? E.g. this section is for batch sizes, next one is for concurrency sizes, etc.

tests/test_config.py
```diff
@@ -2388,6 +2400,172 @@ def test_model_type_llm(self):
             config.inference_output_fields, DEFAULT_LLM_INFERENCE_OUTPUT_FIELDS
         )
 
+    def test_config_parameter_creation_default(self):
+        """
+        Test that config parameters are correctly created in default optuna case
```
dyastremsky, line 2405: I don't know too much about optuna, but do we need tests that are for default optuna multi-model or non-optuna single-model?

tests/test_config.py
```diff
@@ -2388,6 +2400,172 @@ def test_model_type_llm(self):
             config.inference_output_fields, DEFAULT_LLM_INFERENCE_OUTPUT_FIELDS
         )
```
dyastremsky, line 2402: Also, dang, there are a lot of tests in this file! This is likely a reach here, but I wanted to drop the idea of [parameterization](https://docs.pytest.org/en/7.1.x/example/parametrize.html) in case it's possible here. It may not be, but wanted to share in case we can use it in more testing in the future.


This comment is a conversational comment, it's not a review/blocking comment.

Note: I dropped a Pytest link (and switching to it should be as simple as changing the command line command from `unittest` to `pytest`, but [unittest also seems to have options](https://stackoverflow.com/questions/32899/how-do-you-generate-dynamic-parameterized-unit-tests-in-python) if we prefer to not go the PyTest route.

dyastremsky (commented):
Fantastic work on this, Brian! I appreciate your explanations, comments, and cleanup of past code. Some great code quality here.

model_analyzer/config/generate/config_parameters.py
```diff
@@ -26,23 +28,23 @@ class ConfigParameters:
     Contains information about all configuration parameters the user wants to search
     """
 
-    def __init__(self):
-        self._parameters: Dict[str, ConfigParameter] = {}
+    # These map to the run-config-search fields (see config.md for details)
```
dyastremsky, line 31:
```suggestion
    # These map to the run-config-search fields.
    # See github.com/triton-inference-server/model_analyzer/blob/main/docs/config.md
```

model_analyzer/config/generate/config_parameters.py
```diff
@@ -26,23 +28,23 @@ class ConfigParameters:
...
+    exponential_rcs_parameters = ["batch_sizes", "concurrency"]
```
dyastremsky, line 32: I think I'm still missing the context here. Could we sync offline about this? I think something isn't clear here, and I wonder if we could resolve it.

model_analyzer/config/input/config_command_profile.py
dyastremsky: Understood, thanks for sharing the greater code logic. I'll resolve this.

If it's something you're interested, a ternary operator could look quite nice in a situation like this. That's a personal choice though, this looks good too. `new_model["parameters"] = self._fields["batch_sizes"].is_set_by_user() ? self.batch_sizes : []`

tests/test_config.py
dyastremsky: Oh, I mean `import model_analyzer.config.input.config_defaults as config_defaults` or equivalent. Not `from model_analyzer.config.input.config_defaults import *`.

model_analyzer/config/generate/config_parameters.py
```diff
@@ -26,23 +29,22 @@ class ConfigParameters:
     Contains information about all configuration parameters the user wants to search
     """
 
-    def __init__(self):
-        self._parameters: Dict[str, ConfigParameter] = {}
+    exponential_parameters = ["batch_sizes", "concurrency"]
+    linear_parameters = ["instance_group"]
```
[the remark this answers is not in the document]
dyastremsky: Thanks for clarifying!

tests/test_config.py
nv-braf: Oh, yes you could do that but then the names get longer when they're used. My preference is to call them out explicitly.

model_analyzer/config/generate/config_parameters.py
dyastremsky: Decision: Just add headers for the two "types". The comment with the explanation and documentation link will go into SearchParameter.

model_analyzer/config/generate/config_parameters.py
```diff
@@ -53,12 +54,146 @@ def get_type(self, name: str) -> ParameterType:
...
+    def _populate_parameters(
+        self,
+        config: ConfigCommandProfile,
+        parameters: Dict[str, Any],
+    ) -> None:
+        self._populate_parameter(
+            parameter_name="batch_sizes",
+            rcs_min_value=config.run_config_search_min_model_batch_size,
+            rcs_max_value=config.run_config_search_max_model_batch_size,
+            parameter_list=parameters["batch_sizes"],
```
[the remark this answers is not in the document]
dyastremsky: Since we're using multiple lists, leaving it this way sounds good.

model_analyzer/config/generate/config_parameters.py
```diff
@@ -53,12 +55,147 @@ def get_type(self, name: str) -> ParameterType:
...
+        self._populate_model_config_parameters(config, model_config_parameters)
+
+    def _populate_parameters(
+        self,
+        config: ConfigCommandProfile,
+        parameters: Dict[str, Any],
+    ) -> None:
+        self._populate_parameter(
+            parameter_name="batch_sizes",
+            rcs_parameter_min_value=config.run_config_search_min_model_batch_size,
```
dyastremsky, line 80: We talked about potentially breaking this into two populate parameters (RCS vs list parameter)

nv-braf: Broken!

dyastremsky (approved):
Thanks for addressing the changes so well! The code looks fantastic. Great work!

PR #873, opened 2024-05-09
Run CPU-only unit tests automatically for pull requests

dyastremsky:
This pull request adds an automation to automatically run unit tests for new pull requests. It checks them against Python 3.8 and 3.11. I updated the supported Python versions to be aligned with those supported in PyTriton, Triton CLI, and GenAI-Perf, especially since Ubuntu 22.04 only supports 3.9+.

The automation uses unittest-parallel to speed up the running of unit tests. I also added the mypy cache to gitignore to avoid accidentally committing those files.

You can see the results in the new jobs that are running for this PR.


nv-braf (approved):
Looks good!

PR #877, opened 2024-05-15
Optuna Search Class

nv-braf:
Initial work to create an Optuna Search class w/ some basic unit testing.
There is still a ton of work to do, and this has been highlighted and tagged with stories in the comments below.


tests/test_optuna_run_config_generator.py
```diff
@@ -0,0 +1,134 @@
...
+        trial = self._rcg._study.ask()
+        self._rcg._create_trial_objectives(trial)
+        run_config = self._rcg._create_objective_based_run_config()
+
+        model_config = run_config.model_run_configs()[0].model_config()
+        perf_config = run_config.model_run_configs()[0].perf_config()
+
+        self.assertEqual(model_config.to_dict()["name"], self._test_config_dict["name"])
+
+        # These values are the result of using a fixed seed of 100
```
tgerdesnv, line 99: Are we at risk of this failing if optuna version changes? Looks like our requirements.txt uses > version instead of pinning an exact version.

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+        self,
+        model_name: str,
+        model: ModelProfileSpec,
+        concurrency: int,
+    ) -> PerfAnalyzerConfig:
+        perf_analyzer_config = PerfAnalyzerConfig()
+
+        perf_analyzer_config.update_config_from_profile_config(model_name, self._config)
+
+        perf_config_params = {"batch-size": 1, "concurrency-range": concurrency}
```
tgerdesnv, line 292: hardcoded batch size?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+        perf_analyzer_config.update_config(perf_config_params)
+
+        perf_analyzer_config.update_config(model.perf_analyzer_flags())
+        return perf_analyzer_config
+
+    def _print_debug_logs(
```
tgerdesnv, line 298: Any reason to create a function that isn't implemented or called? You can create it if/when it is time to implement.

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+from model_analyzer.triton.model.model_config_variant import ModelConfigVariant
+
+from .config_generator_interface import ConfigGeneratorInterface
+
+logger = logging.getLogger(LOGGER_NAME)
+
+
+class OptunaRunConfigGenerator(ConfigGeneratorInterface):
+    """
+    Hill climbing algorithm to create RunConfigs
```
tgerdesnv, line 53: not hill climbing?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+
+    def __init__(
+        self,
+        config: ConfigCommandProfile,
+        gpus: List[GPUDevice],
```
tgerdesnv, line 59: can we pass in the GPU count instead of all GPUs?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+        models: List[ModelProfileSpec],
+        client: TritonClient,
```
tgerdesnv, line 61: any reason to pass in the client?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+        model_variant_name_manager: ModelVariantNameManager,
+        search_parameters: SearchParameters,
+        metrics_manager: MetricsManager,
```
tgerdesnv, line 64: any reason to pass this in?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+            The next RunConfig generated by this class
+        """
+        default_run_config = self._create_default_run_config()
+        yield default_run_config
+        self._default_measurement = self._last_measurement
+
+        # TODO: TMA-1884: Need a default + config option for trial number
+        n_trials = 20
+        # TODO: TMA-1885: Need an early exit strategy
+        for _ in range(n_trials):
```
tgerdesnv, line 135: Should the generator be the one that knows when to end? Or should the person calling the generator decide when to stop?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+            self._create_trial_objectives(trial)
+            run_config = self._create_objective_based_run_config()
+            yield run_config
+            score = self._calculate_score()
+            self._study.tell(trial, score)
+
+    def _create_trial_objectives(self, trial) -> None:
+        # TODO: TMA-1925: Use SearchParameters here
+        self._instance_count = trial.suggest_int("instance_count", 1, 8)
+        self._batch_size = int(2 ** trial.suggest_int("batch_size", 1, 10))
```
tgerdesnv, line 146: I don't love setting member variables for each iteration and then consuming them. It hides a lot of the data flow. It feels like create_trial_objectives should return some trial_objectives data struct, which is then consumed downstream?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+        for _ in range(n_trials):
+            trial = self._study.ask()
+            self._create_trial_objectives(trial)
+            run_config = self._create_objective_based_run_config()
+            yield run_config
+            score = self._calculate_score()
+            self._study.tell(trial, score)
+
+    def _create_trial_objectives(self, trial) -> None:
+        # TODO: TMA-1925: Use SearchParameters here
```
tgerdesnv, line 144: I see lots of hardcoding limits (for example, instance count 1 to 8). Is SearchParameters going to take care of fixing that?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+        ]
+
+        param_combo["max_batch_size"] = self._batch_size
+
+        return param_combo
+
+    def _calculate_score(self) -> float:
+        if self._last_measurement:
+            score = self._default_measurement.compare_measurements(  # type: ignore
+                self._last_measurement
```
tgerdesnv, line 197: Should the score be based on comparing the last measurement? or the default measurement?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -0,0 +1,302 @@
...
+        default_run_config = RunConfig(self._triton_env)
+        # TODO: TMA-1927: Add support for multi-model
+        default_model_run_config = self._create_default_model_run_config(
+            self._models[0]
+        )
+        default_run_config.add_model_run_config(default_model_run_config)
+
+        return default_run_config
+
+    def _create_default_model_run_config(
```
tgerdesnv, line 215: Is this any different code from other RCGs? If not, we should not be cutting and pasting.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Yes, but not a good one. It's because I copy/pasted the QRCG to start my code and it had this as an arg. I've removed this from both Optuna and Quick RCGs

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: We can! I've changed this in both the Optuna and Quick RCGs

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: The plan is to have a class/method that determines when we should end. I already have a story for this TMA-1885.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Yes, this is all temporary and SearchParameters will replace this code.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: This is temporary and will be replaced when SearchParameters gets used. I agree a struct is a good choice for passing these values around

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: I believe comparing against the default measurement will provide us with a better search path, but I have no proof of this.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Agreed. I've filed TMA-1930 for this work.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: This is what's used in the quick RCG as well. I've fixed it in both places.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: If this code was going into main then this would be bad form. This is just a placeholder so I can keep the TODO in the right location (in case someone else picks up this work).

tests/test_optuna_run_config_generator.py
nv-braf: We might be. I don't know of a better way to lock this down.

tgerdesnv: You can require exact version instead of >= in requirements.txt

model_analyzer/config/generate/optuna_run_config_generator.py
tgerdesnv: ok, I misread. I thought this was comparing against the previous measurement.

model_analyzer/config/generate/optuna_run_config_generator.py
tgerdesnv: is the TODO going to be resolved before this is merged into main?

model_analyzer/config/generate/optuna_run_config_generator.py
tgerdesnv: ok but it is still hardcoded? Isn't it possible for the user to specify a batch size?

model_analyzer/config/generate/optuna_run_config_generator.py
tgerdesnv: Looks like you removed this as well. Resolving

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Yes, this will be done before it goes into main

tests/test_optuna_run_config_generator.py
nv-braf: Done

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: No, not presently. But, we could add this. I've filed TMA-1934 to track this work.

tgerdesnv (approved):

PR #881, opened 2024-05-22
Using SearchParameters in OptunaRCG

nv-braf:
Connected up SearchParameters class in the OptunaRCG. Updated unit testing to reflect this and I've also run live tests which include searching `max_queue_delay` which is not possible via quick search.

### DIFF
--- a/model_analyzer/config/generate/optuna_run_config_generator.py
+++ b/model_analyzer/config/generate/optuna_run_config_generator.py
@@ -231,6 +231,7 @@ class OptunaRunConfigGenerator(ConfigGeneratorInterface):
         self, trial_objectives: TrialObjectives
     ) -> ParameterCombo:
         param_combo: ParameterCombo = {}
+        # TODO: TMA-1938: Need to look at model in ParameterSearch and add this as a parameter
         param_combo["dynamic_batching"] = []
 
         # TODO: TMA-1927: Add support for multi-model
--- a/model_analyzer/config/generate/optuna_run_config_generator.py
+++ b/model_analyzer/config/generate/optuna_run_config_generator.py
@@ -51,7 +51,8 @@ from .config_generator_interface import ConfigGeneratorInterface
 
 logger = logging.getLogger(LOGGER_NAME)
 
-TrialObjectives: TypeAlias = Dict[str, Any]
+TrialObjective: TypeAlias = Union[str | int]
+TrialObjectives: TypeAlias = Dict[str, TrialObjective]
 ParameterCombo: TypeAlias = Dict[str, Any]
 
 
@@ -157,22 +158,20 @@ class OptunaRunConfigGenerator(ConfigGeneratorInterface):
             parameter = self._search_parameters.get_parameter(parameter_name)
 
             if parameter:
-                trial_objectives = self._add_trial_objective(
-                    trial, parameter_name, parameter, trial_objectives
+                trial_objectives[parameter_name] = self._create_trial_objective(
+                    trial, parameter_name, parameter
                 )
 
         # TODO: TMA-1884: Need an option to choose btw. concurrency formula and optuna searching
-        trial_objectives = self._set_objective_concurrency(trial_objectives)
+        trial_objectives["concurrency"] = self._get_objective_concurrency(
+            trial_objectives
+        )
 
         return trial_objectives
 
-    def _add_trial_objective(
-        self,
-        trial: optuna.Trial,
-        name: str,
-        parameter: SearchParameter,
-        trial_objectives: TrialObjectives,
-    ) -> TrialObjectives:
+    def _create_trial_objective(
+        self, trial: optuna.Trial, name: str, parameter: SearchParameter
+    ) -> TrialObjective:
         if parameter.category is ParameterCategory.INTEGER:
             objective = trial.suggest_int(
                 name, parameter.min_range, parameter.max_range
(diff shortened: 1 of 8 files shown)


model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -44,20 +51,31 @@
...
+TrialObjectives: TypeAlias = Dict[str, Any]
+ParameterCombo: TypeAlias = Dict[str, Any]
+
 
 class OptunaRunConfigGenerator(ConfigGeneratorInterface):
     """
     Use Optuna algorithm to create RunConfigs
     """
 
+    # This list represents all possible parameters Optuna can currently search for
```
tgerdesnv, line 63: What is the plan to take this from a fixed list to full flexibility?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -126,24 +145,70 @@ def get_configs(self) -> Generator[RunConfig, None, None]:
         # TODO: TMA-1885: Need an early exit strategy
         for _ in range(n_trials):
             trial = self._study.ask()
-            self._create_trial_objectives(trial)
-            run_config = self._create_objective_based_run_config()
+            trial_objectives = self._create_trial_objectives(trial)
```
tgerdesnv, line 148: 👍

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -126,24 +145,70 @@ def get_configs(self) -> Generator[RunConfig, None, None]:
...
-        # TODO: TMA-1925: Use SearchParameters here
-        self._instance_count = trial.suggest_int("instance_count", 1, 8)
-        self._batch_size = int(2 ** trial.suggest_int("batch_size", 1, 10))
+    def _create_trial_objectives(self, trial: optuna.Trial) -> TrialObjectives:
+        trial_objectives: TrialObjectives = {}
+        for parameter_name in OptunaRunConfigGenerator.optuna_parameter_list:
+            parameter = self._search_parameters.get_parameter(parameter_name)
+
+            if parameter:
+                trial_objectives = self._add_trial_objective(
```
tgerdesnv, line 160: This takes and returns (the same) dict? Is there a cleaner way to do that? I can see us getting burned by this. Maybe just return a single trial objective? And accumulate them in a list/dict?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -126,24 +145,70 @@ def get_configs(self) -> Generator[RunConfig, None, None]:
...
+        trial_objectives[name] = objective
 
-    def _create_objective_based_run_config(self) -> RunConfig:
-        param_combo = self._create_parameter_combo()
+        return trial_objectives
+
+    def _set_objective_concurrency(
+        self, trial_objectives: TrialObjectives
+    ) -> TrialObjectives:
+        concurrency = (
```
tgerdesnv, line 196: Is this going to work consistently? Or are we going to need to lean on Optuna to pick the load? I believe we've observed LLMs and other models not follow this rule.

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -159,27 +224,38 @@ def _create_objective_based_run_config(self) -> RunConfig:
...
 
-    def _create_parameter_combo(self) -> Dict[str, Any]:
-        # TODO: TMA-1925: Use SearchParameters here
-        param_combo: Dict["str", Any] = {}
-        param_combo["dynamic_batching"] = {}
+    def _create_parameter_combo(
+        self, trial_objectives: TrialObjectives
+    ) -> ParameterCombo:
+        param_combo: ParameterCombo = {}
+        param_combo["dynamic_batching"] = []
```
tgerdesnv, line 238: be careful with hardcoding these. If we are moving forward we need full flexibility. Some models don't support dynamic batching.

model_analyzer/config/generate/search_parameters.py
```diff
@@ -49,11 +49,13 @@ def __init__(
 
         self._populate_search_parameters()
 
-    def get_parameters(self) -> List[SearchParameter]:
-        return [v for v in self._search_parameters.values()]
+    def get_parameter(self, name: str) -> Optional[SearchParameter]:
+        try:
```
tgerdesnv, line 53: Doesn't dict.get() perform this for you?

model_analyzer/config/generate/search_parameters.py
```diff
@@ -73,6 +75,9 @@ def get_list(self, name: str) -> Optional[List[Any]]:
     def _populate_search_parameters(self) -> None:
         if self._parameters:
             self._populate_parameters()
+        else:
+            # Always populate batch size if nothing is specified
```
tgerdesnv, line 79: What is the background for this? Is it just batch size that needs to be specified? Should this function name be more generic?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -44,20 +51,31 @@
...
+
     def __init__(
         self,
         config: ConfigCommandProfile,
         gpu_count: int,
         models: List[ModelProfileSpec],
         model_variant_name_manager: ModelVariantNameManager,
-        search_parameters: SearchParameters,
-        seed: Optional[int] = 0,
+        search_parameters: Dict[str, SearchParameters],
```
tgerdesnv, line 77: Why dict? Is it for multi-model support?

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: I believe this will always be a fixed (ever growing) list because we need to understand/hand-code how we take a parameter and write it into the perf/model/build config we will generate.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: We are going to have a user option to either use the concurrency formula or have optuna pick the load (there's already a story filed for this).

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Correct. Also needed for ensemble/BLS/LLM

model_analyzer/config/generate/search_parameters.py
nv-braf: If the user doesn't specify any parameters then we just need to populate batch sizes (for now). This will change when we add the option for the user to have optuna search concurrency as well.

model_analyzer/config/generate/search_parameters.py
nv-braf: It does (apparently I like re-inventing the wheel!). Fixed.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Done

model_analyzer/config/generate/search_parameters.py
tgerdesnv: Can you explain how it will change?
My main point is that it seems like the function should be named something more generic (and for now it only adds batch_size, but in the future it may do more or less or something else)

nv-braf: Updated to make this a call to a more generic method

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: I've added a todo + story for this.

tgerdesnv (approved):

PR #882, opened 2024-05-24
Add Percentage Search Space to Optuna

nv-braf:
Adds the ability for the user to specify min/max % of search space that Optuna will try before terminating.

Note: This is just the first step and this is going to be enhanced in upcoming stories to include an option for min/max num of tries, as well as an algorithm for early termination.


debermudez (approved):
2 small questions discussed offline.
Looks good to me.
Nice work

PR #884, opened 2024-05-28
Add CLI option for min/max number of trials in Optuna Search

nv-braf:
Give the user the ability to set the maximum number of trials (minimum will be used when early exit support is added) during an Optuna search.


model_analyzer/config/input/config_command_profile.py
```diff
@@ -936,6 +938,24 @@ def _add_run_search_configs(self):
                 description="Maximum percentage of the search space to profile when using Optuna",
             )
         )
+        self._add_config(
+            ConfigField(
+                "optuna_min_trials",
```
tgerdesnv, line 943: I see support for min trials, but I don't see it used anywhere?

nv-braf: This is only relevant/used once the code for early exit is added. I've just added the option for it now.

tgerdesnv (approved):

PR #885, opened 2024-05-28
Adding support for concurrency formula as an option in Optuna search

nv-braf:
Gives the user the option to use our concurrency formula rather than having Optuna search the full concurrency range.

By default we allow Optuna to search the concurrency space. This gives the user the option of using our concurrency formula (2 * instance_count * batch_size) instead, which will reduce the search space and should improve the search time.


tests/test_cli.py
```diff
@@ -385,6 +386,7 @@ def _test_boolean_option(self, option_struct):
         cli = option_struct.cli_subcommand()
         _, config = cli.parse()
         option_value = config.get_config().get(option_with_underscores).value()
+        # Boolean values must always default to False
```
debermudez, line 389: why is this necessary?
The arg could use default=True

model_analyzer/config/generate/search_parameters.py
```diff
@@ -90,23 +90,14 @@ def _number_of_configurations_for_parameter(
         return number_of_parameter_configs
 
     def _populate_search_parameters(self) -> None:
-        if self._parameters:
-            self._populate_parameters()
-        else:
-            self._populate_default_parameters()
```
debermudez, line 96: why did this method get removed?

tests/test_optuna_run_config_generator.py
```diff
@@ -159,6 +159,43 @@ def test_create_objective_based_run_config(self):
...
+        trial = rcg._study.ask()
+        trial_objectives = rcg._create_trial_objectives(trial)
+        run_config = rcg._create_objective_based_run_config(trial_objectives)
+
+        model_config = run_config.model_run_configs()[0].model_config()
+        perf_config = run_config.model_run_configs()[0].perf_config()
+
+        self.assertEqual(model_config.to_dict()["name"], self._test_config_dict["name"])
+
+        # These values are the result of using a fixed seed of 100
```
debermudez, line 189: Nice note!

tests/test_search_parameters.py
```diff
@@ -248,6 +248,38 @@ def test_search_parameter_creation_default(self):
...
+
+        yaml_content = """
+        profile_models: add_sub
+        """
+        config = TestConfig()._evaluate_config(args=args, yaml_content=yaml_content)
+
+        analyzer = Analyzer(config, MagicMock(), MagicMock(), MagicMock())
+        analyzer._populate_search_parameters()
+
+        concurrency = analyzer._search_parameters["add_sub"].get_parameter(
```
debermudez, line 277: This wont error out?
get_parameter is designed to return None when a parameter isnt found?

tests/test_cli.py
nv-braf: This is a byproduct of how the `ConfigPrimitive(bool)` works. The field must be False by default and then it is set to True by setting it on the CLI (or in the config). You cannot have a default of True (and set it to False via the CLI/config)

model_analyzer/config/generate/search_parameters.py
nv-braf: This was just a placeholder until this code got added. The new code has the correct functionality.

tests/test_search_parameters.py
nv-braf: Yes, it will return None when the key is not found.

tests/test_cli.py
debermudez: i see. okay

debermudez (approved):

PR #890, opened 2024-05-31
Optuna Early Exit

nv-braf:
Add logic for early exit in Optuna along with a config option to set the threshold value


model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -155,12 +156,13 @@ def get_configs(self) -> Generator[RunConfig, None, None]:
...
 
+        min_configs_to_search = self._determine_minimum_number_of_configs_to_search()
         max_configs_to_search = self._determine_maximum_number_of_configs_to_search()
         # TODO: TMA-1885: Need an early exit strategy
-        for trial_count in range(max_configs_to_search):
+        for trial_number in range(max_configs_to_search):
             trial = self._study.ask()
             trial_objectives = self._create_trial_objectives(trial)
-            logger.debug(f"Trial {trial_count+1} of {max_configs_to_search}:")
+            logger.debug(f"Trial {trial_number+1} of {max_configs_to_search}:")
```
tgerdesnv, line 165: I'm noticing you are printing trial+1, but best_trial_number doesn't have the +1. Not sure it will burn us anywhere, but wanted to draw your attention to it. Wondering if it would be easier to change the range from 1:N instead of 0:N-1

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -208,7 +216,7 @@ def _determine_trials_based_on_max_percentage_of_search_space(self) -> int:
 
         return max_trials_based_on_percentage_of_search_space
 
-    def _decide_between_percentage_and_trial_count(
+    def _decide_max_between_percentage_and_trial_count(
```
tgerdesnv, line 219: I think function names should be based on what they accomplish (determine_max_configs_to_search) instead of HOW they do it (deciding between two values)

model_analyzer/config/input/config_command_profile.py
```diff
@@ -957,6 +958,15 @@ def _add_run_search_configs(self):
                 description="Maximum number of trials to profile when using Optuna",
             )
         )
+        self._add_config(
+            ConfigField(
+                "optuna_early_exit_threshold",
+                flags=["--optuna_early_exit_threshold"],
+                field_type=ConfigPrimitive(int),
+                default_value=DEFAULT_OPTUNA_EARLY_EXIT_THRESHOLD,
+                description="Number of trials to attempt before triggering early exit when using Optuna",
```
tgerdesnv, line 967: Is this true? Code implies it is a number of attempts in a row with no improvement?

nv-braf: The code is correct. It's the number of trials without improvement (min. number of trials must still be met). I'll update the description.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Good suggestion. I've made the change.

tgerdesnv (approved):

PR #891, opened 2024-06-03
Check that model supports dynamic batching when creating param_combo

nv-braf:
Check that model supports dynamic batching before adding in when creating param_combos.


tgerdesnv (approved):
Approved. Longer term, I think there is room to refactor or create a param_combo class to make sure that this knowledge isn't needed everywhere in the code base.

PR #892, opened 2024-06-03
Adding support for client batch size

nv-braf:
Adds support for client batch size. 

FYI, I had reversed the roles of `max_batch_size` and `batch_sizes` when I originally created the SearchParameters/Optuna classes. This is now fixed and is reflected in the classes and unit tests.

### DIFF
--- a/model_analyzer/config/generate/optuna_run_config_generator.py
+++ b/model_analyzer/config/generate/optuna_run_config_generator.py
@@ -65,6 +65,7 @@ class OptunaRunConfigGenerator(ConfigGeneratorInterface):
     # This list represents all possible parameters Optuna can currently search for
     optuna_parameter_list = [
         "batch_sizes",
+        "max_batch_size",
         "instance_group",
         "concurrency",
         "max_queue_delay_microseconds",
@@ -291,7 +292,7 @@ class OptunaRunConfigGenerator(ConfigGeneratorInterface):
         concurrency_formula = (
             2
             * int(trial_objectives["instance_group"])
-            * int(trial_objectives["batch_sizes"])
+            * int(trial_objectives["max_batch_size"])
         )
         concurrency = (
             self._config.run_config_search_max_concurrency
@@ -344,8 +345,8 @@ class OptunaRunConfigGenerator(ConfigGeneratorInterface):
                 }
             ]
 
-        if "batch_sizes" in trial_objectives:
-            param_combo["max_batch_size"] = trial_objectives["batch_sizes"]
+        if "max_batch_size" in trial_objectives:
+            param_combo["max_batch_size"] = trial_objectives["max_batch_size"]
 
         if "max_queue_delay_microseconds" in trial_objectives:
             param_combo["dynamic_batching"] = {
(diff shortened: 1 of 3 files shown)


model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -435,8 +436,16 @@ def _create_model_run_config(
         model_config_variant: ModelConfigVariant,
         trial_objectives: TrialObjectives,
     ) -> ModelRunConfig:
+        trial_batch_size = (
+            int(trial_objectives["batch_sizes"])
```
tgerdesnv, line 440: batch_sizes implies a list or multiple values. Is this always a scalar?
Why does it need to be cast to an int? Is the value coming out of trial_objectives a float?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -435,8 +436,16 @@ def _create_model_run_config(
...
+            if "batch_sizes" in trial_objectives
+            else DEFAULT_BATCH_SIZES
+        )
         perf_analyzer_config = self._create_perf_analyzer_config(
-            model.model_name(), model, int(trial_objectives["concurrency"])
+            model.model_name(),
```
tgerdesnv, line 445: I'm not a fan of 4 args to a function where you have to know the exact order to pass them in. At a minimum it should be named args.

model_analyzer/config/generate/search_parameters.py
```diff
@@ -153,6 +152,31 @@ def _populate_concurrency(self) -> None:
...
+            )
+        elif "max_batch_size" in self._model_config_parameters.keys():
+            parameter_list = self._model_config_parameters["max_batch_size"]
+
+            self._populate_list_parameter(
+                parameter_name="max_batch_size",
+                parameter_list=parameter_list,
+                parameter_category=ParameterCategory.INT_LIST,
+            )
+        else:
```
tgerdesnv, line 173: Is there any difference between the first IF and this final ELSE?

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: The cast is to pass static type checking (it could be a `str | int`, but in this case we know it's an int).
`batch_sizes` is what the parameter is called in the config (even though it's a scalar value)

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: I've made them named args.

model_analyzer/config/generate/search_parameters.py
nv-braf: No, I've refactored this to make it cleaner.

model_analyzer/config/generate/search_parameters.py
```diff
@@ -204,12 +190,22 @@ def _populate_instance_group(self) -> None:
...
             self._populate_rcs_parameter(
                 parameter_name="instance_group",
                 rcs_parameter_min_value=self._config.run_config_search_min_instance_count,
                 rcs_parameter_max_value=self._config.run_config_search_max_instance_count,
             )
 
+    def _is_key_in_model_config_parameters(self, key: str) -> bool:
+        key_found = bool(
+            self._model_config_parameters
+            and key in self._model_config_parameters.keys()
```
tgerdesnv, line 204: minor nit:

    "key in mydict": This is the standard and recommended way to check for key existence in a dictionary. It is concise and directly checks the key in the dictionary.
    "key in mydict.keys()": This is redundant and not recommended. It achieves the same result but with additional overhead.

nv-braf: Fixed.

tgerdesnv (approved):

PR #898, opened 2024-06-10
Adding BLS composing models to SearchParameters

nv-braf:
Composing models in a BLS can also have search parameters:
```
run_config_search_mode: optuna
use_concurrency_formula: True
profile_models: bls
bls_composing_models:
  add:
    model_config_parameters:
      max_batch_size: [1, 4, 8]
      dynamic_batching:
        max_queue_delay_microseconds: [100, 200, 300]
  sub:
    model_config_parameters:
      max_batch_size: [32, 64, 128]
      dynamic_batching:
        max_queue_delay_microseconds: [400, 500, 600]
```
This story adds support for this to the SearchParameters class (which is used by OptunaRCG).

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -423,5 +423,14 @@ class Analyzer:
     def _populate_search_parameters(self):
         for model in self._config.profile_models:
             self._search_parameters[model.model_name()] = SearchParameters(
-                self._config, model.parameters(), model.model_config_parameters()
+                self._config,
+                model.parameters(),
+                model.model_config_parameters(),
+                is_bls_model=bool(self._config.bls_composing_models),
+            )
+        for model in self._config.bls_composing_models:
+            self._search_parameters[model.model_name()] = SearchParameters(
+                self._config,
+                model.parameters(),
+                model.model_config_parameters(),
             )
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -84,6 +84,7 @@ class Analyzer:
         )
 
         self._search_parameters: Dict[str, SearchParameters] = {}
+        self._composing_search_parameters: Dict[str, SearchParameters] = {}
 
     def profile(
         self, client: TritonClient, gpus: List[GPUDevice], mode: str, verbose: bool
@@ -119,6 +120,7 @@ class Analyzer:
         self._create_metrics_manager(client, gpus)
         self._create_model_manager(client, gpus)
         self._populate_search_parameters()
+        self._populate_composing_search_parameters()
 
         if self._config.triton_launch_mode == "remote":
             self._warn_if_other_models_loaded_on_remote_server(client)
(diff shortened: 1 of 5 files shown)


model_analyzer/analyzer.py
```diff
@@ -423,5 +423,14 @@ def _warn_if_other_models_loaded_on_remote_server(self, client):
...
         for model in self._config.profile_models:
             self._search_parameters[model.model_name()] = SearchParameters(
-                self._config, model.parameters(), model.model_config_parameters()
+                self._config,
+                model.parameters(),
+                model.model_config_parameters(),
+                is_bls_model=bool(self._config.bls_composing_models),
+            )
+        for model in self._config.bls_composing_models:
+            self._search_parameters[model.model_name()] = SearchParameters(
```
debermudez, line 432: Is there ever an instance where the composing model can have the same name as a higher level model?
Seems like you would overwrite an entry in that case?

model_analyzer/config/generate/search_parameters.py
```diff
@@ -120,11 +122,15 @@ def _populate_search_parameters(self) -> None:
 
     def _populate_parameters(self) -> None:
         self._populate_batch_sizes()
-        self._populate_concurrency()
-        # TODO: Populate request rate - TMA-1903
+
+        if not self._is_bls_model:
```
debermudez, line 126: where is concurrency populated when is it a bls model?

tests/test_search_parameters.py
```diff
@@ -457,6 +457,104 @@ def test_total_possible_configurations(self):
...
+        self.assertEqual(
+            default.DEFAULT_RUN_CONFIG_MIN_INSTANCE_COUNT, instance_group.min_range
+        )
+        self.assertEqual(
+            default.DEFAULT_RUN_CONFIG_MAX_INSTANCE_COUNT, instance_group.max_range
+        )
+
+        # ADD/SUB (composing models)
+        # =====================================================================
+        # The top level model of a BLS can only search instance count
```
debermudez, line 517: duplicate comment?

nv-braf: I've updated the comment to reflect what is supposed to happen for composing models

model_analyzer/config/generate/search_parameters.py
nv-braf: It should be populated for the top-level model only (as concurrency is just specified once as an arg to PA). I've updated the code to reflect this.

model_analyzer/analyzer.py
nv-braf: Good catch! This is why I have separate models and composing_models in the code. I've updated this logic to do the same.

debermudez (approved):

model_analyzer/config/generate/search_parameters.py
debermudez: thanks for explaining that.

PR #899, opened 2024-06-11
Adds BLS scaffolding to everything up to OptunaRCG

nv-braf:
I've split the BLS OptunaRCG work into two stories to make it easier to review. 

This first story contains all the code changes to support BLS outside of the OptunaRCG class:

- Adds populate composing search parameters
- Adds missing check that batch size is supported when generating search parameters (discovered during BLS testing)
- PIpes composing_models all the way to OptunaRCG
- Updates unit-testing to have an actual ModelConfigProfileSpec (which is necessary to check if batch size exists in the protobuf)
- Adds new method to print the full variant name: `model_A::composing_model_A,composing_model_B...`

### DIFF
--- a/model_analyzer/analyzer.py
+++ b/model_analyzer/analyzer.py
@@ -23,6 +23,7 @@ from model_analyzer.cli.cli import CLI
 from model_analyzer.config.generate.base_model_config_generator import (
     BaseModelConfigGenerator,
 )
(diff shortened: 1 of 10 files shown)


tgerdesnv (approved):
LGTM

lkomali (approved):
LGTM

PR #903, opened 2024-06-24
Adding ensemble support to optuna

nv-braf:


tgerdesnv:
"15:48:16 [Model Analyzer] Existing measurement found for run config. Skipping profile"
Does this factor in the concurrency? Or, if a measurement with concurrency = X is found, will it not try concurrency = Y?

model_analyzer/analyzer.py
```diff
@@ -431,6 +431,7 @@ def _populate_search_parameters(self, client, gpus):
                 config=self._config,
                 model=model_profile_spec,
                 is_bls_model=bool(self._config.bls_composing_models),
+                is_ensemble_model=model_profile_spec.is_ensemble(),
```
tgerdesnv, line 434: I'm surprised we don't have an enum for type/mode instead of multiple bools passed in.. unless we support BLS + ensemble at the same time?

nv-braf:
> "15:48:16 [Model Analyzer] Existing measurement found for run config. Skipping profile" Does this factor in the concurrency? Or, if a measurement with concurrency = X is found, will it not try concurrency = Y?

This factors in concurrency (RunConfig = PA config + Model Config).

model_analyzer/analyzer.py
nv-braf: Yes, I could have probably done this with an enum. A model can either be BLS, ensemble, or vanilla.

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -183,7 +184,7 @@ def get_configs(self) -> Generator[RunConfig, None, None]:
             yield run_config
 
             score = self._calculate_score()
-            self._set_best_measurement(run_config, score)
+            self._set_best_measurement(run_config, score, trial_number)
```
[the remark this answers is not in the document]
nv-braf: Additional unit testing could have caught this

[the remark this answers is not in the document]
tgerdesnv: Is there unit testing now? Is there any new unit testing that should be added for similar cases?

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -451,6 +455,44 @@ def _create_objective_based_run_config(
...
+        if model.is_ensemble():
+            param_combo = self._create_ensemble_parameter_combo(
+                composing_trial_objectives
+            )
+        else:
+            param_combo = self._create_non_ensemble_parameter_combo(trial_objectives)
+
+        return param_combo
+
+    def _create_ensemble_parameter_combo(
```
tgerdesnv, line 472: I think a code smell is having a bunch of variables or functions for each type. create_X_parameter_combo, where X is ensemble vs bls vs multi-model vs other, and then having a bunch of embedded if/else all over the code. This is really where a base class + implementation classes for each type + a factory would come in handy. Consider that for future refactors.

model_analyzer/config/generate/optuna_run_config_generator.py
```diff
@@ -451,6 +455,44 @@ def _create_objective_based_run_config(
...
+        self,
+        composing_trial_objectives: ComposingTrialObjectives,
+    ) -> ParameterCombo:
+        """
+        For the ensemble model the only parameter we need to set
```
tgerdesnv, line 477: I just want to double check how the parent vs composing models are handled. This function is only called for the parent model, right? And the composing models are always handled elsewhere?

nv-braf: That's correct.

model_analyzer/config/generate/optuna_run_config_generator.py
nv-braf: Will do.

model_analyzer/config/generate/optuna_run_config_generator.py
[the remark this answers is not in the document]
nv-braf: No, there is not. This needs some additional testing. I'll file a story for this.

tgerdesnv (approved):

PR #909, opened 2024-07-08
Update README and versions for 2.48.0 / 24.07

fpetrini15:


README.md
mc-nv, line 1: We updating this page with banner:
https://github.com/triton-inference-server/model_analyzer/commit/2a1ae92835357cc518cd632fcc26958656e4aab7

fpetrini15: I saw those changes, but when I looked at the final README on the release branch it looked like it was restored. I figured it was a mistake, but it seems like we restore the README only after it has been tested to be stable. I understand now and will add the banner.

mc-nv: You confusion may come out of that we are restoring it on release date.

mc-nv (approved):

PR #911, opened 2024-07-10
Add support for multi-model profiling to the Optuna RCG

nv-braf:
Adds support for generating multi-model search profiles using the Optuna RCG including new unit testing to cover this code.

Here is the output from a live multi-model run I ran on my local machine using Optuna:

image
Screenshot 2024-07-10 063138

FYI, on my machine this took slightly longer than quick search, but produced a slightly better result (209% gain vs. 164% gain).


dyastremsky (approved):
Very clean. Fantastic work, Brian!

PR #912, opened 2024-07-13
Optuna CI Testing

nv-braf:
Adds four tests to protect Optuna:

- L0_optuna_search (single model)
- L0_optuna_bls
- L0_optuna_ensemble
- L0_optuna_multi_model

To make the review easier:

- These are all based on their quick search counterparts, with the only differences being:
   - Optuna is called (instead of quick) as the --run-config-search-mode in `test.sh`
   - Expected number of measurements changed, based on min/max % of search space, in `check_results.py`

The CI with passing tests can be found at: https://gitlab-master.nvidia.com/dl/dgx/tritonmodelanalyzer/-/pipelines/16546091


model_analyzer/config/input/config_command_profile.py
```diff
@@ -927,7 +927,7 @@ def _add_run_search_configs(self):
         self._add_config(
             ConfigField(
                 "min_percentage_of_search_space",
-                flags=["--min_percentage_of_search_space"],
+                flags=["--min-percentage-of-search-space"],
```
nv-braf, line 930: This was uncovered during the L0 test creation. It slipped through unit testing because I copied/pasted into `test_cli.py` and it wasn't caught in the PR

qa/L0_optuna_bls_model/check_results.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python3
+
+# Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
dyastremsky, line 3:
```suggestion
# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

qa/L0_optuna_bls_model/check_results.py
```diff
@@ -0,0 +1,116 @@
...
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
dyastremsky, line 5: This isn't the latest license. You might want to copy the current license from somewhere (e.g. [here](https://github.com/triton-inference-server/client/blob/db888f1aca588a10f5e4a4b02a4e4ff60d437b6f/src/c%2B%2B/perf_analyzer/genai-perf/genai_perf/parser.py#L1-L25)).

qa/L0_optuna_bls_model/check_results.py
```diff
@@ -0,0 +1,116 @@
...
+
+import argparse
+import sys
+
+import yaml
+
+
+class TestOutputValidator:
+    """
+    Functions that validate the output
```
dyastremsky, line 25: Redundant? If you want a description here, I think it'd be better to say that this is a class that validates X, Y, and Z for A, B (high-level things being validated and for what e.g. which modules).

qa/L0_optuna_bls_model/check_results.py
```diff
@@ -0,0 +1,116 @@
...
+        #  Model - bls:
+        #    concurrency: 1 to 1024 (11)
+        #    instance_group: 1 to 5 (5)
+        #  Composing model - add:
+        #    instance_group: 1 to 5 (5)
+        #
+        #  Minimum number of trials: 14 (5% of search space)
+        #  Maximum number of trials: 28 (10% of search space)
+        #
+        # Then you have 4 x (0-9) for the concurrency sweep on Top 3 + default
```
dyastremsky, line 62: I take it that these can only apply towards the max and not the min? Isn't 4 x 10 + 1 = 41, not 36? The math isn't clear here... tried a few combinations (e.g. top 3 + 1 = 4, not 10)... I can guess at the math, but a bit more clarity here would be helpful. This description is very informative overall though.

qa/L0_optuna_bls_model/test.sh
```diff
@@ -0,0 +1,98 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source ../common/util.sh
+source ../common/check_analyzer_results.sh
+create_logs_dir "L0_bls_model"
+
+# Set test parameters
+MODEL_ANALYZER="$(which model-analyzer)"
```
dyastremsky, line 21: Could you alphabetize this or organize this somehow, where possible? There are a lot of parameters, so that would make it easier to visually find what you want.

qa/L0_optuna_ensemble_model/check_results.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python3
+
+# Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
dyastremsky, line 3: Same issue with the copyright. Could you go through and make sure all copyrights state only 2024 for new files? And double check that the latest copyright text is reflected for new files.

qa/L0_optuna_ensemble_model/check_results.py
```diff
@@ -0,0 +1,116 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import sys
+
+import yaml
+
+
+class TestOutputValidator:
```
dyastremsky, line 23: Would you be able to move this class into a shared util folder, since it seems to be almost/mostly duplicated?

qa/L0_optuna_ensemble_model/test.sh
```diff
@@ -0,0 +1,97 @@
...
+MODEL_NAMES="$(echo $QA_MODELS | sed 's/ /,/g')"
+TRITON_LAUNCH_MODE=${TRITON_LAUNCH_MODE:="local"}
+CLIENT_PROTOCOL="grpc"
+PORTS=($(find_available_ports 3))
+GPUS=($(get_all_gpus_uuids))
+OUTPUT_MODEL_REPOSITORY=${OUTPUT_MODEL_REPOSITORY:=$(get_output_directory)}
+CONFIG_FILE="config.yml"
+FILENAME_SERVER_ONLY="server-metrics.csv"
+FILENAME_INFERENCE_MODEL="model-metrics-inference.csv"
+FILENAME_GPU_MODEL="model-metrics-gpu.csv"
```
dyastremsky, line 34: A lot (all?) of these variables overlap for the optuna tests. I wonder if there's a way to store them in a shared optuna variable file in a shared utils folder that can just be imported. That would reduce code duplication and make maintenance easier.

qa/L0_optuna_search/test.sh
```diff
@@ -0,0 +1,95 @@
...
+create_result_paths
+
+python3 test_config_generator.py --profile-models $MODEL_NAMES
+
+# Run the analyzer and check the results
+RET=0
+
+set +e
+
+MODEL_ANALYZER_ARGS="-m $MODEL_REPOSITORY -f $CONFIG_FILE"
```
dyastremsky, line 46: Would it be possible to make this a multi-line command where you're only setting MODEL_ANALYZER_ARGS once, or is this because Bash won't process them in a cleanly formatted multi-line statement (i.e. Bash needs this to be on one line)?

dyastremsky (commented):
Splendid work here! Generally very clean code. Appreciated the clean-up and loved a lot of choices like using safe loading for the config files.

My comments apply across files, so if you address my comments in one place, please check across files to make sure it's fixed everywhere. I think the biggest thing though is that a lot of the files seem like duplicates. If there are slight differences, e.g. in the check_results files' number of trials, those can still be put into a shared file/folder with slightly different parameters that they receive. That'd make updating, maintaining, and documenting them easier.

I think even the test.sh files might be able to combined into one large test.sh file or have a base script which you import or run, just changing the MA command or variables a bit. With those changes, the check_results.py and test_config_generator.py files would be mostly eliminated except for one copy and the test.sh files would be pretty minimal. That'd make reading and updating them easier.

qa/L0_optuna_bls_model/check_results.py
nv-braf: Fixed all copyright dates.

qa/L0_optuna_bls_model/check_results.py
nv-braf: Might not be the latest, but this is the copyright used throughout the entire MA codebase

qa/L0_optuna_bls_model/check_results.py
nv-braf: At maximum there are 4 configs (top-3, default) x 9 concurrencies (not 10, since at least one measurement was already taken per configuration).

I've updated the comments to better reflect this

dyastremsky (approved):
Brian confirmed that this is the current approach for MA testing due to the lengthy CI time. If MA has a faster CI in the future, we can optimize out the shared pieces while still allowing these tests to run in parallel in the CI.

PR #913, opened 2024-07-18
Add Optuna Seed to Checkpoint

nv-braf:
Adds the seed used in Optuna to the checkpoint file. This seed is then read back anytime a checkpoint exists ensuring that we take the same path through the search algorithm for subsequent runs.

There's not an easy way to unit test this, but I have confirmed this works via a live run on my local machine (and we now have CI tests protecting Optuna).

### DIFF
--- a/tests/test_optuna_run_config_generator.py
+++ b/tests/test_optuna_run_config_generator.py
@@ -68,13 +68,14 @@ class TestOptunaRunConfigGenerator(trc.TestResultCollector):
 
         self._rcg = OptunaRunConfigGenerator(
             config=config,
+            state_manager=MagicMock(),
             gpu_count=1,
             models=self._mock_models,
             composing_models=[],
             model_variant_name_manager=ModelVariantNameManager(),
             search_parameters={"add_sub": search_parameters},
             composing_search_parameters={},
-            seed=100,
+            user_seed=100,
         )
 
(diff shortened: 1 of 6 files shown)


tests/test_optuna_run_config_generator.py
```diff
@@ -315,7 +318,7 @@ def test_create_run_bls_config(self):
         sub_model_config = run_config.model_run_configs()[0].composing_configs()[1]
         perf_config = run_config.model_run_configs()[0].perf_config()
 
-        # BLS (Top Level Model) + PA Config (Seed=100)
+        # BLS (Top Level Model) + PA Config (user_seed=100)
```
debermudez, line 321: All of your seeds are 100.
Are we sure that the seed is actually impacting the randomization vs setting the seed to a known value and assuming it is having an impact?

nv-braf: Yes. Without a seed or with a different seed the unit tests will fail. I also observe in live testing that a different path is taken when different random seeds are chosen.

debermudez: As long as we have done that due diligence, this is good to go.
nice work man!

debermudez (approved):

ganeshku1 (approved):

PR #914, opened 2024-07-18
Trigger CI on Pull Requests

nvda-mesharma:


the-david-oy (approved):

the-david-oy:
@nvda-mesharma @mc-nv Would it be possible to add this type of automation to this repo? If it does not take too much effort.

PR #915, opened 2024-07-19
Add request rate parameter for optuna search

lkomali:
PR to support request-rate parameter in optuna mode.

### DIFF
--- a/model_analyzer/config/generate/optuna_run_config_generator.py
+++ b/model_analyzer/config/generate/optuna_run_config_generator.py
@@ -37,7 +37,10 @@ from model_analyzer.config.generate.search_parameter import (
 )
 from model_analyzer.config.generate.search_parameters import SearchParameters
 from model_analyzer.config.input.config_command_profile import ConfigCommandProfile
-from model_analyzer.config.input.config_defaults import DEFAULT_BATCH_SIZES
+from model_analyzer.config.input.config_defaults import (
+    DEFAULT_BATCH_SIZES,
+    DEFAULT_RUN_CONFIG_MIN_REQUEST_RATE,
+)
 from model_analyzer.config.run.model_run_config import ModelRunConfig
 from model_analyzer.config.run.run_config import RunConfig
 from model_analyzer.constants import LOGGER_NAME
@@ -669,12 +672,17 @@ class OptunaRunConfigGenerator(ConfigGeneratorInterface):
             model_config.get_field("name"), self._config
         )
 
-        default_concurrency = self._calculate_default_concurrency(model_config)
-
-        perf_config_params = {
-            "batch-size": DEFAULT_BATCH_SIZES,
-            "concurrency-range": default_concurrency,
-        }
+        if self._config.is_request_rate_specified(self._search_parameters):
+            perf_config_params = {
+                "batch-size": DEFAULT_BATCH_SIZES,
+                "request-rate-range": DEFAULT_RUN_CONFIG_MIN_REQUEST_RATE,
+            }
+        else:
+            default_concurrency = self._calculate_default_concurrency(model_config)
+            perf_config_params = {
+                "batch-size": DEFAULT_BATCH_SIZES,
+                "concurrency-range": default_concurrency,
+            }
         default_perf_analyzer_config.update_config(perf_config_params)
 
         default_perf_analyzer_config.update_config(model.perf_analyzer_flags())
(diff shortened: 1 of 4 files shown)


tests/test_search_parameters.py
```diff
@@ -74,6 +74,14 @@ def setUp(self):
             max_range=10,
         )
 
+        self.search_parameters._add_search_parameter(
```
nv-braf, line 77: Concurrency and request rate will never be in the same configuration. This is fine for unit testing, but if we're going to do this we need a lot a comments explaining why. Probably better to just create a new search_parameters instance that has request rate instead of concurrency.

tests/test_search_parameters.py
```diff
@@ -455,8 +552,8 @@ def test_total_possible_configurations(self):
             self.search_parameters.number_of_total_possible_configurations()
         )
 
-        # max_batch_size (8) * instance group (8) * concurrency (11) * size (3)
-        self.assertEqual(8 * 8 * 11 * 3, total_num_of_possible_configurations)
+        # max_batch_size (8) * instance group (8) * concurrency (11) * size (3) * request_rate(10)
```
nv-braf, line 555: See my comment above - RR and concurrency should not be mixed.

model_analyzer/config/generate/search_parameters.py
```diff
@@ -130,7 +130,7 @@ def _populate_parameters(self) -> None:
 
         if not self._is_composing_model:
             self._populate_concurrency()
```
nv-braf, line 137: You need to check if request rate has been specified by the user:
There is a function in `config_command_profile::is_request_rate_specified()` that you should be using here

lkomali: yeah, I knew theoretically they shouldn't be set together, but I couldn't figure how to do the check. Thanks, I will make the changes.

model_analyzer/config/generate/search_parameters.py
```diff
@@ -134,7 +134,9 @@ def _populate_parameters(self) -> None:
         self._populate_batch_sizes()
 
         if not self._is_composing_model:
-            if self._config.is_request_rate_specified(self._parameters):
+            if self._parameters and self._config.is_request_rate_specified(
```
nv-braf, line 137: The pre-condition of self._parameters is not correct and should be removed. There are other cases (based on just values in self._config) that can cause this method to return true.

tests/test_optuna_run_config_generator.py
```diff
@@ -270,6 +270,8 @@ def test_create_run_bls_config(self):
         bls_model = ModelProfileSpec(
             config.profile_models[0], config, MagicMock(), MagicMock()
         )
+        print("BLS Model: ")
```
nv-braf, line 273: Remove debug prints

nv-braf (commented):
Looks good. Just need a unit test to protect this and a live run to ensure things are working end-to-end.

tests/test_optuna_run_config_generator.py
```diff
@@ -295,6 +298,7 @@ def test_create_objective_based_run_config_with_request_rate(self):
         self.assertEqual(model_config.to_dict()["name"], self._test_config_dict["name"])
         self.assertEqual(perf_config["batch-size"], DEFAULT_BATCH_SIZES)
         self.assertEqual(perf_config["request-rate-range"], 2048)
+        self.assertNotIn("concurrency", perf_config)
```
nv-braf, line 301: This should be:
```self.assertEqual(perf_config["concurrency-range"], None)```

Comment (lkomali) [reply tests/test_optuna_run_config_generator.py:301]: modified it locally. didn't update the change due to debugging.

**Comment by lkomali** [comment]
Modified TestOptunaRunConfig, SearchParameters and their unit tests to allow optuna search based on request-rate.

**Comment by nv-braf** [comment]
I see changes in your branch that are unrelated (coming from a more recent check-in to main). Please pull main and rebase your branch to remove these.

File: model_analyzer/config/generate/optuna_run_config_generator.py [hunk 7]
```diff
@@ -670,8 +670,10 @@ def _create_default_perf_analyzer_config(
         default_perf_analyzer_config.update_config_from_profile_config(
             model_config.get_field("name"), self._config
         )
-        if self._config.is_request_rate_specified(
-            self._search_parameters[
+
```
nv-braf, line 673: This is technically correct, but you can just call `self._search_parameters.get_parameter("request_rate")` instead

lkomali: This gave error but ```
if self._search_parameters[
                 model_config.get_field("name")
            ].get_parameter("request_rate"):
``` worked.  Pushing the change.

nv-braf (approved):
Please submit once it passes the CI.

PR #921, opened 2024-07-29
Creating new generator class for concurrency sweeping

nv-braf:
This refactors the concurrency sweep that occurs after quick/optuna profiling completes into a new generator (so that the code is no longer duplicated).

No unit testing is done for this class as the !/$ is not there and this class is covered in 8 different tests in the CI.


lkomali (approved):
Looks good to me!

PR #922, opened 2024-07-31
Updating documentation based on VDR

nv-braf:
This covers the following changes requested in the VDR:

- Define commonly used terminology
- Create a support matrix
- Divide search mode and model types
- Update examples on CLI page


lkomali (approved):
LGTM

PR #938, opened 2024-10-08
Update README and versions for 24.10

pvijayakrish:
Updtae README and versions for 24.10


fpetrini15 (approved):
After our discussion, I believe changing the versioning in the dockerfile is optional. I think our build will take the version from `VERSION`.

pvijayakrish:
The version update in Dockerfile will be done pre-NGC release.

PR #952, opened 2024-12-19
Adding option to disable DCGM when in remote mode

nv-braf:
Add a new CLI option to disable DCGM: `--disable-dcgm` 

This will disable MA's ability to get info about the GPUs, which will prevent us from loading models in local, docker, or c_api, as well as verifying that the current GPUs match the ones in a checkpoint.

I've added a check to ensure this new option is only available in remote mode and added a warning message informing the user that MA cannot verify GPUs if a checkpoint is provided.


lkomali (approved):
LGTM

PR #961, opened 2025-02-14
Update 'main' to track the 25.03 release

nv-kmcgill53:


Dockerfile
```diff
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:24.12-py3
-ARG TRITONSDK_BASE_IMAGE=nvcr.io/nvidia/tritonserver:24.12-py3-sdk
+ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:25.01-py3
```
debermudez, line 15: should this be 25.01?

VERSION
mc-nv, line 1: As development is stopped for MA. We may not want to update version and keep last one released

Dockerfile
nv-kmcgill53: Not sure I understand, it is 25.01?

VERSION
debermudez: We just need to make sure nothing is broken each release at this point. 
So long as a pipeline has been triggered to ensure nothing is broken, we can close this PR.

Dockerfile
debermudez: thougth we were already at 25.02
but @mc-nv has the correct comment. This actually doesn't need to be done at all.

VERSION
nv-kmcgill53: What is going to be the strategy if there is a bug which needs to be patched? Will that go into a new release or will it go into a re-release of the final version of MA? This is the last issue I want resolved before I close this PR

debermudez: @nv-braf how are we handling that?

nv-braf: We've stopped active development of the tool (for now), but it is highly likely that we will have to add a new feature or implement a fix in the future, especially if the interface to PA changes. 
What is prevent us from releasing a new version of MA in the future?

nv-kmcgill53: Nothing from the operations side as far as I can tell. We will manage your project how you tell us.

